PR #14, opened 2017-07-28
Initial support for OkHttp3 as backend

omainegra:
This PR includes `sync` and `Future` based versions. Also all the test passed.

### DIFF
--- a/okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
+++ b/okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
@@ -35,9 +35,11 @@ abstract class OkHttpClientHandler[R[_], S](client: OkHttpClient)
     })
 
     //OkHttp support automatic gzip compression
-    request.headers.filter(_._1 != "Accept-Encoding").foreach {
-      case (name, value) => builder.addHeader(name, value)
-    }
+    request.headers
+      .filter(_._1.equalsIgnoreCase(AcceptEncodingHeader) == false)
+      .foreach {
+        case (name, value) => builder.addHeader(name, value)
+      }
 
     builder.build()
   }
@@ -74,8 +76,8 @@ abstract class OkHttpClientHandler[R[_], S](client: OkHttpClient)
     Response(body, res.code(), headers.toList)
   }
 
-  private[okhttp] def readResponseBody[T](res: OkHttpResponse,
-                                          responseAs: ResponseAs[T, S]): T = {
+  private def readResponseBody[T](res: OkHttpResponse,
+                                  responseAs: ResponseAs[T, S]): T = {
     responseAs match {
       case IgnoreResponse => res.body().close()
       case ResponseAsString(encoding) =>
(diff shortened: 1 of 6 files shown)


okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    val body = setBody(request.body)
+    builder.method(request.method.m, body.getOrElse {
+      if (HttpMethod.requiresRequestBody(request.method.m))
+        OkHttpRequestBody.create(null, "")
+      else null
+    })
+
+    //OkHttp support automatic gzip compression
+    request.headers.filter(_._1 != "Accept-Encoding").foreach {
```
adamw: In what sense the gzip support is automatic? Does it add the `Accept-Encoding: gzip, deflate` header by default? Is it an error to specify the header yourself?

As for the actual filtering, I think it would be better to do `_._1.eqalsIgnoreCase(AcceptEncodingHeader)` as (1) you never know what case the user is going to specify the header in and (2) we can utilize a constant

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
```diff
@@ -0,0 +1,132 @@
...
+
+    val headers = res
+      .headers()
+      .names()
+      .asScala
+      .flatMap(name => res.headers().values(name).asScala.map((name, _)))
+    Response(body, res.code(), headers.toList)
+  }
+
+  private[okhttp] def readResponseBody[T](res: OkHttpResponse,
```
adamw: [minor] can't this simply be `private`?

adamw:
Looks great, just two minor comments, thanks a lot for the work!

adamw:
Ah, and one more thing - please add yourself to the contributors list at the bottom of the readme :)

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
omainegra: **OkHttp3** automatically send in the request the header `Accept-Encoding: gzip`. And it handle compression automatically, I have tested passing the header but in that case it doesn't handle compression, it's up to us. For more information see: square/okhttp#2277

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
omainegra: Done

adamw:
Release in 0.0.4. Thanks! :)

omainegra:
I'm glad I could help. I will start working in the streaming part with akka
and monix. You can create the issue and assign it to me because I think I
don't have the required permissions.

adamw:
@omainegra you should be able to create issues? other people have :)

PR #24, opened 2017-08-04
Okhttp3 monix

### LINKED ISSUE #17: OkHttp3 backend with Monix response streaming

omainegra:
Monix with OkHttp handler supporting `Observable[ByteBuffer]` streaming. This closes #17


okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
```diff
@@ -0,0 +1,70 @@
...
+          )
+          .runAsync
+
+      override def contentType(): MediaType = null
+    })
+
+  override def responseBodyToStream(
+      res: okhttp3.Response): Try[Observable[ByteBuffer]] =
+    Success(
+      Observable.fromInputStream(res.body().byteStream()).map(ByteBuffer.wrap))
```
adamw: Thanks for the patch! I'm not entirely sure about the streaming support - is it *true* non-blocking streaming like in the other cases? 

That is:

* is the request body sent in a streaming fashion (without buffering everything first in memory)?
* is the response body received in a streaming fashion (again, without buffering in-memory)?
* are both operations non-blocking?

omainegra: Ok, let me try to explain
 1) Yes it is. It was implemented according to *OkHttp* streaming example in its [Recipes](https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/PostStreaming.java). Also it is non-blocking because the streaming is done using a `Monix` task running in the **IO Scheduler** (I will push this right away)
 2) Yes it is. The response body `InputStream` is piped directly from the network (it will buffer if you call `string()`. It is non-blocking for the same reasons as the previous line.

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
```diff
@@ -0,0 +1,72 @@
...
+
+  override def streamToRequestBody(
+      stream: Observable[ByteBuffer]): Option[OkHttpRequestBody] =
+    Some(new OkHttpRequestBody() {
+      override def writeTo(sink: BufferedSink): Unit =
+        stream
+          .consumeWith(
+            Consumer.foreach(chunk => sink.write(chunk.array()))
+          )
+          .runAsync(io)
```
adamw: I'm not sure this is going to work correctly. Looking at [the example](https://github.com/square/okhttp/blob/master/samples/guide/src/main/java/okhttp3/recipes/PostStreaming.java) from OkHttp, the `writeTo` method is **synchronous** - that is, it should only complete when the entire stream has been written. Otherwise, OkHttp has no way to tell if the request is finished or not.

This is still useful - however I think it should be changed to run synchronously and clearly marked in the docs that it might block the calling thread (the thread that's calling `send()`).

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
adamw: I think everything that uses an `InputStream` is potentially blocking - as the `InputStream` interface has no way of signalling "waiting for more input, check back later" or "give me a callback to signal you when new data is ready".

Looking at the implementation of `RealBufferedSource.inputStream()`, it looks like data is buffered into memory, and if it is not available, a potentially blocking call to `source.read` is done. So this is definitely useful as well, but again, I think it would be worth clearly documenting that the call might block a thread - I suppose in this case it could be on the the `Scheduler`'s threads?

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
omainegra: The this is,`writeTo` do nothing. It just tell how to consume the `Observable`. And start it asynchronous, running in another `ThreadPool` recommended for *IO* operations. Hence nothing blocks the thread that calls `send()` I'm working in in demo to prove it (or to convince myself I'm wrong)

adamw: Yes, and that's the problem - I think that's not the contract of `OkHttpRequestBody.writeTo`. As far as I understand the examples, it should block until the content is written in full. Otherwise, OkHttp has no way of determining when the request body is finished and the request send complete.

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
omainegra: The `InputStream` is wrapper around an `Observable`, so the blocking operations are moved away form the calling thread to another `ThreadPool` for *IO* operations. Also the buffering `RealBufferedSource.inputStream()` is just the buffer size with as usual it's 8192 bytes

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
omainegra: Ok, I get your point, let me dig deeper in that. I think I have a turn around to keep non-blocking the streaming while calling `OkHttpRequestBody.writeTo` as per its contract

omainegra: It turns out that `OkHttpRequestBody.writeTo` is called by OkHttp in another `ThreadPool`, different to the one which call `send()`. So I guess it's save to wait there until the stream is sent to the sink.

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+  * Created by omainegra on 8/4/17.
+  */
+class OkHttpMonixClientHandler private (client: OkHttpClient)(
+    implicit s: Scheduler)
+    extends OkHttpAsyncClientHandler[Task, Observable[ByteBuffer]](client,
+                                                                   TaskMonad) {
+
+  private lazy val io = Scheduler.io("sttp-monix-io")
```
adamw: I think we should at least give the user an option to provide a custom io scheduler, not always create a new one. Maybe an additional parameter, `ioScheduler: Option[Scheduler] = None` (defaulting to no scheduler, in which situation we create it)?

tests/src/test/scala/com/softwaremill/sttp/StreamingTests.scala
```diff
@@ -131,9 +134,70 @@ class StreamingTests
     }
   }
 
+  def monixAsyncHttpClientStreamingTests(): Unit = {
```
adamw: maybe the tests could be parametrized with a `SttpHandler[Task, Observable[ByteBuffer]]` and same code used for both the async-http and okhttp handlers?

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
```diff
@@ -108,38 +114,43 @@ class OkHttpSyncClientHandler private (client: OkHttpClient)
...
 
+class OkHttpFutureClientHandler private (client: OkHttpClient)(
+    implicit ec: ExecutionContext)
+    extends OkHttpAsyncClientHandler[Future, Nothing](client, new FutureMonad) {}
+
 object OkHttpFutureClientHandler {
   def apply(okhttpClient: OkHttpClient = new OkHttpClient())(
       implicit ec: ExecutionContext = ExecutionContext.Implicits.global)
-    : SttpHandler[Future, Nothing] =
+    : OkHttpFutureClientHandler =
```
adamw: do you think it's better to return a more specific type?

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+      override def writeTo(sink: BufferedSink): Unit = {
+        val f = stream
+          .consumeWith(
+            Consumer.foreach(chunk => sink.write(chunk.array()))
+          )
+          .runAsync(io)
+
+        // We could safely block until the observable is consumed because OkHttp execute
+        // this method asynchronous in another ThreadPool.
+        Await.ready(f, Duration.Inf)
```
adamw: This looks like a dangerous thing to do. Will this fail when the request times out, if there's no data sent for a long time?

Another thing is that this blocks the thread calling `send()`. Maybe when there's a streaming body, we should run the construction of the OkHttp request (which involves setting the body), on the `io` thread pool - that is, wrapping it in an async `Task`. What do you think?

tests/src/test/scala/com/softwaremill/sttp/StreamingTests.scala
omainegra: Done

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
omainegra: I improve this part with suggestion from Alexandru in Monix gitter channel. The tip was convert the `Observable` to an `Iterable`. About the blocking part that's not correct. The thread that call `send` is never blocked because `OkHttp` internally run that code in another thread pool. Please check this example I've prepared.

```scala
// Start HttpServer
implicit val actorSystem: ActorSystem = ActorSystem()
implicit val materializer = ActorMaterializer()

val serverRoutes: Route =
  path("echo") {
    post {
      entity(as[String])(body => {
        complete(body)
      })
    }
  }

val endpoint = uri"http://localhost:31532"
Await.ready(
  Http().bindAndHandle(serverRoutes, endpoint.host, endpoint.port.get),
  Duration.Inf)

// Run Demo
implicit val scheduler = Scheduler.io("io")
implicit val handler = OkHttpMonixClientHandler()

val body = "sample string"
val observable = Observable
  .fromIterable(body.getBytes("utf-8"))
  .map(byte => ByteBuffer.wrap(Array(byte)))
  .delayOnNext(1.second)
  .doOnNext(byte => println2(s"Item produced: $byte"))

val request = sttp.post(uri"$endpoint/echo").streamBody(observable)
val response = request.response(asString).send().runAsync

while (!response.isCompleted) {
  println2(s"Still waiting ...")
  Thread.sleep(1000)
}
println2(s"Got response: ${response.value.get.get.body}")

Await.ready(actorSystem.terminate(), Duration.Inf)
assert(response.value.get.get.body == body)
```

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
omainegra: I review the code and actually there is no need for another `Scheduler`, the default provided is enough

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
```diff
@@ -0,0 +1,79 @@
...
+import monix.reactive.{Consumer, Observable}
+import okhttp3.{MediaType, OkHttpClient, RequestBody => OkHttpRequestBody}
+import okio.BufferedSink
+
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+import scala.language.higherKinds
+import scala.util.{Failure, Success, Try}
+
+/**
```
omainegra: Done

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
omainegra: Fixed

okhttp-client-handler/monix/src/main/scala/com/softwaremill/sttp/okhttp/monix/OkHttpMonixClientHandler.scala
adamw: First, sorry for the late reply - I was on vacations. And thanks for the example - it's always good to test things as it turns out :) Looks much better now, with the iterable change.

okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
```diff
@@ -46,12 +46,13 @@ abstract class OkHttpClientHandler[R[_], S](client: OkHttpClient)
...
-  private def setBody(requestBody: RequestBody[S]): Option[OkHttpRequestBody] = {
-    requestBody match {
+  private def setBody[T](request: Request[T, S]): Option[OkHttpRequestBody] = {
+    request.body match {
       case NoBody => None
-      case StringBody(b, encoding, _) =>
-        Some(OkHttpRequestBody.create(MediaType.parse(encoding), b))
-      case ByteArrayBody(b, _) => Some(OkHttpRequestBody.create(null, b))
+      case StringBody(b, _, _) =>
+        Some(OkHttpRequestBody.create(null, b))
```
adamw: One last thing - any reason why you no longer use the `encoding` here?

omainegra: Yes, about that , the thing is that `MediaType.parse(encoding)` always fail because of the missing information (like type and subtype). With the encoding only we can't create the `MediaType`

adamw: Ah, that's a good reason :)

adamw:
Releasing 0.0.7 - thanks again!

PR #25, opened 2017-08-05
Add circe support module

aeons:
I don't know if this is something you would like here or if it should be kept separate.

It's not a lot of code, but I can see it being useful.

I'm not completely sold on having the `ApplicationJsonContentType` in this module instead of in core, but if there's not references to anything json related in core, I guess it's ok.


circe/src/main/scala/com/softwaremill/sttp/circe.scala
```diff
@@ -0,0 +1,32 @@
...
+package object circe {
+  private[sttp] val ApplicationJsonContentType = "application/json"
+
+  private def serializeBody[B](body: B)(
+      implicit encoder: Encoder[B]): RequestBody[Nothing] =
+    SerializableBody((b: B) => StringBody(encoder(b).noSpaces, Utf8), body)
+
+  implicit final class CirceRequestTOps[U[_], T, +S](val req: RequestT[U, T, S])
+      extends AnyVal {
+    def jsonBody[B: Encoder](body: B): RequestT[U, T, S] =
```
adamw: Here's another approach, maybe worth considering: right now all the bodies are set via the `.body` overloaded method, with the exception of stream bodies (`.streamBody`), which has to be separate as it changes an additional type parameter on `RequestT`. There's also a variant of the `.body` method which allows any body as long at it serializes to a `BasicBody`:

```
type BodySerializer[B] = B => BasicRequestBody
```

Maybe this could be made into a `BodySerializer` as well? Maybe this would keep help the API more consistent?

One downside is that the default content type wouldn't be set. But maybe it would be possible to extend the `BasicRequestBody` with a `def defaultContentType: Option[String]` method, which would specify the content type to set, if there's none and one is defined?

It would then be the responsibility of the `BodySerializer` to return a `BasicRequestBody` with the appropriate data set (and that would be the responsibility of the library).

What do you think?

adamw:
I was wondering about adding JSON support, and didn't come to any final conclusions, so let's try :) I left two notes about the code.

Thanks for the patch!

build.sbt
```diff
@@ -126,6 +126,17 @@ lazy val okhttpClientHandler: Project = (project in file(
     )
   ) dependsOn core
 
+lazy val circe: Project = (project in file("circe"))
+  .settings(commonSettings: _*)
+  .settings(
+    name := "circe",
+    libraryDependencies ++= Seq(
+      "io.circe" %% "circe-core" % "0.8.0",
```
aeons: Yes. :)

circe/src/main/scala/com/softwaremill/sttp/circe.scala
aeons: That sounds like it could work. However, since the body is not serialized until the request is actually evaluated, the `defaultContentType` would have to be set on the `SerializableBody`, right?

aeons: No, that doesn't make sense. I get it now I think. I'll try working it out tomorrow.

adamw:
Thanks for the changes! However, I had something a bit different in mind: adding the defaultContentType to BasicRequestBody. That way:

* it would be posaible to vary the content type basing on the input
* BodySerializer would remain a simple function type alias
* existing defaults (eg set in body(String)) would be instead set as defaults on respective BasicRequestBody types.

Your solutioj would work good as well. What do you think?

Adam

aeons:
That would work as well. The reason I did this instead is that currently the `BodySerializer` is run at execution time, so it will not be possible to inspect the content type of the request before sending it.

adamw:
Ah true... good point :) question is, do we want to serialize the body when sending or when creating the request? The other body types are eager I think?

aeons:
I like the neatness of having the serializer be lazy, but you are right that the basic body types are all eager.

If it's changed, the `SerializableBody` goes out the window, right? The serializer will just be run on the call to `.body` and only the basic body will be saved.

adamw:
Yes, `SerializableBody` would go out of the window. I think that doing the serialization once when the body is set is least surprising and safest - and also what other libraries do.

adamw:
Released in 0.0.6

PR #27, opened 2017-08-18
Add fs2 streaming module and refactor streaming tests

aeons:
Sorry, this is all in one lump, but I didn't want to do a lot of duplicated work adding tests and then refactoring them.

I'm not 100% sold on the names for the streaming test traits/classes, but it's a decent start, imo.

### DIFF
--- a/async-http-client-handler/fs2/src/main/scala/com/softwaremill/sttp/asynchttpclient/fs2/Fs2AsyncHttpClientHandler.scala
+++ b/async-http-client-handler/fs2/src/main/scala/com/softwaremill/sttp/asynchttpclient/fs2/Fs2AsyncHttpClientHandler.scala
@@ -1,0 +1,88 @@
+package com.softwaremill.sttp.asynchttpclient.fs2
+
+import java.nio.ByteBuffer
+
+import cats.effect._
+import com.softwaremill.sttp.asynchttpclient.AsyncHttpClientHandler
+import com.softwaremill.sttp.{MonadAsyncError, SttpHandler}
+import fs2._
+import fs2.interop.reactivestreams._
+import org.asynchttpclient.{
+  AsyncHttpClient,
+  AsyncHttpClientConfig,
+  DefaultAsyncHttpClient
+}
+import org.reactivestreams.Publisher
+
+import scala.concurrent.ExecutionContext
+import scala.language.higherKinds
+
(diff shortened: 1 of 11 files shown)


adamw:
Thanks! And sorry for the delay, I've been on vacations.

I think there's some overlap in the streaming tests refactor with #24 - I'll try to merge that one first, and than this PR (as 24 < 27 :) ), which will probably create some conflicts. You've taken a bit of a different approach to the tests, feel free to comment on #24 if you think there are some advantages/disadvantages of either one.

aeons:
I looked at that PR and didn't notice the changes to the tests :)

I like his naming, but my modularization better.

I will update my PR after you merge #24, with what I think would be a good combination, and then we can take a look again, if that fits you.

adamw:
#24 merged

aeons:
I believe I took the best parts of my ideas and the best parts of @omainegra's and combined them.

Currently, each type of streaming handler has its own file, to unclutter the main test file. Adding a new streaming handler is just adding one file and adding one line in `StreamingTests`.

Let me know what you think.

async-http-client-handler/fs2/src/main/scala/com/softwaremill/sttp/asynchttpclient/fs2/Fs2AsyncHttpClientHandler.scala
```diff
@@ -0,0 +1,88 @@
...
+  DefaultAsyncHttpClient
+}
+import org.reactivestreams.Publisher
+
+import scala.concurrent.ExecutionContext
+import scala.language.higherKinds
+
+class Fs2AsyncHttpClientHandler[F[_]: Effect] private (
+    asyncHttpClient: AsyncHttpClient,
+    closeClient: Boolean)(implicit ec: ExecutionContext)
```
adamw: is the `ExecutionContext` needed for the fs2<->reactive streams integration?

aeons: Yes.

https://github.com/zainab-ali/fs2-reactive-streams/blob/master/core/src/main/scala/fs2/interop/reactivestreams/package.scala

PR #31, opened 2017-08-28
Close Ok Http bodies

aeons:
It seems Ok Http bodies were not closed in all cases, which emitted a warning some time after running tests.

I don't know how Ok Http works, so this is a best effor at closing the bodies in all cases.

I'm not sure about the `ResponseAsStream` case.

### DIFF
--- a/okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
+++ b/okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
@@ -167,7 +167,7 @@ abstract class OkHttpAsyncClientHandler[R[_], S](client: OkHttpClient,
 
           override def onResponse(call: Call, response: OkHttpResponse): Unit =
             try success(readResponse(response, r.responseAs))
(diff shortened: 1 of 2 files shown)


okhttp-client-handler/src/main/scala/com/softwaremill/sttp/okhttp/OkHttpClientHandler.scala
```diff
@@ -139,7 +149,7 @@ abstract class OkHttpAsyncClientHandler[R[_], S](client: OkHttpClient,
 
           override def onResponse(call: Call, response: OkHttpResponse): Unit =
             try success(readResponse(response, r.responseAs))
-            catch { case e: Exception => error(e) }
+            catch { case e: Exception => error(e) } finally response.close()
```
adamw: Indeed, the responses need closing, however I think that's too aggressive - it fails in case of streams, as the response is closed before the stream is consumed (as the failing test shows).

aeons:
I changed the closing for the async case. It is now not done by the base handler, and is the responsibility of the caller.

I updated the Monix wrapper and closed the body after the stream is done (I think, I don't know Monix very well).

adamw:
thanks!

PR #32, opened 2017-08-28
Rename handler subprojects to be consistent

aeons:
To be consistent with the Async Http Client handlers.

You could argue the others should be renamed, to be consistent with the package names :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ val query = "http language:scala"
 // `sort` is removed, as the value is not defined
 val request = sttp.get(uri"https://api.github.com/search/repositories?q=$query&sort=$sort")
   
-implicit val handler = HttpURLConnectionSttpHandler
+implicit val handler = HttpURLConnectionHandler
 val response = request.send()
 
 // response.header(...): Option[String]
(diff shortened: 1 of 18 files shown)


omainegra:
Indeed. Also there is `AkkaHttpSttpHandler` which has `Sttp` in its name, which sounds redundant.

adamw:
Indeed, that's inconsistent. But as you point out, wouldn't actually `[backend library][wrapper name]Handler` make more sense? E.g. `AsyncHttpClientCatsHandler` and so on. I think the backend library is more important here.

aeons:
Yes, the package names are actually `[backend library]-handler-[effect library]`.

I will redo this PR with everything changed around :)

adamw:
Needs a rebase

aeons:
On it :)

aeons:
All handlers are now named `[ClientLibrary][EffectLibrary]Handler`. `OkHttp` is now always `OkHttp`, without `Client`. `sttp` was removed from some handler names.

aeons:
I didn't change any of the module names, so they are still kinda inconsistent.

adamw:
Which module names? I only noticed `okhttp-client-handler` which I renamed to `okhttp-handler` - anything else?

aeons:
Thanks, that was mostly the one i was thinking about. You could argue that `async-http-client-handler-cats` and `AsyncHttpClientCatsHandler` is not consistent, but I think it reads better this way.

PR #37, opened 2017-09-25
Fix NPE in HTTPUrlConnBackend in case of empty response

cb372:
I'm not sure of the exact conditions under which `HTTPUrlConnection#getInputStream()` can return null, but I added a test for the exact scenario that I encountered: client sends a POST request, server responds with a 401 with no body and no content-type.

Fixed with a simple null check.


adamw:
Release as 0.0.16, thanks!

PR #38, opened 2017-10-25
Test stub matching partial function mapping request to response

amorfis:
Allows test stub to match partial function like: 
```
stub.whenRequestMatchesPartial({
  case r if r.method == Method.POST => Response(Right(10), 200, Nil, Nil)
  case r if r.uri.path.endsWith(List("partialAda")) => Response(Right("Ada"), 200, Nil, Nil)
})
```


adamw:
Could you also add docs here: https://github.com/softwaremill/sttp/blob/master/docs/backends/testing.rst & to credits: https://github.com/softwaremill/sttp/blob/master/docs/credits.rst ?

amorfis:
Docs added

adamw:
released in 1.0.1

Issue #10, opened 2017-07-28
Proxy server support


adamw:

robsonpeixoto:
That is a good way to define proxy
http proxy: http://docs.python-requests.org/en/master/user/advanced/#proxies
socks5 proxy: http://docs.python-requests.org/en/master/user/advanced/#socks

omainegra:
Anybody working on this? I can take it

adamw:
@omainegra go ahead :)

adamw:
any luck with that issue maybe? :)

omainegra:
@adamw Still working on this, sorry for the delay but I got hit by Irma and electricity and internet just came back.

adamw:
@omainegra sorry to hear that, hope that you & your family is fine! no hurry with the code, just checking :)

atais:
I would really like to see that working 👍

adamw:
@omainegra hey, is everything fine? we didn't hear from you for a while

adamw:
@omainegra I hope you are fine, for now I implemented proxy support as an option passed to the backend when creating it. Hope to hear from you in the future :)

adamw:
Released in 0.0.17

atais:
How can I use it with `Monix` Backend?

adamw:
For example:

```
implicit val backend = AsyncHttpClientMonixBackend(
  options = SttpBackendOptions.httpProxy("some.host", 8080))

sttp.get(uri"...").send()
```

Sure, reading the default java proxy parameters is an option - though I don't like this style of "global" configuration. Maybe you'd like to attempt making a PR? :)

atais:
> Maybe you'd like to attempt making a PR? :)

I sure will try 👍

PR #39, opened 2017-10-29
Loading system proxy values by default

atais:
Like I have suggested in #10 I believe loading of System proxy values should be added.

I have created a simple code to handle that action.

----
Just a quick note:

I have found that generally, proxy topic is a bit broader: [link](https://stackoverflow.com/a/120802/1549135)

> Also don't forget the http.nonProxyHosts property! 

So this would mean some extra logic in the lib. However, I did not dive deeper into this as I did not find it necessary. Just mentioning fyi. 

Thanks!

### DIFF
--- a/core/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
+++ b/core/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
@@ -47,7 +47,7 @@ object SttpBackendOptions {
     Empty.copy(proxy = loadSystemProxy)
 
   def connectionTimeout(ct: FiniteDuration): SttpBackendOptions =
-    Empty.connectionTimeout(ct)
+    Default.connectionTimeout(ct)
 
   def httpProxy(host: String, port: Int): SttpBackendOptions =
     Empty.httpProxy(host, port)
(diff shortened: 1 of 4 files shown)


core/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
```diff
@@ -37,12 +40,38 @@ object SttpBackendOptions {
...
-  val Default: SttpBackendOptions = SttpBackendOptions(30.seconds, None)
+  private val Empty: SttpBackendOptions =
+    SttpBackendOptions(30.seconds, None)
+
+  val Default: SttpBackendOptions =
+    Empty.copy(proxy = loadSystemProxy)
 
   def connectionTimeout(ct: FiniteDuration): SttpBackendOptions =
-    Default.connectionTimeout(ct)
+    Empty.connectionTimeout(ct)
```
adamw: I'm wondering, wouldn't the end-user expect `SttpBackendOptions.connectionTimeout` to change only the connection timeout comparing to the defaults? In other words, shouldn't we modify `Default` here?

adamw:
Ah and another thing - could you add yourself to `credits.rst`? :)

core/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
atais: Wow. Did not think about it, but you are right. Fixed.

adamw:
Released in 1.0.2 - thanks! :)

PR #46, opened 2017-11-11
Uri enhancement

Iryna-:


adamw:
Thanks! Maybe you could also add a test verifying that the conversion works?

adamw:
I'll merge & add the unit test

Issue #51, opened 2017-11-29
SttpBackendStub.thenRespondOk() creates a Response with Left(body) instead of Right(body)


dpovey:
This is not what I expected. I was given to understand that the contract for Sttp is that successful responses should always be a Right(body).  I think this should change from:

```
 def thenRespondOk(): SttpBackendStub[R, S] =
      thenRespondWithCode(200)
```

to

```
 def thenRespondOk(): SttpBackendStub[R, S] =
      thenRespond("")
```

Another option is to have thenRespondWithCode change to:

```
def thenRespondWithCode(code: Int,
                            msg: String = ""): SttpBackendStub[R, S] =
      val body = if (code >= 200 && code < 300) Right(msg) else Left(msg)
      thenRespond(Response[Nothing](body, code, Nil, Nil))
```

Which also allow more sensible mocking of 201 etc responses.

PR #52, opened 2017-11-29
#51 When stubbing using with thenRespondWithCode, treat 2xx as a success and return Right(body)

dpovey:
Change SttpBackendStub so that response obey the contract that a 2xx will be returned as Right(xxx) and other codes will be Left(xxx).


adamw:
Thanks, released 1.1.1!

PR #56, opened 2017-12-23
Add TryMonad, implement a delegating TryBackend

joshlemer:
Allows users to use a simple synchronous HTTP backend which doesn't throw exception.

### DIFF
--- a/core/src/main/scala/com/softwaremill/sttp/TryBackend.scala
+++ b/core/src/main/scala/com/softwaremill/sttp/TryBackend.scala
@@ -19,11 +19,10 @@ class TryBackend[-S](delegate: SttpBackend[Id, S]) extends SttpBackend[Try, S] {
   override def responseMonad: MonadError[Try] = TryMonad
 }
 
-object TryBackend {
+object TryHttpUrlConnectionBackend {
   def apply(options: SttpBackendOptions = SttpBackendOptions.Default,
-            customizeConnection: HttpURLConnection => Unit = { _ =>
-              ()
-            }): SttpBackend[Try, Nothing] =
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/softwaremill/sttp/TryBackend.scala
```diff
@@ -0,0 +1,29 @@
...
+  override def send[T](request: Request[T, S]): Try[Response[T]] =
+    Try(delegate.send(request))
+
+  override def close(): Unit = delegate.close()
+
+  override def responseMonad: MonadError[Try] = TryMonad
+}
+
+object TryBackend {
+  def apply(options: SttpBackendOptions = SttpBackendOptions.Default,
```
adamw: Thanks a lot for the contribution! Definitely worth including :)

I'm just wondering whether to include the `TryBackend` object - won't it be confusing that `TryBackend` creates a wrapped `HttpURLConnectionBackend`, while `new TryBackend(...)` is a generic wrapper?

It would be a bit more code to instantiate a default `Try`-backed synchronous wrapper, but for the sake of simplicity, `TryBackend(HttpURLConnectionBackend())` wouldn't leave any doubts.

joshlemer: You're absolutely correct. Iis a little bit inconsistent isn't it? I was considering making it a TryHttpURLConnectionBackend() but didn't know if that would be too verbose. I'll change it a bit tomorrow.

adamw:
Released in 1.1.3 - thanks!

joshlemer:
:-)

Issue #57, opened 2017-12-25
Cross-compile to Scala.js


joshlemer:
I'm working on this in this branch https://github.com/joshlemer/sttp/tree/scalajs   
  
Here I will keep a list of any java dependencies we would need to eliminate, or implement in scala:

### java.net.IDN
```
[error] Referring to non-existent class java.net.IDN$
[error]   called from com.softwaremill.sttp.Uri.encodeHost()java.lang.String
[error]   called from com.softwaremill.sttp.Uri.toString()java.lang.String
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$5(com.softwaremill.sttp.Uri,java.lang.String)org.scalatest.compatible.Assertion
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$4(java.lang.String,scala.Tuple2)scala.Unit
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$2(scala.Tuple2)scala.Unit
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.<init>()
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.__exportedInits
[error]   exported to JavaScript with @JSExport
```
  
### java.net.URLEncoder 
```
[error] Referring to non-existent class java.net.URLEncoder$
[error]   called from com.softwaremill.sttp.Uri.encodeQuery(java.lang.String,com.softwaremill.sttp.Uri$QueryFragmentEncoding)java.lang.String
[error]   called from com.softwaremill.sttp.Uri.encodeQueryFragments$1(scala.collection.immutable.List,scala.Boolean,scala.collection.mutable.StringBuilder)java.lang.String
[error]   called from com.softwaremill.sttp.Uri.toString()java.lang.String
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$5(com.softwaremill.sttp.Uri,java.lang.String)org.scalatest.compatible.Assertion
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$4(java.lang.String,scala.Tuple2)scala.Unit
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.$$anonfun$new$2(scala.Tuple2)scala.Unit
[error]   called from com.softwaremill.sttp.UriInterpolatorTests.<init>()
[...]

PR #58, opened 2017-12-29
Remove dependence on StringTokenizer

joshlemer:
StringTokenizer is not supported by Scala.js (#57 ), and rather than implement it, since the [official docs](https://docs.oracle.com/javase/7/docs/api/java/util/StringTokenizer.html) claims that   
  
>StringTokenizer is a legacy class that is retained for compatibility reasons although its use is discouraged in new code. It is recommended that anyone seeking this functionality use the split method of String or the java.util.regex package instead.
  
I figure it is easier to just use `String.split`


adamw:
Thanks!

Issue #59, opened 2018-01-09
add statusText/reason/message to Response


ILLKO:
very common, how to name it better?

| lib  | int  | text  | 
| --- | --- | --- |
| sttp  | code  |  add statusText/reason/message? |   
| js fetch  |  status | statusText  |   
| gigahorse  | status  | statusText |   
| akka http | StatusCode.intValue | StatusCode.reason |
| HttpURLConnection | getResponseCode | getResponseMessage |
| asynchttpclient | getStatusCode | getStatusText |
| okhttp3 | code | message |

adamw:
Released in 1.1.5. Thanks @intracer !

adamw:
sorry, 1.1.4 :)

PR #60, opened 2018-01-09
add status message to Response #59

intracer:


ILLKO:
It is not the whole status line, just the element "3. A status text" (as called in js fetch, gigahorse and asynchttpclient). I called it message as sttp seamed to folllow java's HttpURLConnection and rename getResponseCode to code, hence getResponseMessage to message.

adamw:
@ILLKO then maybe call it `statusText`? I think that `message` might be too general and can lead to confusion, as to what this actually is. `statusText` is rather unambiguous.

intracer:
renamed to statusText

PR #61, opened 2018-01-24
Allow SttpBackendStub to accept monad with response

amorfis:
This is the only way not to hold thread busy when testing with SttpBackendStub and Futures with long response time (to test, well, scenario when server takes a long time to respond).

### DIFF
--- a/core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
+++ b/core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
@@ -103,11 +103,16 @@ class SttpBackendStub[R[_], S] private (
     def thenRespondWithCode(code: Int,
                             msg: String = ""): SttpBackendStub[R, S] = {
       val body = if (code >= 200 && code < 300) Right(msg) else Left(msg)
-      thenRespondWithMonad(rm.unit(Response(body, code, msg, Nil, Nil)))
+      thenRespond(Response(body, code, msg, Nil, Nil))
     }
     def thenRespond[T](body: T): SttpBackendStub[R, S] =
-      thenRespondWithMonad(
-        rm.unit(Response[T](Right(body), 200, "OK", Nil, Nil)))
+      thenRespond(Response[T](Right(body), 200, "OK", Nil, Nil))
+    def thenRespond[T](resp: => Response[T]): SttpBackendStub[R, S] = {
+      val m: PartialFunction[Request[_, _], R[Response[_]]] = {
+        case r if p(r) => rm.unit(resp)
+      }
+      new SttpBackendStub(rm, matchers.orElse(m), fallback)
+    }
     def thenRespondWithMonad(resp: => R[Response[_]]): SttpBackendStub[R, S] = {
       val m: PartialFunction[Request[_, _], R[Response[_]]] = {
         case r if p(r) => resp
--- a/core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
+++ b/core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
@@ -28,7 +28,7 @@ import scala.util.{Failure, Success, Try}
   */
 class SttpBackendStub[R[_], S] private (
     rm: MonadError[R],
-    matchers: PartialFunction[Request[_, _], Response[_]],
+    matchers: PartialFunction[Request[_, _], R[Response[_]]],
     fallback: Option[SttpBackend[R, S]])
     extends SttpBackend[R, S] {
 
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
```diff
@@ -99,12 +103,13 @@ class SttpBackendStub[R[_], S] private (
     def thenRespondWithCode(code: Int,
                             msg: String = ""): SttpBackendStub[R, S] = {
       val body = if (code >= 200 && code < 300) Right(msg) else Left(msg)
-      thenRespond(Response(body, code, msg, Nil, Nil))
+      thenRespondWithMonad(rm.unit(Response(body, code, msg, Nil, Nil)))
     }
     def thenRespond[T](body: T): SttpBackendStub[R, S] =
-      thenRespond(Response[T](Right(body), 200, "OK", Nil, Nil))
-    def thenRespond[T](resp: => Response[T]): SttpBackendStub[R, S] = {
```
adamw: Can't the `thenRespond(=> Response[T])` variant be kept as-is?

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
```diff
@@ -99,12 +103,13 @@ class SttpBackendStub[R[_], S] private (
...
-      val m: PartialFunction[Request[_, _], Response[_]] = {
+      thenRespondWithMonad(
+        rm.unit(Response[T](Right(body), 200, "OK", Nil, Nil)))
+    def thenRespondWithMonad(resp: => R[Response[_]]): SttpBackendStub[R, S] = {
```
adamw: And maybe an overloaded version of `thenRespond(=> R[Response[_]])` would work as well?

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
```diff
@@ -159,13 +164,19 @@ object SttpBackendStub {
...
-    r.body match {
-      case Left(_) => r
-      case Right(body) =>
-        val newBody: Any = tryAdjustResponseBody(ra, body).getOrElse(body)
-        r.copy(body = Right(newBody))
+  private[sttp] def tryAdjustResponseType[DesiredRType, RType, M[_]](
+      rm: MonadError[M],
+      ra: ResponseAs[DesiredRType, _],
+      m: M[Response[RType]]): M[Response[DesiredRType]] = {
+    rm.map[Response[RType], Response[DesiredRType]](m) { r =>
```
adamw: [minor] probably could be a bit shorter by just saying `.map { _.body match { ... `

core/src/test/scala/com/softwaremill/sttp/testing/SttpBackendStubTests.scala
```diff
@@ -155,6 +157,25 @@ class SttpBackendStubTests extends FlatSpec with Matchers with ScalaFutures {
 
   }
 
+  it should "not hold the calling thread when passed a future monad" in {
```
adamw: I'm not sure this is a good test ... the fact that the calling thread isn't held is ensured by the way the `Future` is constructed - not by anything in sttp itself. For this test to fail sttp would need to block on futures somewhere, which would be very hard to do given the generic `Monad` implementation :)

Maybe a better test case here would be that when using `.thenRespondWithMonad`, the specified response is indeed returned, when the future returned by `.send` completes?

amorfis: That was my first attempt. Getting `Future` from `.send` and checking if that was the same as passed to `.thenRespondWithMonad`. However, it was not the same because of mapping in `SttpBackendStub.tryAdjustResponseType` (line 171).

Unless you mean just testing basic functionality of `.thenRespondWithMonad`? It's true that it's hard to block on future in a case we have here. It's all because that was our test scenario, which STTP was not feasible for, thus the change :)

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
amorfis: How? We use `r` in both match cases.

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
amorfis: Unfortunately, this throws StackOverflowError, because after type erasure it matches `.thenRespond(body: T)`, which calls `.thanRespond` then, which matches `.thenRespond(body: T)`... 

We could change the name of method with `body: T` parameter, but I don't think that's a good idea ;)

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
amorfis: Indeed it can and should. My mistake.

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
adamw: True :)

core/src/test/scala/com/softwaremill/sttp/testing/SttpBackendStubTests.scala
adamw: I meant testing the basic functinality of `.thenRespondWithMonad`. I don't think it matters if it's the exact same future or not :)

core/src/main/scala/com/softwaremill/sttp/testing/SttpBackendStub.scala
adamw: Ah right. Well, we'll have to live with that ;)

core/src/test/scala/com/softwaremill/sttp/testing/SttpBackendStubTests.scala
amorfis: Test for response retrieved from Monad added

adamw:
Released in 1.1.5 with slight naming modifications

PR #75, opened 2018-03-27
Added overloaded thenRespondWrapped which can return response monad but takes request as parameter

amorfis:
…ut takes request as parameter


adamw:
Released in 1.1.12

PR #79, opened 2018-04-19
Improve documentation for fs2

lancelet:
It took me a while to discover that `fs2` provides a streaming backend for `cats.effect.IO`. This PR expands the documentation for the `asynchttpclient` backend to cover `fs2` in the same way as Monix.

- Add fs2 dependencies and implicit backend val.
- Add fs2 streaming example.
- Fix Monix request example to include the implicit backend val.


lancelet:
(PS: AFAICT, the build failure is not related to this documentation-only change.)

adamw:
Thanks! Indeed, that section of the documentation was lacking :)

The build failure is unrelated, it's something that happens only on Travis, didn't have time yet to debug ;)

PR #82, opened 2018-04-23
Make max redirects configurable

n4to4:
It would be nice if `maxRedirects` was configurable.

### DIFF
--- a/core/src/main/scala/com/softwaremill/sttp/RequestT.scala
+++ b/core/src/main/scala/com/softwaremill/sttp/RequestT.scala
@@ -229,6 +229,9 @@ case class RequestT[U[_], T, +S](
   def followRedirects(fr: Boolean): RequestT[U, T, S] =
     this.copy(options = options.copy(followRedirects = fr))
 
+  def maxRedirects(n: Int): RequestT[U, T, S] =
+    this.copy(options = options.copy(maxRedirects = n))
+
   def tag(k: String, v: Any): RequestT[U, T, S] =
     this.copy(tags = tags + (k -> v))
 
@@ -281,4 +284,8 @@ class SpecifyAuthScheme[U[_], T, +S](hn: String, rt: RequestT[U, T, S]) {
     rt.header(hn, s"Bearer $token")
 }
 
-case class RequestOptions(followRedirects: Boolean, readTimeout: Duration)
+case class RequestOptions(
+    followRedirects: Boolean,
+    readTimeout: Duration,
+    maxRedirects: Int = FollowRedirectsBackend.MaxRedirects
+)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/softwaremill/sttp/RequestT.scala
```diff
@@ -229,6 +229,9 @@ case class RequestT[U[_], T, +S](
   def followRedirects(fr: Boolean): RequestT[U, T, S] =
     this.copy(options = options.copy(followRedirects = fr))
 
+  def maxRedirects(n: Int): RequestT[U, T, S] =
+    this.copy(options = options.copy(maxRedirects = n))
```
adamw: Maybe this should interact with `followRedirects` as well? If you are setting `maxRedirects`, then you probably also want to follow redirects?

Going a bit further, maybe it would make sense to have a single `redirects: Option[Int]` field, so that the state `followRedirects = true, maxRedirects = X` is not possible to represent? But maybe that would be not so obvious looking at the API ...

n4to4: Makes sense. Adding `followRedirects = true` should look better.

Now I have another concern though, what should we do if a negative count is passed? Maybe we should just set `followRedirects = true, maxRedirects = FollowRedirectsBackend.MaxRedirects`? I assume throwing an exception just for constructing request is a bad idea. (not sure)

Regarding API, I've researched a little about other Scala/Java HTTP libraries. Only `async-http-client` has similar config API as `sttp` and configurable maxRedirects, though. (`http4s`'s API looks a bit different)
Since `async-http-client` is the background for `sttp`, I think it is reasonable to have Boolean and Int to configure `followRedirects` and `maxRedirects`. What do you think?

| library                | follow redirects?   | max redirect count |
|------------------------|---------------------|--------------------|
| [async-http-client][1] | `boolean` in config | `int` in config    |
| [http4s][2]            | Use `FollowRedirect` middleware | `Int` parameter |
| [dispatch][3]          | `Boolean` parameter | N/A                |
| [scalaj-http][4]       | `Boolean` parameter | N/A                |

[1]: https://github.com/AsyncHttpClient/async-http-client/blob/async-http-client-project-2.4.5/client/src/main/java/org/asynchttpclient/DefaultAsyncHttpClientConfig.java#L134-L135
[2]: https://github.com/http4s/http4s/blob/v0.18.9/client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala#L42
[3]: https://github.com/dispatch/reboot/blob/v0.14.0/core/src/main/scala/requests.scala#L454-L459
[4]: https://github.com/scalaj/scalaj-http/blob/2.3.0/src/main/scala/scalaj/http/Http.scala#L70

adamw:
released as 1.1.13, thanks!

PR #86, opened 2018-05-03
Separated out the project dependencies from build.sbt to Dependencies…

dagaharshit:
This commit will take out all the external library dependencies into a new file such that it would be easier to maintain all the external dependencies.

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -1,0 +1,90 @@
+import sbt._
+
+object Dependencies {
+
+  import Versions._
+
+  val akkaHttp = "com.typesafe.akka" %% "akka-http" % akkaHttpVersion
+
+  val akkaStreams = "com.typesafe.akka" %% "akka-stream" % akkaStreamsVersion
+
+  val akkaStreamProvided = "com.typesafe.akka" %% "akka-stream" % akkaStreamsVersion % "provided"
+
+  val asyncHttpClient = "org.asynchttpclient" % "async-http-client" % asyncHttpClientVersion
+
+  val brave = "io.zipkin.brave" % "brave" % braveVersion
+
+  val braveInstrumentationHttp = "io.zipkin.brave" % "brave-instrumentation-http" % braveVersion
+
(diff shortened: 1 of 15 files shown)


adamw:
Thanks! Though I must admit I'm not a fan of such a separation - for example, you can't see what are the dependencies for specific modules. But maybe other might weight in with their comments?

annabee:
I prefer having dependencies in a separate file too (much easier to read), but I would separate them logically into collections corresponding to modules.

project/Dependencies.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  val betterFiles = "com.github.pathikrit" %% "better-files" % betterFilesVersion
+
+  val logbackClassic = "ch.qos.logback" % "logback-classic" % logbackClassicVersion
+
+  /**
+    * Libraries to be used in Test.........
+    */
+
+  val braveInstrumentationHttpTest = "io.zipkin.brave" % "brave-instrumentation-http-tests" % braveVersion
```
annabee: I would suggest using `% test` when declaring the dependencies. I think it makes more sense.

project/Dependencies.scala
```diff
@@ -0,0 +1,82 @@
+import sbt._
+import Keys._
+
+object Dependencies {
+
+  val scalaVer = "2.12.6"
```
annabee: What do you think about factoring the version numbers into a separate Version object?

project/Dependencies.scala
dagaharshit: I've updated them.

project/Dependencies.scala
dagaharshit: Moved to Versions object in same file.

dagaharshit:
Hi @annabee. I have made the required changes. Requesting you to kindly review. Thanks !

darkfrog26:
I also prefer dependencies to be in the `build.sbt` file as it makes it a particular pain for people just taking a look at the project for the first time trying to decide what dependencies it has.

adamw:
Closing for now due to lack of feedback.

PR #99, opened 2018-05-31
Add constants for common HTTP header names, media types, and status codes

### LINKED ISSUE #88: Provide constants for standard http values
It would be handy to have constants for standard http values like header names, status codes, content types, etc. 

Play exposes these via https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/http/StandardValues.scala

Have you considered providing these?

jsw:
This resolves #88 

This adds trait/object pairs for HTTP header names, media types, and status codes.

### Header Names
A complete list of header names is [here](https://www.iana.org/assignments/message-headers/message-headers.xml#perm-headers), though not sure it's worth adding all of them, so I started with the set defined by [Play](https://github.com/playframework/playframework/blob/5f576efda61d9a1bd4cd27215e1fe221ddbaa3a9/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L237). I'm open to using a different source.

### Media Types
A complete list of media types is [here](https://www.iana.org/assignments/media-types/media-types.xhtml), though not sure it's worth adding all of them, so I started with the set defined by [Play](https://github.com/playframework/playframework/blob/5f576efda61d9a1bd4cd27215e1fe221ddbaa3a9/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L101). I'm open to using a different source.

### Status Codes
This was easy to use the complete list defined [here](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). I did not include any non-standard ones like 418, 420, etc.

Note I had to change the imports to `BasicTests.scala` to avoid conflict with `StatusCodes` and `MediaTypes`.


jsw:
I'm not sure what happened, but my fork is way behind, possibly because I already had the repo forked from previous commits. I think I'll just recreate this PR from scratch...

PR #100, opened 2018-05-31
Add HeaderNames, MediaTypes, StatusCodes

### LINKED ISSUE #88: Provide constants for standard http values

jsw:
This resolves #88 

This adds trait/object pairs for HTTP header names, media types, and status codes.

### Header Names
A complete list of header names is [here](https://www.iana.org/assignments/message-headers/message-headers.xml#perm-headers), though not sure it's worth adding all of them, so I started with the set defined by [Play](https://github.com/playframework/playframework/blob/5f576efda61d9a1bd4cd27215e1fe221ddbaa3a9/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L237). I'm open to using a different source.

### Media Types
A complete list of media types is [here](https://www.iana.org/assignments/media-types/media-types.xhtml), though not sure it's worth adding all of them, so I started with the set defined by [Play](https://github.com/playframework/playframework/blob/5f576efda61d9a1bd4cd27215e1fe221ddbaa3a9/framework/src/play/src/main/scala/play/api/http/StandardValues.scala#L101). I'm open to using a different source.

### Status Codes
This was easy to use the complete list defined [here](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml). I did not include any non-standard ones like 418, 420, etc.

The next step would be to integrate these constants with sttp itself. I haven't looked at how much work that would be. Could do it either as part of this PR or a separate PR.


adamw:
I've updated the code using the constants, maybe you can check if everything looks good? Always good to have another pair of eyes :) https://github.com/softwaremill/sttp/commit/574aee7a322abae646fd7c851e8259c85ec841ae

I suppose what's missing is some mention in docs?

jsw:
Commit looks good. Agree the docs need a mention (maybe just a single page describing the 3 types of available constants with github links to see the full set of values?) and probably updated examples, e.g. [here](https://github.com/softwaremill/sttp/blob/master/docs/requests/headers.rst#headers).

Curious if you have considered scaladocs. I usually use the IDE, but it's sometimes nice to use a browser. Not sure what most other people do.

adamw:
@jsw do you mean publishing scaladocs? why not ... but where? :) Would you maybe have some time to add the docs?

jsw:
@adamw Good question. I just discovered [javadoc.io](https://www.javadoc.io/).

I can update the main docs.

adamw:
@jsw if you know how to integrate that into the build, go ahead :)

PR #104, opened 2018-06-04
add docs on http constants

jsw:
@adamw I didn't update the existing docs to use these constants. Let me know if you want me to do that, and if you prefer objects or traits. I wasn't sure the best place to place this new page, so I just made it top-level for now.


adamw:
Thanks!

PR #109, opened 2018-06-05
Removing usage of java.net.URLDecoder from core

pcejrowski:
Since there is no `java.net.URLDecoder` in `scala-native` and the decode functionality is implemented in `Rfc3986` class, it would be handy to get rid of that dependency in the entire `core` module.


adamw:
Thanks!

PR #110, opened 2018-06-05
Removing URLEncoder dependency

pcejrowski:
Scala Native doesn't have `java.net.URLEncoder` class so I switched to `Rfc3986` implementation for encoding. However, it caused some incompatibilities -- see changes in tests.
There is no `java.net.IDN` as well, so I dropped the `toASCII` conversion here. 
Why is that necessary?

**edit:** reverted test values to the previous version, since they were failing on Travis...


adamw:
The IDN conversion is necessary as some backends can't cope with domain names which have UTF characters. So we need to keep that - but it was already only done in the JVM version, so why is that a problem?

pcejrowski:
Correct. My bad. Closing this one.

PR #112, opened 2018-06-06
Don't ignore custom akka-http connection settings

sbrunk:
I've encountered an issue when trying to use a proxy with the akka-http backend. Proxy settings seem to be ignored. Here's a minimal example:

```scala
import com.softwaremill.sttp._
import com.softwaremill.sttp.akkahttp.AkkaHttpBackend
import scala.concurrent.Await
import scala.concurrent.duration._
import scala.concurrent.ExecutionContext.Implicits.global

implicit val backend = AkkaHttpBackend(
  options = SttpBackendOptions.httpProxy("localhost", 1234)
)
val result = sttp.get(uri"https://google.com").send()
Await.result(result, 5 seconds)
```

That should fail because there's no http proxy running at localhost:1234 but it doesn't:
```
res0: com.softwaremill.sttp.Response[String] =
Response(Right(<!doctype html><html itemscope=...
```

So I tried to create custom `ClientConnectionSettings` as described in the [akka-http](https://doc.akka.io/docs/akka-http/current/client-side/client-transport.html#use-http-s-proxy-with-) proxy docs and set them on the backend as part of `customConnectionPoolSettings`:

```Scala
val proxyHost = "localhost"
val proxyPort = 8888

val httpsProxyTransport = ClientTransport.httpsProxy(InetSocketAddress.createUnresolved(proxyHost, proxyPort))

val settings = ConnectionPoolSettings(system)
  .withConnectionSettings(ClientConnectionSettings(system)
    .withTransport(httpsProxyTransport))
implicit val backend = AkkaHttpBackend(
  customConnectionPoolSettings = Some(settings)
)
```

Still no luck. So I looked at the code and I think the issue is the following:

1. Given `ClientConnectionSettings` that are set as part of `ConnectionPoolSettings` are ignored because you create your own `connectionSettings` and update the `connectionPoolSettings` with them in the `send` method. That behaviour is a bit surprising for users of the API IMHO.
2.  Proxy settings  set through `SttpBackendOptions` are not used because `base.withTransport(ClientTransport.httpsProxy(p.inetSocketAddress))` updates the `ClientConnectionSettings` attached to `connectionPoolSettings` which are being ignored. That kind of proves my point about surprising behaviour. :-)

The proposed solution that should fix both issues is to remove `connectionSettings` completely and  just selectively update `ClientConnectionSettings` attached to `ConnectionPoolSettings`.

### DIFF
--- a/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
@@ -37,11 +37,10 @@ class AkkaHttpBackend private (actorSystem: ActorSystem,
 
   private val http = Http()
 
-  private val connectionSettings = ClientConnectionSettings(actorSystem)
-    .withConnectingTimeout(opts.connectionTimeout)
-
   private val connectionPoolSettings = {
-    val base = customConnectionPoolSettings.getOrElse(ConnectionPoolSettings(actorSystem))
+    val base = customConnectionPoolSettings
+      .getOrElse(ConnectionPoolSettings(actorSystem))
+      .withUpdatedConnectionSettings(_.withConnectingTimeout(opts.connectionTimeout))
     opts.proxy match {
       case None => base
       case Some(p) =>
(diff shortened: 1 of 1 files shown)


adamw:
Ah, good catch! I didn't notice that e.g. `withTransport` in effect updates the connection settings.

sbrunk:
Yeah I think that has actually changed in some recent akka-http release.

adamw:
1.2.0-RC6 is is published

PR #120, opened 2018-07-12
update async-http-client to fix below issue

nitsh:
https://github.com/AsyncHttpClient/async-http-client/issues/1559


adamw:
Thanks!

PR #90, opened 2018-05-12
Set read timeout properly for AsyncHttpClientBackend

rfkm:

PR #124, opened 2018-07-17
Differentiate between readTimeout and requestTimeout

lu4nm3:
Up until version 1.1.14, configuring the `readTimeout` for an sttp request actually configured the `requestTimeout` internally for the asynchttpclient request. However, this behavior was modified in this PR https://github.com/softwaremill/sttp/pull/90 to make things more consistent where setting the `readTimeout` for an sttp request also set the `readTimeout` for an asynchttpclient request.

It appears, however, that this is not enough when making long-lived, streaming requests similar to the following:

    val response: IO[Response[Stream[IO, ByteBuffer]]] =
      sttp
        .post(uri"...")
        .response(asStream[Stream[IO, ByteBuffer]])
        .send()
Which always keep timing out after 60,000ms. Using `readTimeout` to try to prolong the request timeout obviously has no effect.

What this means is that we need to expose the ability to configure the `requestTimeout` in addition to the `readTimeout`.

### DIFF
--- a/async-http-client-backend/src/main/scala/com/softwaremill/sttp/asynchttpclient/AsyncHttpClientBackend.scala
+++ b/async-http-client-backend/src/main/scala/com/softwaremill/sttp/asynchttpclient/AsyncHttpClientBackend.scala
@@ -152,6 +152,7 @@ abstract class AsyncHttpClientBackend[R[_], S](asyncHttpClient: AsyncHttpClient,
     val rb = new RequestBuilder(r.method.m)
       .setUrl(r.uri.toString)
       .setReadTimeout(if (readTimeout.isFinite()) readTimeout.toMillis.toInt else -1)
+      .setRequestTimeout(if (readTimeout.isFinite()) readTimeout.toMillis.toInt else -1)
     r.headers.foreach { case (k, v) => rb.setHeader(k, v) }
     setBody(r, r.body, rb)
     rb.build()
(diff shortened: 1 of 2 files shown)


lu4nm3:
@adamw Those are good points. You're right in that asyc-http-client seems to be the only one that truly distinguishes between read and request timeouts. 

The `AkkaHttpBackend`, for example, supports configuring things like `connectionTimeout` which is a timeout for establishing a connection and `idleTimeout` which as its name implies times out idle connections. Similarly, the `OkHttpBackend` has the concept of a connection timeout and a read timeout but nothing to really control the timeout of the request overall. So yeah, there's no real consistency between the different backends.

Of your 2 suggestions, I would probably favor the second once since it would be a bit more "universal" in the sense that the readTimeout has a more widespread use.

Another idea I had is to replace `readTimeout` with `requestTimeout` since the latter would apply to the overall request, regardless of the type of request, and not just for the cases where the request is reading. I updated the PR with the changes that I think would be needed for this. Let me know what you think. Otherwise I think it makes the most sense to go with your second approach.

lu4nm3:
That makes sense. I also prefer to keep things simpler and backwards compatible. I went ahead and updated the PR. Let me know what you think.

adamw:
Looks good, thanks! I'll make a release today or tomorrow

adamw:
A bit later as always, but it's released now in 1.3.0-RC2

Issue #121, opened 2018-07-14
Feature Request: Single Import Usage?


lihaoyi:
Currently the getting started example is this:

```scala
import com.softwaremill.sttp._
implicit val backend = HttpURLConnectionBackend()
sttp.get(uri"http://httpbin.org/ip").send()
```

Any chance we could add a convenience import to avoid the `implicit val`?

```scala
// brings in an `implicit val backend = HttpURLConnectionBackend()` automatically
import com.softwaremill.sttp.default._
sttp.get(uri"http://httpbin.org/ip").send()
```

Would make sttp much more convenient to use from e.g. the Ammonite REPL

adamw:
Interesting idea. Definitely possible - some changes would be needed (e.g. moving the contents of the package object to a trait, and extending it in both packages), but nothing major. Probably the package could be called sth indicating that it brings in a synchronous backend (`import com.softwaremill.sttp.sync._`?).

One worry I would have, though, is that using sttp in a REPL and in a "normal" application would be different. In REPL you would import the package that brings in the synchronous backend, while in a normal application you would import the one which requires a backend to be defined. Don't you think that this could confuse users?

A solution here could be to name the package so that it indicates that it's a "quick" or "REPL-only" solution, so maybe:

* `import com.softwaremill.sttp.quick._`
* `import com.softwaremill.sttp.repl._`

?

lihaoyi:
Sure the exact package name doesn't matter to me, any of your suggestions work

lihaoyi:
I'm not sure if there are other stateless backends too (e.g. an async one?) but for all the stateless ones it would make sense to add a default "no config" import to use them easily

PR #125, opened 2018-07-18
add one-import usage for sttp

lihaoyi:
https://github.com/softwaremill/sttp/issues/121

I basically moved the package object into a trait, made the package object inherit from it, and added two `simpleApi` objects (one on JVM, one on JS) that let you `import com.softwaremill.sttp.simpleApi._` and get the normal stuff imported into scope together with a default configuration of the relevant backend (`HttpUrlConnectionBackend` or `FetchBackend`)

Seems to work:
```scala
lihaoyi sttp$ sbt core/test:console

scala> import com.softwaremill.sttp.simpleApi._
import com.softwaremill.sttp.simpleApi._

scala> sttp.get(uri"http://www.google.com").send()
res0: com.softwaremill.sttp.Id[com.softwaremill.sttp.Response[String]] =
Response(Right(<!doctype html><html itemscope="" itemtype="http://schema.org/WebPage" lang="en-SG"><head><meta content="text/html; charset=UTF-8" http-equiv="Content-Type"><meta content="/logos/doodles/2018/gino-bartalis-104th-birthday-5380660112916480-law.gif" itemprop="image"><meta content="Gino Bartali&#8217;s 104th Birthday" property="twitter:title"><meta content="Gino Bartali&#8217;s 104th Birthday #GoogleDoodle" property="twitter:description"><meta content="Gino Bartali&#8217;s 104th Birthday #GoogleDoodle" property="og:description"><meta content="summary_large_image" property="twitter:card"><meta content="@GoogleDoodles" property="twitter:site"><meta content="https://www.google.com/...
```

let's see what CI says


mukel:
The `FetchBackend` is meant for browsers, it doesn't work on node. Ideally the "default" JS backend should work on both environments similar to `RosHttp`. But that's a story for another PR.

core/shared/src/main/scala/com/softwaremill/sttp/SttpApi.scala
```diff
@@ -0,0 +1,233 @@
...
+import java.nio.ByteBuffer
+
+import com.softwaremill.sttp.internal.SttpFile
+
+import scala.annotation.{implicitNotFound, tailrec}
+import scala.collection.immutable.Seq
+import scala.concurrent.duration._
+import scala.language.higherKinds
+
+trait SttpApi extends sttpExtensions {
```
adamw: Maybe `sttpExtensions` could now become `SttpApiExtensions`?

core/js/src/main/scala/com/softwaremill/sttp/simpleApi.scala
```diff
@@ -0,0 +1,5 @@
+package com.softwaremill.sttp
+
+object simpleApi extends SttpApi{
```
adamw: I think I would be leaning towards calling this package `quick`. What do you think?

adamw:
Thanks for the PR! I left two naming notes. Also, I think the README and docs can be updated with the new usage?

adamw:
@mukel if you would know how to write a node backend, that would be great :)

core/js/src/main/scala/com/softwaremill/sttp/simpleApi.scala
lihaoyi: sure

core/shared/src/main/scala/com/softwaremill/sttp/SttpApi.scala
lihaoyi: Maybe, but this was the existing name so I feel changing it is kind of orthogonal to this PR.

lihaoyi:
I renamed the things to `quick` as you requested

core/shared/src/main/scala/com/softwaremill/sttp/SttpApi.scala
adamw: That was for consistency, as previously these were extensions to the `sttp` package object. Hence the lower-case name prefix :)

lihaoyi:
capitalized SttpExtensions

PR #128, opened 2018-07-20
fix error with empty charset in header

jarrodu:
I had an issue with a RSS feed that gave a bad header.

```
implicit val backend = HttpURLConnectionBackend()
sttp.get(uri"http://rss.dw.com/xml/podcast_inside-europe").send()
```
This threw an exeception. `java.nio.charset.IllegalCharsetNameException`

With `curl` I get a this header.
```
HTTP/1.1 200 OK
Date: Fri, 20 Jul 2018 20:04:26 GMT
Content-Type: application/xml; charset=
Vary: Accept-Encoding
Access-Control-Allow-Origin: *
Set-Cookie: SERVERID=s2; path=/
Cache-control: private
Age: 0
Accept-Ranges: bytes
Connection: keep-alive
```

The `charset` is not defined properly.

This PR fixes the problem and allows me to get the data from the RSS feed.

If you need more details let me know.

### DIFF
--- a/core/shared/src/main/scala/com/softwaremill/sttp/internal/package.scala
+++ b/core/shared/src/main/scala/com/softwaremill/sttp/internal/package.scala
@@ -11,7 +11,7 @@ package object internal {
 
   private[sttp] def encodingFromContentType(ct: String): Option[String] =
     ct.split(";").map(_.trim.toLowerCase).collectFirst {
-      case s if s.startsWith("charset=") && s.substring(8) != "" => s.substring(8)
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! Do you think it would be possible to write a test for this? Also, maybe `trim`ing the result of the substring operation and then checking if it's not empty would be even more client-proof?

jarrodu:
I added your suggestion about 'trim'. You are right. This is better.

Maybe you could ask the original author of the file to write a test that covers the content of the file. ;-)

adamw:
Thanks! I guess I'm the original author ;-)

PR #129, opened 2018-07-20
Fixing tests for Core Native

pcejrowski:
There were two tests failing:
* different error message when file creation fails - 8af00ec unifies the behavior with JVM.
* missing `ISO-8859-2` support - similarly to ScalaJS, Native doesn't support that encoding, but on the way, I detected that no support for response content encoding was implemented. 09b2298 adds it.

### DIFF
--- a/core/jvm/src/test/scala/com/softwaremill/sttp/testing/HttpTestExtensions.scala
+++ b/core/jvm/src/test/scala/com/softwaremill/sttp/testing/HttpTestExtensions.scala
@@ -94,7 +94,7 @@ trait HttpTestExtensions[R[_]] extends TestHttpServer { self: HttpTest[R] =>
     }
   }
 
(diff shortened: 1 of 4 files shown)


adamw:
Thanks! This helped :)

Now, I'm *sometimes* (that's the best type of failure) getting errors in this test:

```
[info] - parse an empty error response as empty string *** FAILED ***
[info]   "[I'm compressed!]" was not equal to "[]" (SyncHttpTest.scala:323)
[info]   Analysis:
[info]   "[I'm compressed!]" -> "[]"
```

usually it passes, but every now and then it fails, sometimes with the long `Lorem ipsum ...` text, sometimes with this.

pcejrowski:
@adamw I had to run it 17 (sic!) times to reproduce, but here is the fix: https://github.com/softwaremill/sttp/pull/130 .
I ran the new code 100x, so hopefully, that should have been the issue.

PR #130, opened 2018-07-24
Curl Backend: fixing unpredictable behavior when empty response

pcejrowski:
The space allocated via `malloc` contains garbage and if there is no zero-terminated content in the response from curl, that space stays untouched and garbage is returned to the user.


adamw:
Thanks! :) That must have been a hard find

PR #138, opened 2018-08-03
Feature/non proxy hosts support 

nitsh:


core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
```diff
@@ -56,18 +74,21 @@ object SttpBackendOptions {
...
-    def proxy(t: ProxyType)(host: String, port: Int) = Proxy(host, port, t)
+    def proxy(t: ProxyType)(host: String, port: Int, nonProxyHosts: List[String]) = Proxy(host, port, t, nonProxyHosts)
 
     import ProxyType._
-    val socks = system("socksProxyHost", "socksProxyPort", proxy(Socks), 1080)
-    val http = system("http.proxyHost", "http.proxyPort", proxy(Http), 80)
-    val https = system("https.proxyHost", "https.proxyPort", proxy(Http), 443)
+    val socks = system("socksProxyHost", "socksProxyPort", None, proxy(Socks), 1080)
+    val http = system("http.proxyHost", "http.proxyPort", Some("http.nonProxyHosts"), proxy(Http), 80)
+    val https = system("https.proxyHost", "https.proxyPort", Some("http.nonProxyHosts"), proxy(Http), 443)
```
adamw: shouldn't this be `https.nonProxyHosts`?

core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
```diff
@@ -56,18 +74,21 @@ object SttpBackendOptions {
     Empty.socksProxy(host, port)
 
   private def loadSystemProxy: Option[Proxy] = {
-    def system(hostProp: String, portProp: String, make: (String, Int) => Proxy, defaultPort: Int) = {
+    def system(hostProp: String, portProp: String, nonProxyHostsPropOption: Option[String], make: (String, Int, List[String]) => Proxy, defaultPort: Int) = {
       val host = Option(System.getProperty(hostProp))
       def port = Try(System.getProperty(portProp).toInt).getOrElse(defaultPort)
-      host.map(make(_, port))
+      def nonProxyHosts: List[String] = nonProxyHostsPropOption
+        .flatMap(nonProxyHostsProp => Try(System.getProperty(nonProxyHostsProp).split("|").toList).toOption)
```
adamw: The `|` syntax of separating non-proxy hosts is somehow standard?

core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
```diff
@@ -21,8 +23,24 @@ case class SttpBackendOptions(
...
 
 object SttpBackendOptions {
-  case class Proxy(host: String, port: Int, proxyType: ProxyType) {
-    def asJava = new java.net.Proxy(proxyType.asJava, inetSocketAddress)
+  case class Proxy(host: String, port: Int, proxyType: ProxyType, nonProxyHosts: List[String] = Nil) {
+    def asJavaProxySelector: net.ProxySelector = new net.ProxySelector {
+      override def select(uri: URI): util.List[net.Proxy] = {
+        val proxyList = new util.ArrayList[net.Proxy](1)
+        val uriHost = uri.getHost
+        if (uriHost.startsWith("127.0.0.1") || nonProxyHosts.forall(uriHost.startsWith)) {
```
adamw: Shouldn't this be an `exists`? If so, maybe this can be factored out into a method on `Proxy` (there's another similar piece of code in `HttpURLConnectioBackend`)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -42,17 +42,18 @@ class AkkaHttpBackend private (actorSystem: ActorSystem,
     val base = customConnectionPoolSettings
       .getOrElse(ConnectionPoolSettings(actorSystem))
       .withUpdatedConnectionSettings(_.withConnectingTimeout(opts.connectionTimeout))
-    opts.proxy match {
-      case None => base
-      case Some(p) =>
-        base.withTransport(ClientTransport.httpsProxy(p.inetSocketAddress))
-    }
+    base
```
adamw: [minor] no need to introduce the `base` val here at this point :)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -42,17 +42,18 @@ class AkkaHttpBackend private (actorSystem: ActorSystem,
...
   }
 
   override def send[T](r: Request[T, S]): Future[Response[T]] = {
     implicit val ec: ExecutionContext = this.ec
 
-    val settings = connectionPoolSettings
+    val connectionPoolSettingsWithProxy =  opts.proxy match {
+      case Some(p) if !p.nonProxyHosts.contains(r.uri.host) =>
```
adamw: in other places you use `startsWith`?

adamw:
Thanks for the PR! I've left a couple of comments

core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
nitsh: According to this doc : https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html

nitsh: I just realised that it also allows pattern matching of hosts.

core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
nitsh: I'm still unsure if I should be using startsWith or plain equals. I guess plain equals should work.

nitsh:
I have to figure how to test this feature. Any suggestions?

core/shared/src/main/scala/com/softwaremill/sttp/SttpBackendOptions.scala
nitsh: I'm following this spec : https://docs.oracle.com/javase/8/docs/api/java/net/doc-files/net-properties.html And it says https would use the nonProxyHosts of http

nitsh:
@adamw I'm planning to reuse the code from AsyncHttpClient for wildcard matching - https://github.com/AsyncHttpClient/async-http-client/blob/master/client/src/main/java/org/asynchttpclient/proxy/ProxyServer.java#L100

adamw:
@nitsh sounds good :)

nitsh:
@adamw Done with changes and added tests. The wildcard(*) is only supported when it is a prefix or suffix like in asynchttpclient.

adamw:
Sorry for the delay, but released now in 1.3.1.

PR #143, opened 2018-09-05
Fixing the parsing of nonProxyHosts

janlindemann:
related to #132 

Passing the regex `"|"` to the split function would split the string on each character, i.e. `"localhost|127.*"` into `List("l", "o", "c", "a", ...)`. By passing `'|'` instead, the string will be splitted correctly into `List("localhost", "127.*")`.


adamw:
released in 1.3.2, thanks!

Issue #145, opened 2018-09-10
Can not send requests via socks proxy with Async-http-client


chrilves:
Hi,

I experienced some problems when trying to send request via a Socks proxy. The backend i used is Akka-http-client with Cats-effect. The execution of requests sent by socks results in the following exception:

```
org.asynchttpclient.exception.RemotelyClosedException: Remotely closed
  at org.asynchttpclient.exception.RemotelyClosedException.INSTANCE(Unknown Source)
```

## Steps to reproduce

```scala
import cats.effect._
import com.softwaremill.sttp._
import com.softwaremill.sttp.asynchttpclient.cats.AsyncHttpClientCatsBackend

implicit val backend: SttpBackend[IO, Nothing] =
    AsyncHttpClientCatsBackend[IO](options = SttpBackendOptions.socksProxy(host, port))

sttp.get(uri"https://github.com/softwaremill/sttp").send().unsafeRunSync()
```

it seems to come from how the `AsyncHttpClientBackend` configures the `ProxyServer`.
A pull request is on its way.

chrilves:
The [Pull request #146](https://github.com/softwaremill/sttp/pull/146)

PR #146, opened 2018-09-10
Fix ProxyServer config for Socks proxies in AsyncHttpClientBackend

chrilves:
Attempt to fix issue [#145 Can not send requests via socks proxy with Async-http-client](https://github.com/softwaremill/sttp/issues/145).

I'm not entirely satisfied by this because it does not handle SOCKS_V4 servers. I tried it only with Cats-effect's `IO` and `ssh` as socks server.


adamw:
Should be released soon in 1.3.3 :)

adamw:
Thanks!

PR #149, opened 2018-09-24
Filter out Some(null)

Starzu:
When `response.headers.get(HeaderNames.ContentType)` returns `null`, then `_.toOption` returns `Some(null)`. The implementation of `encodingFromContentType` suggests that it is not what you expected.


adamw:
Thanks! :)

Starzu:
@adamw Could you release version `1.3.4` with this bugfix?

adamw:
Yes, should be in central soon :)

PR #150, opened 2018-09-24
Fix NullPointerException of nonProxyHosts

5000164:
When System.getProperty(nonProxyHostsProp) returns null, the value becomes Success(null) by Try and NullPointerException occurs.


adamw:
Thanks! However, getting the property might throw an exception (due to security manager), so I think both `Try` and null-handling should be combined

5000164:
Thank you for your reply.
I fixed it.
How is this?

adamw:
Released in 1.3.5

PR #152, opened 2018-10-06
Add play-json support

stewSquared:
As of play 2.6, play-json is a standalone library, so we can support it without needing all of play framework as a dependency.

For more information: https://www.playframework.com/documentation/2.6.x/Migration26#Play-JSON-moved-to-separate-project


adamw:
Released as 1.3.6, thanks!

PR #156, opened 2018-10-23
Depend on json4s interface instead of native implementation

pm47:
I'm wondering why the json4s module explicitely depends on the json4s-native implementation, instead of using the interfaces and let the user choose a backend.

This is what https://github.com/hseeberger/akka-http-json is doing, I guess there must be a reason why you chose to do it that way?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -356,12 +356,15 @@ lazy val circe = crossProject(JSPlatform, JVMPlatform)
 lazy val circeJS = circe.js.dependsOn(coreJS)
 lazy val circeJVM = circe.jvm.dependsOn(coreJVM)
 
+lazy val json4sVersion = "3.6.0"
+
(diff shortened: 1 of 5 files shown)


adamw:
Thanks! Probably no reason, the person who contributed this probably uses the native version :). Maybe the docs would need updating to mention that an implicit `Serialization` value is needed?

pm47:
> Maybe the docs would need updating to mention that an implicit Serialization value is needed?

Done.

There is a tradeoff with this change though: users must be careful to link with a version of json4s backend that matches the one of json4s-core we are compiling sttp with. I think that is very acceptable, but it feels a bit less "plug and play".

adamw:
Released in 1.3.9. Thanks!

pm47:
That was fast!

PR #160, opened 2018-10-29
Add support for obtaining non 2xx response body.

mkubala:
This PR is aimed to bring support for obtaining unsuccessful response bodies. 

This will be especially helpful when working with endpoints that in case of error respond with a JSON object containing more detailed information about the failure cause (i.e. validation errors or service-specific error codes & reason descriptions).


adamw:
I've made two small changes, hope you don't mind :)

One thing I'm wondering about  - maybe it would make sense to put `parseResponseIf` in `RequestOptions`?

adamw:
Released in 1.4.0, thanks!

PR #163, opened 2018-10-31
Fix for deserialization errors while decoding Option[T] from empty responses

### LINKED ISSUE #161: Cannot decode `Option[T]` from 204 No Content responses
Sometimes I need to consume HTTP APIs that (depending on some circumstances) can respond with 204 status and no content.

I've been trying to handle this case using `.response(asJson[Option[T]])` which unfortunately leads to `Left(DeserializationError(,io.circe.ParsingFailure: exhausted input,ParsingFailure: exhausted input))` while using Circe.

I don't know if this issue affects also other serializers (I've been using STTP only with Circe so far..) but as long as there is no sensible reason for keep deserialization failing on the above case I'd like to provide appropriate tests cases for all of them and ensure that their behaviour is consistent.

mkubala:
This PR is aimed to fix #161

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -336,11 +336,18 @@ lazy val okhttpMonixBackend: Project =
   okhttpBackendProject("monix")
     .dependsOn(monixJVM % "compile->compile;test->test")
 
-lazy val jsonCommon: Project = (project in file("json/common"))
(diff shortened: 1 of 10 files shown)


adamw:
Released in 1.4.1

PR #164, opened 2018-11-05
Add `mapK` extensions for SttpBackend using cats/scalaz

kubukoz:
Simple, binary-breaking POC without tests/much documentation for now, as I want to hear what you think first :)

Often, in the cats-effect world, we want to have a single effect type (e.g. `EitherT[IO, SomeError, ?]`) in our programs, but the constraints that are placed on e.g. HTTP clients (like `AsyncHttpClientFs2Backend`) are impossible or hard to provide instances for (you can't have `ConcurrentEffect` for an arbitrary `EitherT`, unless you have a way to go from the error to `Throwable` - and I'm not sure that's a good idea), so we have to work with two effect types and make conversions between them in our code.

With `mapK`, we could make the conversion once and let the wrapping backend do the rest.

------
Work sponsored by [Ocado Technology](https://ocadotechnology.com).


adamw:
Doesn't look very scary :) Why would it be binary-breaking?

Anyway ... adding a backend wrapper is not a problem. I'm not sure about adding a new method to `SttpBackend`, though. That's not a very technical argument but the simplicity of the current 3 methods is kind of nice :).

But maybe, it would make sense to actually put this in modules implementing `cats`/`scalaz` integration? Then you wouldn't have to re-invent `FunctionK`, and `mapK` could be added with an extension method?

adamw:
(the integration could the be brought into scope e.g. with `import com.softwaremill.sttp.cats._`)

kubukoz:
Adding a method to a trait is breaking for <2.12 :)

Having 3 abstract methods is nice, but this wouldn't be an abstract one, so
I think we'd still be good... But I see what you mean :)

Yes, it probably makes sense to add to the cats/scalaz modules. I'll
proceed with that and some tests when I have some more time.

kubukoz:
I made `mapK` an extension in cats/scalaz modules for now, let me know if it looks good or if we need any extra work (e.g. tests) :)

Wesołych świąt ;)

adamw:
@kubukoz Thanks!

Most of other functionalities are provided in package objects, so maybe this one could be as well? Also, maybe `Implicits` could be a trait (extended by the package object), so that you can create your own single-import stack?

Finally, a short example in the docs would for sure be very beneficial for others :)

folex:
Hi! This is a great PR, thanks a lot! How can I help to move it a bit closer to be merged?

kubukoz:
Hi @folex, actually I have some bandwidth today so I'll try to get it there...

kubukoz:
@adamw I looked for places where the documentation could end up, but I think we'd need some new files... and I don't want to delay this PR further by learning how to work with readthedocs.io, so for now I'd ask you or @folex to finish this up...

As far as the objects/companion objects decision goes, I went with normal objects, as I think package objects are problematic for compilation performance and have some quirks to them. Projects like cats use usual objects for implicits.

kubukoz:
Rebased to latest master

adamw:
@kubukoz Thanks! I'll try to release shortly.
@kubukoz / @folex If you'd have time to add docs, that would be great. The docs are "normal" .rst files, readthedocs is just hosting.

Interesting about package objects & compilation - maybe something to consider for sttp2 & tapir :) Do you have any pointers on why they impact compilation performance?

DieMyst:
Hello @adamw. When can we see this project with these changes published?
Thank you!

kubukoz:
@adamw I have a rough memory of hearing that from someone in Virtuslab, maybe @romanowski knows more :D

adamw:
Released in 1.5.12

romanowski:
cc @lbialy

PR #165, opened 2018-11-12
Add support for variable request method

satabin:
In some cases, one may have a generic way of constructing a request out
of several pieces, where the method is passed as a variable of type
`Method`. Add a `method` setter that both set the method and the uri
amkes this possible.


adamw:
Thanks!
Published as 1.5.0-SNAPSHOT, I'll try to do a release in the coming days

PR #168, opened 2018-12-09
Add proxy auth

ybasket:
Some proxies (in my case luminati.io, but it's a general thing) require authentication and sttp so far models this in a way (on `RequestT`, setting the `Proxy-Authentication` header there) which doesn't work for all proxies - forcing the user to manually create the underlying backend before being able to use such a proxy with sttp.

This PR extends `SttpBackendOptions` to allow proxy username and password to be passed and adapts the `akka-http`, `async-http-client` and `okhttp` backends (these seem to be all which support this) to set themselves up accordingly.

I unfortunately don't know where to get a proper proxy set up with authentication, therefore no dedicated tests…if someone has an idea, please let me know!


adamw:
Thanks! One thing I'm wondering about, maybe you know, and maybe this could be added to the docs: when using the `Proxy-Authorization` header is enough, and when to use the options?

adamw:
Released as 1.5.1

PR #186, opened 2019-02-15
Maybe parse response based on returned metadata.

Ferada:
This extends `parseResponseIf` to cover the whole metadata, not only the status code.

I'm suggesting this because I have a case where I really also want to check some of the returned headers (e.g. content type) to skip parsing the returned response as a file.

### DIFF
--- a/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
@@ -79,7 +79,7 @@ class AkkaHttpBackend private (actorSystem: ActorSystem,
           .flatMap(encodingFromContentType)
 
         val responseMetadata = ResponseMetadata(headers, code, statusText)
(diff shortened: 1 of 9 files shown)


adamw:
Thanks!

Though I'm afraid I'll have to leave this PR for later, as it would be a breaking change. Maybe for 1.6

Ferada:
> breaking change

I guess I'm not quite clear on which classes are part of the API, because it's simply adding another method, or is that more of a concern for any other implementations of the backend interface?

adamw:
Ah, true! My bad, I thought this is a rename :) 

Though compilation fails in one of the modules: `[error] (asyncHttpClientBackend / Compile / compileIncremental) Compilation failed` (from travis). Would you be able to fix it?

Ferada:
> Though compilation fails in one of the modules: `[error] (asyncHttpClientBackend / Compile / compileIncremental) Compilation failed` (from travis). Would you be able to fix it?

Yes I will (soon).

Ferada:
@adamw thanks and for the typo fix too, somehow the second commit didn't even trigger Travis.

adamw:
Released in 1.5.11

Issue #182, opened 2019-02-08
POST / REDIRECT / GET


Ferada:
I have an API here that follows the usual pattern of requiring a POST, then it will redirect you to a different URL. With `followRedirects` on, `sttp` follows the redirect, but then does the POST again, which then (thankfully not silently in this case) caused a 501 "Not Implemented" output.

Edit: [Similar curl issue](https://github.com/curl/curl/issues/578) - so I suppose it's something where a reasonable default behaviour would be nice that can be used instead of implementing this yourself? Or, if not, maybe a small note in the docstring of the option then?

adamw:
Interesting problem :) 

Looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections, it seems that handling POST 301/302 redirects is something of a grey area - some clients use GET for the redirect, some use POST, some do nothing.

So it seems we should somehow support 301/302 POST redirects - through an additional option? Always using GET? Always preserving the method? Doing nothing?

(btw. there are also 307/308 redirects where the method should always be preserved)

Ferada:
Yes I think it'd also be completely okay to keep it as it is, with a warning attached to remind people that indeed the `POST` will be attempted again at the new location.

However, that'd mean, since there's no general purpose "follow redirects" wrapper, that users will have to keep writing code for that in case they just need behaviour like a (typical) browser / command line tool.

I could see a new flag like `redo(Action)AfterRedirect` (or even more specific, `useGetAfterRedirect`) that would keep the current behaviour when unchanged (backwards compatibility) and that would, when set to a different value, not issue a `POST`/`PUT`/... action, but a `GET` instead. I'm not sure that covers all bases though.

About the different codes ... beats me, I'd go for least surprise, i.e. match `curl`/`wget`'s behaviour :)

adamw:
1.5.10 contains a new option, `redirectToGet` (defaults to `false`), which should cause POSTs to be redirected to GETs. Maybe in the next major release we can consider swapping the default.

Could you test it?

Ferada:
Splendid, yes, I will.

PR #198, opened 2019-04-26
Filter out content header when redirecting to GET

satabin:
In #182, the possibility to redirect POST and PUT to a GET has been
introduced. When redirecting the content is dropped. However, if the
content headers are not cleaned, in particuler the `Content-Length` one,
the underlying backend might hang, waiting for body data to be
available. Because no content is sent in the redirect, it is safe and
better to remove all content related headers in the redirected request.


adamw:
Released in 1.5.13, thanks!

PR #199, opened 2019-04-26
Add usingConfigBuilder for Async HTTP Client backend

ybasket:
Sometimes one wants to change one single option in the underlying client, but with what sttp currently offers one can only create the complete config oneself (which means manually setting all the things sttp usually gets from `SttpBackendOptions`) - this PR allows to modify the config after it was initialised by sttp, making custom setups easier and less boilerplaty.


adamw:
Released in 1.5.14, thanks!

PR #203, opened 2019-04-30
Adding sttp MonadError instance for cats MonadError

matwojcik:
There is `AsyncMonadAsyncError`, which adds `MonadAsyncError` for any `Async[F]` instance, but there was no sync `MonadError` instance basing on cats `MonadError`, so I've added `CatsMonadError`, which is then extended by `AsyncMonadAsyncError`

I'd consider also adding implicits into `com.softwaremill.sttp.impl.cats.implicits`, ideally dividing them into cats-like hierarchy, so:
`implicit.all._` - containing both:
`instances.all._` - `MonadAsyncError`, `MonadError` (with some traits hierarchy to proper order them) 
`syntax.all._` - `mapK`

or simpler 
`implicit._` - containing both:
`instances._` - `MonadAsyncError`, `MonadError` (with some traits hierarchy to proper order them) 
`syntax._` - `mapK`

Pls let me know what's your opinion on this MR, and the implicit proposal ;)

### DIFF
--- a/implementations/cats/src/main/scala/com/softwaremill/sttp/impl/cats/AsyncMonadAsyncError.scala
+++ b/implementations/cats/src/main/scala/com/softwaremill/sttp/impl/cats/AsyncMonadAsyncError.scala
@@ -5,22 +5,11 @@ import com.softwaremill.sttp.MonadAsyncError
 
 import scala.language.higherKinds
 
-class AsyncMonadAsyncError[F[_]](implicit F: Async[F]) extends MonadAsyncError[F] {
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! Regarding the implicits - if adding them makes the backend code more readable - sure! I wouldn't go very far in structuring them - as the goal is to keep the instances to an absolute minimum anyway. I wouldn't want sttp to contain a poor copy of cats/scalaz :)

matwojcik:
@adamw I can see 1.5.15 was released after merge of this, but in published jar the changes are not there :o Broken release or something?

adamw:
@matwojcik yes, I merged it during the release, hence the code is not included. Will be in the next one :)

PR #205, opened 2019-05-06
Add implicit MonadError, MonadAsyncError for their cats equivalents

matwojcik:
Added implicits in `com.softwaremill.sttp.impl.cats.implicits` for `MonadError`, `MonadAsyncError` basing on `CatsMonadError` and `AsyncMonadAsyncError`.

This was proposed earlier in previous MR:
https://github.com/softwaremill/sttp/pull/203

The change is breaking, as if someone has already implemented their own instances of sttp `MonadError` and has it in scope. On the other hand there is probably low benefit of having own instance, as it most probably is completely the same.


adamw:
Released in 1.5.16 - thanks!

PR #206, opened 2019-05-06
Curl converter

ghostbuster91:


core/shared/src/main/scala/com/softwaremill/sttp/RequestT.scala
```diff
@@ -244,6 +244,14 @@ case class RequestT[U[_], T, +S](
     backend.send(this.asInstanceOf[RequestT[Id, T, S]])
   }
 
+  def toCurl(implicit isIdInRequest: IsIdInRequest[U]): String = {
+    // we could avoid the asInstanceOf by creating an artificial copy
+    // changing the method & url fields using `isIdInRequest`, but that
+    // would be only to satisfy the type checker, and a needless copy at
+    // runtime.
```
adamw: duplicate comment? ;-)

core/shared/src/main/scala/com/softwaremill/sttp/ToCurlConverter.scala
```diff
@@ -0,0 +1,58 @@
...
+        case (k, v) => s"""-H "$k: $v""""
+      }
+      .mkString(" ")
+  }
+
+  private def extractBody(r: R): String = {
+    r.body match {
+      case StringBody(text, _, _) if r.headers.toMap.get(HeaderNames.ContentType).forall(_ == MediaTypes.Form) =>
+        s"""-F '$text'"""
+      case StringBody(text, _, _) => s"""--data '$text'"""
```
adamw: are there any escaping rules if `text` contains `'`?

core/shared/src/main/scala/com/softwaremill/sttp/ToCurlConverter.scala
```diff
@@ -0,0 +1,58 @@
...
+      case NoBody                 => ""
+    }
+  }
+
+  private def extractOptions(r: R): String = {
+    val sb = mutable.ListBuffer[String]()
+    if (r.options.followRedirects) {
+      sb.append("-L")
+    }
+    sb.append(s"--max-redirs=${r.options.maxRedirects}")
```
adamw: only if follow redirects == true?

core/shared/src/main/scala/com/softwaremill/sttp/ToCurlConverter.scala
```diff
@@ -0,0 +1,58 @@
...
+    sb.mkString(" ")
+  }
+
+  private def addSpaceIfNotEmpty(fInput: R => String): R => String =
+    t => if (fInput(t).isEmpty) "" else s" ${fInput(t)}"
+}
+
+object ToCurlConverter {
+  implicit def requestToCurl[R <: Request[_, _]]: ToCurlConverter[R] = new ToCurlConverter[R]
```
adamw: is this still needed?

adamw:
Maybe some mention in the docs?

ghostbuster91:
Any ideas where to add this mention? Is `->Request->Debugging` a good place?

adamw:
Either there, or maybe as a section in "backends"? It is kind of a manually invoked backed ;-) And definitely another type of interpreter if we treat backedns as interpreters

ghostbuster91:
I like the way of thinking about toCurl converter as of just another type of request interpreter. But that's not the first thing which comes to mind when you think about requests debugging ;)

PR #208, opened 2019-05-10
Make it possible to create Akka Http backends from routes

kubukoz:
@adamw I tried to reuse as much sttp code as possible, so this is what I got as a result.

~~I wanted to get a simpler API for route-based backends (like `AkkaHttpClient.fromRoute(route: Route)`, but the actual `HttpRequest => HttpResponse` code that Akka uses is limited to tests (and I don't think anybody will want to add yet another sttp module to their build).~~ The easiest way to use the new functionality looks like this:

```scala
AkkaHttpBackend.usingClient(
  system,
  http = AkkaHttpClient.fromAsyncHandler(Route.asyncHandler(route))
)
```

~~Let me know if that looks good and I'll work on other things like documentation and moving the client to a new file :)~~ done

### DIFF
--- a/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
@@ -50,9 +50,9 @@ object AkkaHttpClient {
     }
   }
 
-  def fromStrict(run: HttpRequest => HttpResponse): AkkaHttpClient = new AkkaHttpClient {
+  def fromAsyncHandler(run: HttpRequest => Future[HttpResponse]): AkkaHttpClient = new AkkaHttpClient {
     def singleRequest(request: HttpRequest, settings: ConnectionPoolSettings): Future[HttpResponse] =
-      Future.successful(run(request))
+      run(request)
   }
 }
 
@@ -94,8 +94,8 @@ private class AkkaHttpBackend(
     val settings = connectionPoolSettingsWithProxy
       .withUpdatedConnectionSettings(_.withIdleTimeout(r.options.readTimeout))
 
-    requestToAkka(r).flatMap(setBodyOnAkka(r, r.body, _)) match {
-      case Success(request) =>
+    Future.fromTry(requestToAkka(r).flatMap(setBodyOnAkka(r, r.body, _))).flatMap {
+      request =>
         http
           .singleRequest(request, settings)
           .flatMap { hr =>
@@ -121,7 +121,6 @@ private class AkkaHttpBackend(
             body.map(Response(_, code, statusText, headers, Nil))
           }
 
-      case Failure(exception) => Future.failed(exception)
     }
   }
 
--- a/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
@@ -94,34 +94,34 @@ private class AkkaHttpBackend(
     val settings = connectionPoolSettingsWithProxy
       .withUpdatedConnectionSettings(_.withIdleTimeout(r.options.readTimeout))
 
-    Future.fromTry(requestToAkka(r).flatMap(setBodyOnAkka(r, r.body, _))).flatMap {
-      request =>
+    Future
+      .fromTry(requestToAkka(r).flatMap(setBodyOnAkka(r, r.body, _)))
+      .flatMap { request =>
         http
           .singleRequest(request, settings)
-          .flatMap { hr =>
-            val code = hr.status.intValue()
-            val statusText = hr.status.reason()
+      }
+      .flatMap { hr =>
+        val code = hr.status.intValue()
+        val statusText = hr.status.reason()
 
-            val headers = headersFromAkka(hr)
-            val charsetFromHeaders = headers
-              .collectFirst {
-                case (HeaderNames.ContentType, value) => value
-              }
-              .flatMap(encodingFromContentType)
-
-            val responseMetadata = ResponseMetadata(headers, code, statusText)
-            val body = if (r.options.parseResponseIf(responseMetadata)) {
-              bodyFromAkka(r.response, decodeAkkaResponse(hr), charsetFromHeaders, responseMetadata)
-                .map(Right(_))
-            } else {
-              bodyFromAkka(asByteArray, decodeAkkaResponse(hr), charsetFromHeaders, responseMetadata)
-                .map(Left(_))
-            }
-
-            body.map(Response(_, code, statusText, headers, Nil))
+        val headers = headersFromAkka(hr)
+        val charsetFromHeaders = headers
+          .collectFirst {
+            case (HeaderNames.ContentType, value) => value
           }
+          .flatMap(encodingFromContentType)
 
-    }
+        val responseMetadata = ResponseMetadata(headers, code, statusText)
+        val body = if (r.options.parseResponseIf(responseMetadata)) {
+          bodyFromAkka(r.response, decodeAkkaResponse(hr), charsetFromHeaders, responseMetadata)
+            .map(Right(_))
+        } else {
+          bodyFromAkka(asByteArray, decodeAkkaResponse(hr), charsetFromHeaders, responseMetadata)
+            .map(Left(_))
+        }
+
+        body.map(Response(_, code, statusText, headers, Nil))
+      }
   }
 
   override def responseMonad: MonadError[Future] = new FutureMonad()(ec)
@@ -324,7 +324,7 @@ private class AkkaHttpBackend(
 }
 
 object AkkaHttpBackend {
-  private def make(
+  def make(
       actorSystem: ActorSystem,
       ec: ExecutionContext,
       terminateActorSystemOnClose: Boolean,
@@ -379,9 +379,9 @@ object AkkaHttpBackend {
       customHttpsContext: Option[HttpsConnectionContext] = None,
       customConnectionPoolSettings: Option[ConnectionPoolSettings] = None,
       customLog: Option[LoggingAdapter] = None
-  )(
-      http: AkkaHttpClient = AkkaHttpClient.fromAkkaHttpExt(actorSystem, customHttpsContext, customLog)
-  )(implicit ec: ExecutionContext = ExecutionContext.Implicits.global): SttpBackend[Future, Source[ByteString, Any]] =
+  )(implicit ec: ExecutionContext = ExecutionContext.Implicits.global): SttpBackend[Future, Source[ByteString, Any]] = {
+    val http: AkkaHttpClient = AkkaHttpClient.fromAkkaHttpExt(actorSystem, customHttpsContext, customLog)
+
     make(
       actorSystem,
       ec,
@@ -390,4 +390,5 @@ object AkkaHttpBackend {
       customConnectionPoolSettings,
       http
     )
+  }
 }
--- a/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
@@ -10,7 +10,7 @@ import akka.http.scaladsl.model.HttpHeader.ParsingResult
 import akka.http.scaladsl.model.headers.{BasicHttpCredentials, HttpEncodings, `Content-Length`, `Content-Type`}
 import akka.http.scaladsl.model.{Multipart => AkkaMultipart, _}
 import akka.http.scaladsl.settings.ConnectionPoolSettings
-import akka.http.scaladsl.{ClientTransport, Http, HttpsConnectionContext}
+import akka.http.scaladsl.{ClientTransport, HttpsConnectionContext}
 import akka.stream.ActorMaterializer
 import akka.stream.scaladsl.{FileIO, Sink, Source, StreamConverters}
 import akka.util.ByteString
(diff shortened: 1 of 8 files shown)


akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -21,14 +21,48 @@ import scala.collection.immutable.Seq
 import scala.concurrent.{ExecutionContext, Future}
 import scala.util.{Failure, Success, Try}
 
-class AkkaHttpBackend private (
+trait AkkaHttpClient {
```
kubukoz: This will end up in a new file

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -303,24 +330,22 @@ class AkkaHttpBackend private (
 }
 
 object AkkaHttpBackend {
-  private def apply(
+  private def make(
```
kubukoz: The compiler was complaining about overloads with similar shapes...

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -358,14 +385,15 @@ object AkkaHttpBackend {
       customHttpsContext: Option[HttpsConnectionContext] = None,
       customConnectionPoolSettings: Option[ConnectionPoolSettings] = None,
       customLog: Option[LoggingAdapter] = None
+  )(
+      http: AkkaHttpClient = AkkaHttpClient.fromAkkaHttpExt(actorSystem, customHttpsContext, customLog)
```
kubukoz: Maybe we should just use `fromAkkaHttpExt(actorSystem, None, None)` here and not take parameters for `customHttpsContext` and `customLog`? If somebody wants to use custom ones they can provide an AkkaHttpClient with changed parameters themselves.

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -358,14 +385,15 @@ object AkkaHttpBackend {
...
   )(implicit ec: ExecutionContext = ExecutionContext.Implicits.global): SttpBackend[Future, Source[ByteString, Any]] =
```
kubukoz: To be honest I wanted to remove this default. Defaults don't play well with implicits (in terms of understandability) :)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -358,14 +385,15 @@ object AkkaHttpBackend {
...
-    AkkaHttpBackend(
+    make(
       actorSystem,
       ec,
       terminateActorSystemOnClose = false,
```
kubukoz: LMK what you think about solving this by having a new backend class that wraps over an existing backend but also executes some shutdown hooks. This will allow us to completely hide ActorSystem from AkkaHttpBackend (only Materializer and Config will be needed)

build.sbt
```diff
@@ -306,7 +306,9 @@ lazy val akkaHttpBackend: Project = (project in file("akka-http-backend"))
       akkaHttp,
       // provided as we don't want to create a transitive dependency on a specific streams version,
       // just as akka-http doesn't
-      akkaStreams % "provided"
+      akkaStreams % "provided",
+      "com.typesafe.akka" %% "akka-stream-testkit" % "2.5.22" % Test,
```
kubukoz: TODO: extract vals with versions and reuse them here and in the core deps

implementations/monix/js/src/main/scala/com/softwaremill/sttp/impl/monix/FetchMonixBackend.scala
```diff
@@ -61,7 +62,7 @@ class FetchMonixBackend private (fetchOptions: FetchOptions)
             }
           }
         }
-        go().doOnSubscriptionCancel(Task(reader.cancel("Response body reader cancelled")))
+        go().doOnSubscriptionCancel(Task(reader.cancel("Response body reader cancelled")).void)
```
kubukoz: This is just to remove a compilation warning (non-unit value discard)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -291,7 +317,8 @@ class AkkaHttpBackend private (
   }
 
   override def close(): Unit = {
-    if (terminateActorSystemOnClose) actorSystem.terminate()
+    val _ = if (terminateActorSystemOnClose) actorSystem.terminate()
```
kubukoz: Workaround for value discard warning

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -21,14 +21,48 @@ import scala.collection.immutable.Seq
...
+        connectionContext.getOrElse(http.defaultClientHttpsContext),
+        settings,
+        customLog.getOrElse(system.log)
+      )
+    }
+  }
+
+  def fromStrict(run: HttpRequest => HttpResponse): AkkaHttpClient = new AkkaHttpClient {
+    def singleRequest(request: HttpRequest, settings: ConnectionPoolSettings): Future[HttpResponse] =
+      Future.successful(run(request))
```
adamw: what about exceptions thrown by `run`? :)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
adamw: ok

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
```diff
@@ -62,26 +94,20 @@ class AkkaHttpBackend private (
...
         val code = hr.status.intValue()
         val statusText = hr.status.reason()
 
         val headers = headersFromAkka(hr)
         val charsetFromHeaders = headers
-          .find(_._1 == HeaderNames.ContentType)
-          .map(_._2)
+          .collectFirst {
+            case (HeaderNames.ContentType, value) => value
+          }
```
adamw: 👍

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
adamw: yeah, happens :)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
adamw: I would prefer not to change the API, or at least keep it backwards-(source)-compatible. This change might wait until sttp 2 (when it comes ;) ). So I would keep the signatures of existing methods intact, and add a new variant which takes an `AkkaHttpClient`. The others might delegate to that particular variant.

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
adamw: well, see above :)

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
adamw: Not sure if I see the benefit - apart from one field less in the backend? Would that "raw" backend be somehow reusable in other contexts?

adamw:
Looks good! Thanks :)

adamw:
Ah, probably some mention in the docs would be useful :)

kubukoz:
Sure, that's the next step ;) I'm currently fighting with akka-http because there's apparently a bug in handling some routes... will report back.

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
kubukoz: I assume it's pure... but of course we can loosen that assumption (or accept `HttpRequest => Try[HttpResponse]`).

In fact, I think the `fromStrict` function will be obsolete anyway, as the example I posted doesn't work for routes that don't immediately return a response (due to how `.response` works). `check` and a `Promise` will be required. I'll tell more when I'm able to run any of my tests with this :D

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
kubukoz: Probably not ;) just using a less powerful abstraction.

kubukoz:
Turns out the bug is in sttp. I added a few test cases to the new test, two of them fail (and judging by the fact that directly calling the endpoints with akka-http's testkit syntax doesn't make them fail, I think it's sttp's fault). I'll see if I can find the solution to that.

Basically it looks like sttp doesn't forward the response if an asynchronous boundary is introduced in the future that the directive waits for.

kubukoz:
Oops, my bad. It's actually something wrong in akka-http.

kubukoz:
I got the two pairs of tests nondeterministically fail together, so it seems like the failures are indeed caused by akka-http's behavior. I'll try to reproduce the bug there.

kubukoz:
It seems like part of the issue is that akka-http behaves weirdly if you run the request after an asynchronous boundary.

A workaround involved:
- not using `(request ~> route).response`, but using the full `request ~> route ~> check {...}` chain and using a Promise to capture the completed route's response
- a small change in the sttp backend that removes that boundary (which was by not using `Future.fromTry`, instead we have to manually lift the failure and pass the success value to the rest of the computation).

I'm not sure if that'll be the end of it though, as some user tests might already be running on some different thread, and the akka-http testkit might not work again...

akka-http-backend/src/test/scala/com/softwaremill/sttp/akkahttp/AkkaHttpRouteBackendTest.scala
```diff
@@ -0,0 +1,107 @@
...
+  implicit val timeout = RouteTestTimeout(5.seconds)
+
+  lazy val backend: SttpBackend[Future, Nothing] = {
+    AkkaHttpBackend.usingActorSystem(system) {
+      new AkkaHttpClient {
+        override def singleRequest(request: HttpRequest, settings: ConnectionPoolSettings): Future[HttpResponse] = {
+          val p = Promise[HttpResponse]()
+          val f = p.future
+
+          request ~> Route.seal(Routes.route) ~> check {
```
jrudolph: That's actually not how you would create a handler function from a route. This API is only used as a testing API.

The straight-forward way is using `Route.asyncHandler`:

https://github.com/akka/akka-http/blob/f57ec755a9107a8c1c25412ecde3149a91261e7c/akka-http/src/main/scala/akka/http/scaladsl/server/Route.scala#L71-L78

kubukoz: This is great! Thanks.

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
kubukoz: OK, now I have a benefit - no need to start an actor system at all to create the testing backend...

akka-http-backend/src/main/scala/com/softwaremill/sttp/akkahttp/AkkaHttpBackend.scala
kubukoz: Can we make `make` public then? Otherwise we'll need to duplicate both variants (the one that takes an ActorSystem and one that creates it itself).

kubukoz:
I think we're ready for review @adamw

docs/backends/akkahttp.rst
```diff
@@ -52,3 +52,14 @@ To receive the response body as a stream::
...
+
+Testing
+---------------
+
+For testing, you can create a backend using any `HttpRequest => Future[HttpResponse]` function - including a `Route`.
+
+That way, you can "mock" a server that the backend will talk to, without starting any actual server or making any HTTP calls.
+
+  val backend: SttpBackend[Future, Nothing] = {
+    AkkaHttpBackend.usingClient(system, http = AkkaHttpClient.fromAsyncHandler(Route.asyncHandler(Routes.route)))
```
adamw: maybe we could have `AkkaHttpBackend.usingClient(system, AkkaHttpClient.fromRoute(Route))`? I think this will be the most common use-case.

Or even `AkkaHttpBackend.usingRoute(route)`?

I'm also thinking about names ... the route-based client is in fact a stub client -an alternative to `SttbBackendStub`. And that's when it will be used. Maybe the names should somehow reflect that? Just so that the reader of the code knows what this is about.

So maybe: `AkkaHttpClient.stubFromRoute` / `AkkaHttpClient.stubFromAsyncHandler`?

kubukoz: I thought about that, but `usingRoute` will have like 6 implicit parameters (with defaults), same as `Route.asyncHandler`.

+1 for the stub naming

adamw:
Thanks! Released in 1.5.17

PR #245, opened 2019-07-08
make HttpURLConnectionBackend extensible

chergey:
Hi, 
We have been using this library in our tool runs in Hadoop YARN and deals with hadoop security that has custom authentication mechanism and special class for (http) authentication - [see](https://github.com/hopshadoop/hops/blob/5c3f2e501840fa123e672b24a53ffabfcd85bd95/hadoop-hdfs-project/hadoop-hdfs/src/main/java/org/apache/hadoop/hdfs/web/URLConnectionFactory.java#L149). It would be nice if we could just extend HttpURLConnectionBackend and override openConnection.

### DIFF
--- a/core/jvm/src/main/scala/com/softwaremill/sttp/HttpURLConnectionBackend.scala
+++ b/core/jvm/src/main/scala/com/softwaremill/sttp/HttpURLConnectionBackend.scala
@@ -56,7 +56,7 @@ class HttpURLConnectionBackend protected (opts: SttpBackendOptions, customizeCon
 
   override val responseMonad: MonadError[Id] = IdMonad
 
-  protected def openConnection(uri: Uri): HttpURLConnection = {
+  private def openConnection(uri: Uri): HttpURLConnection = {
     val url = new URL(uri.toString)
     val conn = opts.proxy match {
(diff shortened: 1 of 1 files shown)


chergey:
@adamw Can you look ?

adamw:
Making the class protected is one option; but maybe what you would in fact need is a way to customise the `URL`? The same way as there's a hook to customise the `HttpURLConnection`. 

Instead of a hard-coded `val url = new URL(uri.toString)`, this could be `val url = createURL(uri)`, where the default implementation is `new URL(_)`. Would that work for you?

adamw:
@chergey ^

adamw:
Would this approach: https://github.com/softwaremill/sttp/commit/00892b3ddf5bd91dbee65ededf2864dd9f97f58c work for you as well? The idea is basically to pass parameters instead of overloading classes.

adamw:
@chergey ^ :)

chergey:
@adamw 
Thanks for the new openConnection option. Just as I was going to commit it to my branch, I saw your commit.

adamw:
Released in 1.6.2

PR #247, opened 2019-07-08
more correct Scala version number handling in build

SethTisue:
this works with nightly builds of Scala, and thus is needed for the Scala
community builds


adamw:
Thanks!

PR #257, opened 2019-08-09
bodyEnc method, typed charsets

fsanaulla:


core/shared/src/main/scala/com/softwaremill/sttp/Response.scala
```diff
@@ -23,11 +25,15 @@ case class Response[T](
 ) extends ResponseExtensions[T]
     with ResponseMetadata {
 
-  lazy val body: Either[String, T] = rawErrorBody match {
+  lazy val body: Either[String, T] =
+    bodyEnc(StandardCharsets.UTF_8)
+
+  def bodyEnc(ch: Charset): Either[String, T] = rawErrorBody match {
```
ybasket: Maybe use a non-abbreviated name which is a bit clearer?

Also, I wonder whether it would make sense to clarify the semantics and cover more use cases by introducing an additional `force` parameter (with default `false`):
* If `false`, behave as you have it right now, respecting the charset from headers and falling back to `ch`
* If `true`, ignore charset info from headers and decode using `ch`

fsanaulla: Make sense. 

> Maybe use a non-abbreviated name which is a bit clearer?

Maybe you can provide some concrete examples?

> If true, ignore charset info from headers and decode using ch

ybasket: I'm not sure what examples you ask for…name could be something like `bodyUsingEncoding(charset: Charset, force: Boolean = false)`. 

An example for `force` usage would a response with `Content-Type: text/plain; charset=ISO-8859-1` which you want to decode using `UTF-8` because you know it's set wrong by the server.

adamw: An alternative to consider: add `contentType`/`header` method to `Response`, as in `RequestT`, which would allow changing the given response object. Then, you could modify the content type as needed, and call `.body` as now.

The downside is that I'm not sure if it makes any sense apart from the above scenario, where you want to read the body using a different encoding. So might be confusing in the end, as you don't really build the responses. Or maybe you do, in tests, and that would be useful as well?

adamw:
This is fixed in 2.0.0-M5 using the new response specifications (see https://blog.softwaremill.com/sttp2-an-overview-of-proposed-changes-8de23c94684f)

adamw:
Thanks for the PR anyway, though :)

PR #263, opened 2019-08-15
Add Uri.parse

ybasket:
Useful for some cases, for example when implementing a pureconfig reader for `Uri`s


adamw:
Is this different/better from `uri"$myUriAsString"`?

ybasket:
Good catch. Changed it to use the string interpolator as that makes more sense in terms of consistency.

The value of this is that it can be eta-expanded and makes the possibility of failures visible. So when parsing a `Uri` inside something (like pureconfig's `ConfigReader` or circe's `Decoder`) which has `emapTry` or similar, you can just write `.emapTry(Uri.parse)`.

adamw:
Released in 1.6.5, thanks!

PR #275, opened 2019-09-06
[No need to merge] An example to use sbt-sonatype 3.5

xerial:
A follow-up for https://github.com/xerial/sbt-sonatype/issues/93


adamw:
Thanks! I made the changes here instead: https://github.com/softwaremill/sbt-softwaremill

PR #283, opened 2019-09-11
Upgrade to cats 2.0.0 and monix 3.0.0

leonardehrenfried:
I have not been able to fully figure out how the build process works, but here is my first stab at this.

Is `.settings(commonJvmJsSettings)` the right set of settings?


adamw:
Looks good, thanks! :)

adamw:
Yeah the build can be confusing, but that's what you end up with supporting a mixture of jvm/js/native and 2.11/2.12/2.13 :) 

As monix now supports 2.13, it's enough to just drop that line

PR #301, opened 2019-09-27
Accumulate multiple cookies into single header

### LINKED ISSUE #279: Multiple Cookie headers sent in HTTP request
Per each invocation of `sttp.cookie` a `Cookie` header is added. Having multiple invocations can be useful if you are sharing a common request of some sort but want to add a certain cookie to a specific request.

RFC 6265 says that

> When the user agent generates an HTTP request, the user agent MUST NOT attach more than one Cookie header field.

ghostbuster91:
This fixes #279


adamw:
Released in 2.0.0-M3, thanks!

Issue #306, opened 2019-10-02
Allow configuration of URI path encoding.


jonathanhood:
I'm attempting to deal with URIs that have some special requirements (taken from [RFC8040 Section 3.5.3](https://tools.ietf.org/html/rfc8040#section-3.5.3)) for how encoding is performed - particularly with regards to how `,` is handled. As it stands, trying to handle the special case in that RFC "on top" of the existing `Uri` structure in sttp results in double-encoding since sttp, itself, wants to perform the encoding and is unaware of these special considerations.

Now, asking sttp to become aware of these particular constraints of my admittedly uncommon use case is a bit much. Instead, I'd like some way to inject some configuration (similiar, perhaps, to how `QueryFragmentEncoding` is handled) that allows me to at least disable or modify sttp's built in encoding.

An example of the configuration structure:

```scala
sealed trait PathEncoding
object PathEncoding {
  // (default) Automatically encode all path segments
  case object All extends PathEncoding

  // Disable segment encoding and assume provided segments
  // are already properly encoded.
  case object Disabled extends PathEncoding
}
```

This would require changes in encoding (during ``toString``) similiar to how query parameter encoding is already performed based this configuration provided during ``Uri`` construction.

jonathanhood:
It looks like this request is, perhaps, similar (at least in proposed solution) to #287.

aappddeevv:
I think that escaping should be an instance provided to a constructor in some way so escaping can be completely customized. Also adding something like @adamw mentioned using a Literal class would also be a good thing as well for simple cases.

@adamw, what do you think? I'd like to move off a forked version of sttp. Literal would take care of the immediate need and seems useful in the long run anyway.

jonathanhood:
I suppose the paths could change to be a sequence of some kind of abstract type, e.g.:

```scala
sealed abstract trait PathSegment
case class Literal(seg: String) extendsPathSegment {
  override def toString: String = seg
}
case class Escaped(seg: String) extends PathSegment {
  def encode(x: String): String = ???
  override def toString: String = encode(seg)
}


case class Uri(path: List[PathSegment], ...) {
   // somewhere in here...
  val encodedPath: String = path.map(_.toString).mkString("/")
}
```

However, that would at least break compatibility with existing API users. Thoughts on that compatibility issue?

adamw:
[...]

PR #309, opened 2019-10-03
Allow path segment encoding to be disabled by configuration.

jonathanhood:
This PR applies the solution originally suggested in #306 for inclusion in a 1.x release of sttp. It adds some simple configuration to `Uri` which allows callers to disable automatic encoding of path segments entirely.

### DIFF
--- a/core/shared/src/main/scala/com/softwaremill/sttp/Uri.scala
+++ b/core/shared/src/main/scala/com/softwaremill/sttp/Uri.scala
@@ -2,7 +2,7 @@ package com.softwaremill.sttp
 
 import java.net.URI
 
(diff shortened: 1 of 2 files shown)


adamw:
Released in 1.7.0, thanks!

jonathanhood:
Thanks for the quick feedback on this one!

aappddeevv:
I'll have to check. If its only turning off escaping in the path segments then I don't think it will. The issue is escaping across all of the Uri, path segments and all other parts. But I can't remember the code I wrote to get around this so I need to go look at it. I think that's why I needed a flag that turns it all off everywhere in the .toString method.

PR #310, opened 2019-10-03
sttp2: Allow path segment encoding to be customized per-segment.

jonathanhood:
This PR provides similar functionality to #309 which also attempts to address #306. This approach provides a better API for that feature but is also not compatible with how paths were handled in sttp 1.x releases.


adamw:
Implemented differently using path/query/fragment segments, see #306. Thanks for the PR anyway, though :)

PR #319, opened 2019-10-13
Make type signature of AsyncHttpClientZioBackend less scary

williamhaw:
# Problem
When Intellij automatically adds the type annotation for this line of code: 
`implicit val backend = AsyncHttpClientZioBackend()`
it becomes the following:
`implicit val backend: SttpBackend[({type λ[β$3$] = ZIO[Any, Throwable, β$3$]})#λ, Nothing] = AsyncHttpClientZioBackend()`.

# Why this PR is needed
This PR simplifies the type signature of the public API of AsyncHttpClientZioBackend, making it easier for people new to ZIO (for e.g. me) to understand how to use it with the sttp client.

# Changes
Change all signatures in AsyncHttpClientZioBackend from `zio.IO[Throwable, ?]` to `zio.Task`

# Notes
This PR should be backwards compatible as an `IO[Throwable, ?]` **is** a `Task`. Nonetheless I have manually tested that declaring 
`implicit val backend: SttpBackend[Task, Nothing] = AsyncHttpClientZioBackend()`
works on v1.7.1 and on this PR.


adamw:
Released in 1.7.2, thanks! :)

PR #327, opened 2019-10-18
Strip sensitive headers during redirect

### LINKED ISSUE #326: Strip sensitive headers during redirect
This involves changing `FollowRedirectsBackend`. Something similar to https://github.com/http4s/http4s/blob/master/client/src/main/scala/org/http4s/client/middleware/FollowRedirect.scala

ghostbuster91:
This closes #326

### DIFF
--- a/core/shared/src/main/scala/sttp/client/FollowRedirectsBackend.scala
+++ b/core/shared/src/main/scala/sttp/client/FollowRedirectsBackend.scala
@@ -2,10 +2,9 @@ package sttp.client
 
 import java.net.URI
 
-import sttp.model._
 import sttp.client.monad.MonadError
 import sttp.client.ws.WebSocketResponse
-import sttp.model.{Method, StatusCode}
+import sttp.model.{Method, StatusCode, _}
 
 import scala.language.higherKinds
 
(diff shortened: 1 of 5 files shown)


core/shared/src/main/scala/sttp/client/FollowRedirectsBackend.scala
```diff
@@ -74,6 +82,7 @@ class FollowRedirectsBackend[F[_], S, WS_HANDLER[_]](delegate: SttpBackend[F, S,
   }
 
   private val contentHeaders = Set(HeaderNames.ContentLength, HeaderNames.ContentType, HeaderNames.ContentMd5)
+  private val sensitiveHeaders = Set(HeaderNames.Authorization, HeaderNames.Cookie, HeaderNames.SetCookie)
```
adamw: maybe these could be a parameter of the class, with a default value as above?

core/shared/src/main/scala/sttp/client/FollowRedirectsBackend.scala
```diff
@@ -65,7 +64,16 @@ class FollowRedirectsBackend[F[_], S, WS_HANDLER[_]](delegate: SttpBackend[F, S,
     }
 
     val redirectResponse =
-      sendWithCounter(changePostPutToGet(request.copy[Identity, T, S](uri = uri), response.code), redirects + 1)
+      sendWithCounter(
+        changePostPutToGet(
+          request.copy[Identity, T, S](
```
adamw: this could be extracted to a method, just like `changePostPutToGet`

core/shared/src/main/scala/sttp/client/FollowRedirectsBackend.scala
ghostbuster91: I also though about it but was not sure. The fact that you are mentioning it convinces me.

adamw:
Thanks! :)

PR #328, opened 2019-10-19
Use the most recent Scala versions (2.13.1 and 2.12.10)

taryk:
Update Scala
from 2.13.0 to 2.13.1
and from 2.12.9 to 2.12.10


adamw:
Thanks!

PR #330, opened 2019-10-22
Java11 backend

### LINKED ISSUE #325: Java11 backend
There's a new API for making HTTP requests. We could wrap it in a backend, which could be used instead of `HTTPUrlConnection`

ghostbuster91:
This closes #325


build.sbt
```diff
@@ -555,3 +556,12 @@ lazy val prometheusBackend: Project = (project in file("metrics/prometheus-backe
...
   .dependsOn(coreJVM)
+
+//-- http4s
+lazy val java11backend: Project = (project in file("java11-backend"))
+  .settings(commonJvmSettings: _*)
+  .settings(
+    name := "java11-backend",
+    scalacOptions ++= Seq("-J--add-modules", "-Jjava.net.http")
+  )
+  .dependsOn(catsJVM, coreJVM % compileAndTest)
```
adamw: this module needs to be enabled only if the host-java is >= 11. Otherwise we need to skip it. Also, is it available for all Scala versions?

adamw: + on travis we have java 8. Is it possible to run a single test with another java?

java11-backend/src/main/scala/sttp/client/httpclient/HttpClientAsyncBackend.scala
```diff
@@ -0,0 +1,87 @@
...
+
+      client
+        .sendAsync(jRequest, BodyHandlers.ofByteArray())
+        .whenComplete(new BiConsumer[HttpResponse[Array[Byte]], Throwable] {
+          override def accept(t: HttpResponse[Array[Byte]], u: Throwable): Unit = {
+            if (t != null) {
+              try success(readResponse(t, request.response))
+              catch { case e: Exception => error(e) }
+            }
+            if (u != null) {
```
adamw: what if both are null? :)

java11-backend/src/main/scala/sttp/client/httpclient/HttpClientBackend.scala
```diff
@@ -0,0 +1,186 @@
...
+  private[httpclient] def convertRequest[T](request: Request[T, S]) = {
+    val builder = HttpRequest
+      .newBuilder()
+      .uri(request.uri.toJavaUri)
+
+    builder.method(request.method.method, bodyToHttpBody(request, builder))
+    request.headers
+      .filterNot(_.name == HeaderNames.ContentLength)
+      .foreach(h => builder.header(h.name, h.value))
+    builder.timeout(JDuration.ofMillis(request.options.readTimeout.toMillis)).build() //TODO is this correct?
```
adamw: is it?

adamw: assuming the tests pass - then it is :)

java11-backend/src/main/scala/sttp/client/httpclient/HttpClientBackend.scala
```diff
@@ -0,0 +1,186 @@
...
+    val headers = res
+      .headers()
+      .map()
+      .keySet()
+      .asScala
+      .flatMap(name => res.headers().map().asScala(name).asScala.map(Header(name, _)))
+      .toList
+
+    val code = StatusCode(res.statusCode())
+    val message = "???" // TODO
```
adamw: todo :)

ghostbuster91: Yes :) I don't know how to extract this message from the response

java11-backend/src/main/scala/sttp/client/httpclient/HttpClientAsyncBackend.scala
ghostbuster91: They cannot be according to the docs. Should I cover that case also?

adamw: no, that's fine :) great api ;)

PR #335, opened 2019-10-25
Add monix based httpClient backend

ghostbuster91:
- monix based httpClient backend
- httpClient backend request customization
- high level ws interface around monix httpClient websocket with backpressure support
- deduplication of monix related code


docs/backends/httpclient.rst
```diff
@@ -9,6 +9,7 @@ This backend is based on ``java.net.http.HttpClient`` available from Java 11 onw
 
 * a **synchronous** backend: ``HttpClientSyncBackend``
 * an **asynchronous**, ``Future``-based backend: ``HttpClientFutureBackend``
+* an **asynchronous**, ``Monix-Task``-based backend: ``HttpClientMonixBackend`` with streaming support
```
adamw: there's also a summary table in `backends.rst`

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/FlowSubscriberAsMonixSubscriber.scala
```diff
@@ -0,0 +1,191 @@
...
+import monix.execution.{Ack, Cancelable, Scheduler}
+import monix.execution.atomic.Atomic
+import monix.reactive.observers.Subscriber
+
+import scala.annotation.tailrec
+import scala.collection.immutable.Queue
+import scala.concurrent.{Future, Promise}
+
+// based on monix.reactive.Observable.toReactivePublisher
+private[monix] final class FlowSubscriberAsMonixSubscriber[A] private (
```
adamw: that's needed to convert reactive-streams into monix-stream? Why not use the built-in one?

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/HttpClientMonixBackend.scala
```diff
@@ -0,0 +1,80 @@
...
+)(implicit s: Scheduler)
+    extends HttpClientAsyncBackend[Task, Observable[ByteBuffer]](
+      client,
+      TaskMonadAsyncError,
+      closeClient,
+      customizeRequest
+    ) {
+
+  override def send[T](request: Request[T, Observable[ByteBuffer]]): Task[Response[T]] = {
+    super.send(request).guarantee(Task.shift)
```
adamw: we should also shift after `openWebsocket`

maybe, in fact, this can be somehow extracted as a (stackable) trait?

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/internal/NativeWebSocketHandler.scala
```diff
@@ -0,0 +1,124 @@
...
+class AddToQueueListener[F[_]](queue: AsyncQueue[F, WebSocketEvent], isOpen: AtomicBoolean) extends Listener {
+
+  override def onOpen(webSocket: JWebSocket): Unit = {
+    isOpen.set(true)
+    queue.offer(WebSocketEvent.Open())
+    super.onOpen(webSocket)
+  }
+
+  override def onText(webSocket: JWebSocket, data: CharSequence, last: Boolean): CompletionStage[_] = {
+    onFrame(WebSocketFrame.Text(data.toString, last, None))
```
adamw: don't we have to call `webSocket.request(1)`?

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
```diff
@@ -0,0 +1,62 @@
...
+            (a, b) =>
+              (a, b) match {
+                case (Right(acc), Right(f2: WebSocketFrame.Text)) => Right(acc + f2.payload)
+                case _                                            => Left(())
+              }
+          )
+        )
+        .map(payload => payload shouldBe Right(s"echo: test$i"))
+    }
+    fs.foldLeft(Task.now(succeed))(_ >> _)
```
adamw: why didn't the previous impl of `receiveEcho` work? I can't see a major difference :)

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/HttpClientMonixBackend.scala
ghostbuster91: What do you mean by 'stackable' trait? I can see two options: either abstract class or composition. Obviously I prefer the second one.

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/internal/NativeWebSocketHandler.scala
ghostbuster91: No we don't because we only requesting next elements once we poll previous one from the queue. See line 45.

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
ghostbuster91: Because in this case response comes as two messages and we have to concatenate them.

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/HttpClientMonixBackend.scala
adamw: https://www.artima.com/scalazine/articles/stackable_trait_pattern.html

In other words, traits which have `abstract override` methods :)
Not a common Scala construction, but it's there

adamw: But probably that's for another PR
Maybe you can raise an issue to unify that?

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/internal/NativeWebSocketHandler.scala
adamw: Ok, so the queue will ever at most have 2 elements (open + msg), and "normally" it will have 0 or 1?

If so:
1. There's no sense in accepting a buffer capacity in `sttp.client.httpclient.monix.MonixWebSocketHandler`
2. Maybe for performance reasons it would make sense to have *some* internal buffer of messages, but a relatively small one, e.g. for 128 messages? (and this could be parametrised; but the important part is that we will never be in a situation where we will have a "buffer overflow"). But maybe Java already buffers?

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
adamw: Ah! :) makes sense. I would add it as a comment, though, as it's non-obvious what happens here :)

adamw: Or maybe ... `.receiveText` should already combine the frames? 

Or rather, we could have `.receiveTextFrame` which would return a (optionally partial) frame, and `.receiveText` (and symmetrically `.receiveBinary`) which would already concatenate the result and return a `F[String]` (without the meta-data).

The reason here would be that in 99.9% of use-cases you want the concatenated string :)

Concatenating could be a simply recursive function accumulating until you get the final fragment. Wdyt?

ghostbuster91: I like it, but since that change would affect all of the backends  shouldn't it be done as separate PR as well?

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/HttpClientMonixBackend.scala
ghostbuster91: sure, TIL

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
adamw: ok :)

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/FlowReactiveStreamsAdapters.scala
```diff
@@ -0,0 +1,23 @@
+package sttp.client.httpclient.monix
+
+import java.util.concurrent.Flow
+
+import org.reactivestreams.{Publisher, Subscriber, Subscription}
+
+private[monix] final class ReactivePublisherJavaAdapter[T](publisher: Publisher[T]) extends Flow.Publisher[T] {
```
adamw: 👍

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
adamw: will you keep it on your list or maybe create an issue? :)

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/internal/NativeWebSocketHandler.scala
ghostbuster91: Generally I agree, I did it like that only because I wasn't sure how to do it and I thought that keeping this code concise with the rest of the code base won't be a bad idea.

> Ok, so the queue will ever at most have 2 elements (open + msg), and "normally" it will have 0 or 1?

yes

> There's no sense in accepting a buffer capacity in sttp.client.httpclient.monix.MonixWebSocketHandler

yes

> Maybe for performance reasons it would make sense to have some internal buffer of messages, but a relatively small one, e.g. for 128 messages? 

I don't get this. What kind of performance reasons do you have in mind? And what do you mean by internal buffer? Something like current monixQueue but not exposed to the outside world?

> But maybe Java already buffers?

U mean, that this backpressure in java ws implementation implies some internal buffer existence?

adamw: Well, what happens when we don't call `jwebsocket.receive`? I suspect that  the implementation doesn't read from the socket. This causes unreceived data to be buffered in TCP sockets, and then TCP backpressure kicks in and signals the sender that no more data should be sent (as there's unreceived data on the client side). But that's only a guess, plus I don't know websockets/TCP THAT well :)

If it indeed worked this way, it could have consequences when receiving messages: once you call receive, the TCP buffer is freed up (because we receive the message), and subsequent messages can now be received by the connection. This means that, to receive more messages than fits in the TCP buffer (+ the hypothetical app-level buffer that java-websockets have), we need to wait on the network to actually send the data. Which makes the whole process slow.

It could be faster if we buffer up enough messages on the app side not to wait for the network unless we consume fast.

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/MonixWebsocketHandlerTest.scala
ghostbuster91: #338

PR #336, opened 2019-10-27
Add websocket support to async-http-client fs2 backend

ybasket:
Adds an implementation of `AsyncQueue` based on what fs2 offers so that users of the fs2 backend can use WebSocket support as well.

Once this PR is merged, I'll try to add a handler which takes a `fs2.Pipe` to handle incoming and outgoing messages on an even higher level than the current API offers.

### DIFF
--- a/async-http-client-backend/fs2/src/main/scala/sttp/client/asynchttpclient/fs2/Fs2WebSocketHandler.scala
+++ b/async-http-client-backend/fs2/src/main/scala/sttp/client/asynchttpclient/fs2/Fs2WebSocketHandler.scala
@@ -13,31 +13,38 @@ object Fs2WebSocketHandler {
   private class Fs2AsyncQueue[F[_], A](queue: InspectableQueue[F, A])(implicit F: Effect[F]) extends AsyncQueue[F, A] {
     override def clear(): Unit =
       F.runAsync(queue.getSize.flatMap { size =>
-        queue.dequeue.take(size.toLong).compile.drain
-      })(IO.fromEither).unsafeRunSync()
+          queue.dequeue.take(size.toLong).compile.drain
+        })(IO.fromEither)
+        .unsafeRunSync()
 
     override def offer(t: A): Unit =
       F.runAsync(queue.offer1(t))(IO.fromEither(_).flatMap {
-        case true => IO.unit
-        case false => IO.raiseError(new WebSocketBufferFull())
-      }).unsafeRunSync()
+          case true  => IO.unit
+          case false => IO.raiseError(new WebSocketBufferFull())
+        })
+        .unsafeRunSync()
 
     override def poll: F[A] = queue.dequeue1
   }
+
   /**
     * Creates a new [[WebSocketHandler]] which should be used *once* to send and receive from a single websocket.
     * @param incomingBufferCapacity Should the buffer of incoming websocket events be bounded. If yes, unreceived
     *                               events will some point cause the websocket to error and close. If no, unreceived
     *                               messages will take up all available memory.
     */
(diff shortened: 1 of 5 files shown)


async-http-client-backend/fs2/src/main/scala/sttp/client/asynchttpclient/fs2/Fs2WebSocketHandler.scala
```diff
@@ -0,0 +1,50 @@
...
+import sttp.client.asynchttpclient.WebSocketHandler
+import sttp.client.asynchttpclient.internal.{AsyncQueue, NativeWebSocketHandler}
+import sttp.client.impl.cats.CatsMonadAsyncError
+import sttp.client.ws.{WebSocket, WebSocketEvent}
+import sttp.model.ws.WebSocketBufferFull
+
+object Fs2WebSocketHandler {
+  private class Fs2AsyncQueue[F[_], A](queue: InspectableQueue[F, A])(implicit F: Effect[F]) extends AsyncQueue[F, A] {
+    override def clear(): Unit =
+      F.runAsync(queue.getSize.flatMap { size =>
```
ybasket: This approach is *not * generally safe, but given how it is called and executed in this very special setup, it should work as expected. Feel free to challenge though!

adamw:
Looks good, thanks! :)

adamw:
@ybasket I've been thinking about the semaphore - as far as I understand, it's there to provide mutual exclusion for `clear` and `offer`. However, `clear` is only ever used before an `offer` when called from the AHC websocket listener to clear messages in case of an error. As far as I know, the AHC listener is sequential, so maybe it would make sense to change the `AsyncQueue` contract to specify that `offer` and `clear` are thread-unsafe (can't be called concurrently), and then we won't need the semaphore, right?

ybasket:
@adamw It provides mutual exclusion between `clear` and `poll` (not `offer`, that is safe for the reasons you described). As of my understanding, `poll` is not coupled to the AHC's sequentiality (as the user can always ask to `receive` from the websocket), so we need to make sure that no `poll` happens concurrently while clearing the queue (a task fs2 has no direct support for). If that wasn't ensured, we could have the following:
1. AHC listener calls `clear`, that gets the current size `n` of the queue and streams as many elements to Nirvana
2. While the stream inside `clear` is still running, the user code calls `receive`, therefore `poll` gets called, removing one element
3. the stream of `clear` hangs as it sees only `n-1` elements and no new element will ever be enqueued

So I don't see a way to way to implement `AsyncQueue` based on fs2 safely without a semaphore/mutex/similar. One could of course try to do an implementation backed purely by a `Ref[List[A]` and other cats-effect stuff, but that's probably not very efficient (haven't benchmarked though). Alternatively, the need for clearing the whole queue could be revisited – do you know what was the reason for this design choice? As a user I might be interested in all messages delivered before the error…

adamw:
Ah yes, you are right of course :) yes, `poll` and `clear` can be concurrent

Regarding the clearing: maybe indeed we don't really need `clear`. The reasoning was to fail-fast when there's an error, and let the user know as fast as possible. However, if there's an error, any `send`s will fail anyway, and you might be right that there's no reason to block receives. What do you think?

ybasket:
I'm not sure about `clear`, but I tend to say we should remove it.

Maybe we can achieve configurability, so that the user can choose between both behaviours. The hard part here would be to reflect that in types so that in case the user doesn't want `clear`, we save the semaphore overhead and allow a broader range of queues being used (those without `clear` implementation as every fs2 `Queue` which is not an `InspectableQueue`). I need to have another look at the code and play around to see whether that's feasible.

adamw:
Yes, after some thought I think we should remove this logic and thus `clear` as it won't be needed.

adamw:
Btw. the above has been released in 2.0.0-M8

PR #341, opened 2019-10-29
Create opentracing backend

ghostbuster91:
Should we deprecate brave backend?

### DIFF
--- a/metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
+++ b/metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
@@ -1,0 +1,74 @@
+package sttp.client.opentracing
+
+import io.opentracing.tag.Tags
+import io.opentracing.Tracer
+import io.opentracing.propagation.Format
+import sttp.client.monad.MonadError
+import sttp.client.ws.WebSocketResponse
+import sttp.client.{FollowRedirectsBackend, NothingT, Request, Response, SttpBackend}
+import sttp.client.monad.syntax._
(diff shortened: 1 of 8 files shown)


metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
```diff
@@ -0,0 +1,74 @@
...
+  private implicit val _monad: MonadError[F] = responseMonad
+
+  override def send[T](request: Request[T, S]): F[Response[T]] =
+    responseMonad
+      .eval {
+        tracer
+          .buildSpan(
+            request
+              .tag(OpenTracingBackend.OperationIdRequestTag)
+              .getOrElse("default-operation-id")
```
adamw: maybe this can be configurable? and a function - each time returning a different (by default randomised) id?

adamw:
Docs are missing :)

metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
ghostbuster91: According to this it should not be randomized IMO: 
https://github.com/yurishkuro/opentracing-tutorial/tree/master/java/src/main/java/lesson01#annotate-the-trace-with-tags-and-logs

adamw: ah ok :)

adamw:
Released in 2.0.0-M9, thanks! :)

PR #343, opened 2019-10-29
Close fileOutputStream after using

### LINKED ISSUE #342: FileOutputStream os is not closed in FileHelpers.saveFile
https://github.com/softwaremill/sttp/blob/1387c2d4ff00e452bd00c431980c0ef438c50128/core/native/src/main/scala/sttp/client/FileHelpers.scala#L22

This new FileOutputStream os is not closed after `transfer(is, os)`. In Linux, when I use `Uri.parse(url).map(basicRequest.get).map(_.response(asFile(f)).send())` to download an executable file, and use `Runtime.getRuntime.exec` to run it, I will catch IOException like `"error=26, Text file busy"`.
In my humble opinion, this os should be closed after `transfer(is, os)` .

ghostbuster91:
This closes #342

### DIFF
--- a/core/jvm/src/main/scala/sttp/client/internal/FileHelpers.scala
+++ b/core/jvm/src/main/scala/sttp/client/internal/FileHelpers.scala
@@ -2,6 +2,8 @@ package sttp.client.internal
 
 import java.io.{File, FileOutputStream, InputStream}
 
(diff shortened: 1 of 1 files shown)


adamw:
Thanks :)

PR #348, opened 2019-10-30
Unify monix based backends

### LINKED ISSUE #337: Unify monix based backends
Each high level operation in monix based backends should be shifted to default scheduler. Right now we are doing that manually which isn't obviously good. It should be specified in one place on used in all of the related backends.

https://github.com/softwaremill/sttp/pull/335#discussion_r339590961

jasiak:
This closes #337


jasiak:
> a `AsyncHttpClientMonixBackend`?

Dodane.

adamw:
Thanks! :)

PR #358, opened 2019-11-13
Remove redundant shifts in Monix backends

Avasil:
Hello!

I believe `ShiftToDefaultScheduler` workaround is redundant because `Task.async` and other builders take care of it, see: https://github.com/monix/monix/blob/master/monix-eval/jvm/src/test/scala/monix/eval/TaskAsyncAutoShiftJVMSuite.scala


adamw:
Thanks! Indeed it seems this has changed recently :) Was it in 3.1.0? Is there an issue describing it? I'd update https://blog.softwaremill.com/thread-shifting-in-cats-effect-and-zio-9c184708067b as well then

Avasil:
It has changed in `3.0.0-RC4`: https://github.com/monix/monix/releases/tag/v3.0.0-RC4 

https://github.com/monix/monix/pull/921 and then improved in https://github.com/monix/monix/pull/917

The issue: https://github.com/monix/monix/issues/916 motivated by your blog post :D

Monix has also `executeOn` since `2.x` at least which overrides the default scheduler (like `ZIO.on`) but there was a period pre `3.0.0` when `ContextShift.evalOn` wasnt implemented in terms of it. 

The only remaining issue is existence of `shift(ec)` but I think it's going away in CE3

adamw:
Thanks! I've added a note to the blog :)

PR #362, opened 2019-11-18
Digest auth

### LINKED ISSUE #12: Digest authentication support

ghostbuster91:
This closes #12


core/shared/src/main/scala/sttp/client/RequestT.scala
```diff
@@ -320,6 +320,11 @@ class SpecifyAuthScheme[U[_], T, +S](hn: String, rt: RequestT[U, T, S]) {
 
   def bearer(token: String): RequestT[U, T, S] =
     rt.header(hn, s"Bearer $token")
+
+  def digest(user: String, password: String): RequestT[U, T, S] = {
+    import DigestAuthenticationBackend._
+    rt.digestAuth(user, password)
```
adamw: shouldn't this take into account `hn` - the header name? (whether it's proxy or not)

adamw:
We're just missing the docs on how to use digest authentication :)

core/shared/src/main/scala/sttp/client/RequestT.scala
ghostbuster91: I don't think so, it just sets a tag on request which is later read by `DigestAuthenticatorBackend`

adamw: but shouldn't `request.auth.digest` work differently from `request.proxyAuth.digest`?

ghostbuster91: The proxy path works a little bit different than normal path, but it is all in `DigestAuthenticationBackend`

PR #370, opened 2019-11-26
Fix typo of import.

zhongl:
Missing word "client" in groupId  would cause the problem.


adamw:
Definitely, thanks! :)

PR #373, opened 2019-12-02
Fix duplicate headers in multipart Part

vaslabs:
Following this https://github.com/softwaremill/tapir/pull/335

Just in case you'd like to have these changes.

Thank you.


adamw:
Thanks a lot for the PR! However, I think the current behavior is fine, but needs better documentation.

There are in fact three choices on what can be done when a header is added:
(1) always add it
(2) remove the old one, if a header with the same name exists
(3) remove the new one, if a header with the same name exists

Current behavior is (1) or (2). Your change would mean (1) or (3). Both are valid, but both in the `Part` and in `RequestT` the choice is between (1) and (2) (so the new header is *always* added, with the old one being optionally dropped), and I think it would make sense keeping this behavior.

PR #374, opened 2019-12-02
Finagle backend

ayushworks:


adamw:
👍

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
```diff
@@ -0,0 +1,165 @@
...
+import com.twitter.util
+import sttp.client.internal.FileHelpers
+import sttp.model.{Header, Method, Part, StatusCode}
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.util.Duration
+import scala.io.Source
+
+class FinagleBackend(client: Client) extends SttpBackend[TFuture, Nothing, NothingT] {
+  override def send[T](request: Request[T, Nothing]): TFuture[Response[T]] = {
+    val service = client.newService(s"${request.uri.host}:${request.uri.port.getOrElse(80)}")
```
adamw: shouldn't we take into account `request.uri.scheme` as well?

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
```diff
@@ -0,0 +1,165 @@
...
+    */
+  override def openWebsocket[T, WS_RESULT](request: Request[T, Nothing], handler: NothingT[WS_RESULT]): TFuture[WebSocketResponse[WS_RESULT]] = handler
+
+  override def close(): TFuture[Unit] = TFuture.Done
+
+  /**
+    * The effect wrapper for responses. Allows writing wrapper backends, which map/flatMap over
+    * the return value of [[send]] and [[openWebsocket]].
+    */
+  override def responseMonad: MonadError[TFuture] = new MonadError[TFuture] {
```
adamw: it's also often worth overriding `eval` which can have a more efficient "wrapper-native"  implementation in some cases

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
```diff
@@ -0,0 +1,165 @@
...
+
+
+    rr match {
+      case MappedResponseAs(raw, g) =>
+        fromFinagleResponse(raw, r, meta).map(h => g(h,meta))
+
+      case ResponseAsFromMetadata(f) => fromFinagleResponse(f(meta), r, meta)
+
+      case IgnoreResponse =>
+        TFuture.Done
```
adamw: don't we have to read the response's body content anyway? so that the connection can be later released (that's often the case in web clients)

adamw: or maybe finagle has a way of draining the response's body

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
```diff
@@ -0,0 +1,165 @@
...
+import sttp.model.{Header, Method, Part, StatusCode}
+import com.twitter.finagle.util.DefaultTimer
+import com.twitter.util.Duration
+import scala.io.Source
+
+class FinagleBackend(client: Client) extends SttpBackend[TFuture, Nothing, NothingT] {
+  override def send[T](request: Request[T, Nothing]): TFuture[Response[T]] = {
+    val service = client.newService(s"${request.uri.host}:${request.uri.port.getOrElse(80)}")
+    val finagleRequest = requestBodyToFinagle(request)
+    service.apply(finagleRequest).within(Duration.fromMilliseconds(request.options.readTimeout.toMillis))(DefaultTimer).flatMap{
```
adamw: I suspect `within` will complete the future in case of a timeout - but maybe finagle has some built-in timeout methods? So that the connection can be closed (and not only the future completed, with the other branch still running)

ayushworks: Finagle does have a built in [request timeout](https://twitter.github.io/finagle/docs/com/twitter/finagle/Http$Client.html#withRequestTimeout(timeout:com.twitter.util.Duration):com.twitter.finagle.Http.Client ) , will use that

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
ayushworks: there is no method to drain the body except to set the content buffer to `Empty`. Will change it to that

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
ayushworks: done

finagle-backend/src/main/scala/sttp/client/finagle/FinagleBackend.scala
ayushworks: the scheme would be something like `http` or `https` ?

adamw: yes exactly

ayushworks: ok. so i am checking now the schema to create a client with or without `tls`

adamw:
Released in 2.0.0-RC5. Thanks!

ayushworks:
> Released in 2.0.0-RC5. Thanks!

Thanks @adamw ! I wanted to add some documentation as well - https://github.com/softwaremill/sttp/pull/380

PR #381, opened 2019-12-08
use existing ZIO runtime

seamusv:
Uses the existing Runtime rather than creating a new runtime for each request.


adamw:
Thanks! Didn't know about `ZIO.runtime` :)

PR #404, opened 2020-01-22
HttpClient should not depend on cats

aeons:


adamw:
Thanks! It shouldn't, good catch :)

PR #420, opened 2020-01-30
Make ResponseAs covariant on T

### LINKED ISSUE #408: Make ResponseAs[T, S] covariant on T
I wrote following combinator:
```
def fromStatus[T, S](mapping: (Int, ResponseAs[T, S])*): ResponseAs[T, S] = {
  fromMetadata { meta =>
    val statusCode = meta.code.code
    mapping.toMap.getOrElse(statusCode, sys.error(s"No parser for status $statusCode"))
  }
}
```

And want to use it in following way:
```scala
fromStatus(
  200 -> protoBody[MyModel].map(Right(_)),
  404 -> ignore.map(_ => Left(NotFound()))
)
```

But now i forced to explicitly provide all types explicitly:
```scala
fromStatus(
  200 -> protoBody[MyModel].map(Right(_): Either[NotFound, MyModel]),
  404 -> ignore.map(_ => Left(NotFound()): Either[NotFound, MyModel])
)
```

darl:
closes #408

Make `F[_]` also covariant to avoid unsafe casting [here](https://github.com/softwaremill/sttp/blob/8945531eb8da41055cadcb50b5aba04322217de3/async-http-client-backend/src/main/scala/sttp/client/asynchttpclient/AsyncHttpClientBackend.scala#L200)
and similar places.

### DIFF
--- a/async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
+++ b/async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
@@ -22,7 +22,7 @@ import cats.implicits._
 
 import scala.language.higherKinds
 
-class AsyncHttpClientCatsBackend[F[_]: Async: ContextShift] private (
+class AsyncHttpClientCatsBackend[F[+_]: Async: ContextShift] private (
     asyncHttpClient: AsyncHttpClient,
     closeClient: Boolean,
     customizeRequest: BoundRequestBuilder => BoundRequestBuilder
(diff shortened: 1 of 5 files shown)


async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
```diff
@@ -22,7 +22,7 @@ import cats.implicits._
 
 import scala.language.higherKinds
 
-class AsyncHttpClientCatsBackend[F[_]: Async: ContextShift] private (
+class AsyncHttpClientCatsBackend[F[+_]: Async: ContextShift] private (
```
adamw: Is adding variance here somehow related to the original issue? What kind of problems does it solve?

core/shared/src/main/scala/sttp/client/ResponseAs.scala
```diff
@@ -12,7 +12,7 @@ import scala.util.{Failure, Success, Try}
   * @tparam T Target type as which the response will be read.
   * @tparam S If `T` is a stream, the type of the stream. Otherwise, `Nothing`.
   */
-sealed trait ResponseAs[T, +S] {
+sealed trait ResponseAs[+T, +S] {
```
adamw: Maybe you could add your original example e.g. to `RequestTests` just to know that the code compiles (the compilation is the test :) ). + a comment with the issue's number

adamw:
I've taken a bit of a different approach, without the need to make F[_] covariant as well.

PR #436, opened 2020-02-10
Update circe, try adding circe-js 2.13 back

yoohaemin:
See if updating circe version would fix Scala.js 2.13 test


yoohaemin:
Ok there was at least one case where 2.13.1 was green. Please let me know how I should do with this PR, I can't seem to make it all green.

adamw:
Thanks! Released in 2.0.0-RC10

PR #445, opened 2020-02-21
Fix variadic arglist to enable native backend

### LINKED ISSUE #376: Re-enable the scala-native backend
After updating scala-native to 0.4.0-M2, the `CurlBackend` unfortunately broke.

The update was necessary due to update of scalatest. Here are the changes that I've made to make compilation pass: https://github.com/softwaremill/sttp/commit/5af98b15eab2da9675122fd01ad1d70115f6b8c7

@rwhaling suspects this might be due to changes in varargs handling in 0.4. I've tried changing these, but without any luck.

The backend is disabled for now.

carlosrogue:
Fixes #376  bringing back to life scala native backend...Hopefully! I fixed the invocation of varargs as part of the migration to 0.4.x from 0.3.x as mentioned on native official [documentation](https://scala-native.readthedocs.io/en/latest/user/interop.html#variadic-functions).

Also, I got to know an important part of the problem from https://github.com/scala-native/scala-native-loop/commit/3a164e72de36cc2609ff6d60e02229a9b536280d

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,5 +22,9 @@ project/plugins/project/
 
 core/native/local.sbt
 
+.metals/
+.bloop/
+project/metals.sbt
+
 # scala-native
 lowered.hnir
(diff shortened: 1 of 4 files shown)


adamw:
Released in 2.0.0-RC13 - thanks!

PR #452, opened 2020-03-02
Fix AkkaHttpBackend `isContentLength` filter to inspect header name

simon-lam:
Hi all,

It appears the `isContentLength` function in `AkkaHttpBackend` is inspecting header value and not name.  This causes the `Content-Length` header to be passed down to Akka HTTP which then spits out a loud warning every time a request is funneled through our client:

```
WARN  akka.actor.ActorSystemImpl - Explicitly set HTTP header 'Content-Length: 262' is ignored, explicit `Content-Length` header is not allowed. Use the appropriate HttpEntity subtype.
```

It's innocuous but causes a TON of noise in logs; discovered this issue by way of an upgrade of a Lagom project to version 1.6 (and Akka 2.6 under the covers).


adamw:
Thanks! Released in 2.0.2

PR #473, opened 2020-03-13
[OpenTracingBackend] Add transformation for generated span

eugeniyk:
Currently you can customize generated span only by providing custom operation name.
Typically you want to have ability to adjust span by adding some tags or logs - for example, you may want to specify current retry number in the span tag.

This PR adds generic `Span` transformer tag.

Another thing (authors are welcome to change that as well) - I don't think we need to specify opentracing-sttp-client as backend - you can have nested backends so it doesn't really tell you which backend is under the hood (however that would be useful imo - to know whether you are using java NIO, akka-http or some other backend)


adamw:
Released in 2.0.6, thanks!

PR #483, opened 2020-04-01
Fix h2 calculation when query params are present

ghostbuster91:


adamw:
Released in 2.0.7, thanks! :)

PR #491, opened 2020-04-10
Added toString to ResponseMetadata

matwojcik:
Added toString to ResponseMetadata, as it hugely improves debugging (by default it's serialised as ResponseMetadata@hashcode).

I decided not to include headers, as I was not sure if they wouldn't contain some auth secrets which shall not be logged (however wasn't sure, can change if you want).

### DIFF
--- a/core/shared/src/main/scala/sttp/client/ResponseMetadata.scala
+++ b/core/shared/src/main/scala/sttp/client/ResponseMetadata.scala
@@ -13,6 +13,8 @@ trait ResponseMetadata extends HasHeaders {
   def isRedirect: Boolean = code.isRedirect
   def isClientError: Boolean = code.isClientError
   def isServerError: Boolean = code.isServerError
+
+  override def toString: String = s"ResponseMetadata(code=$code, statusText=$statusText)"
 }
 
 object ResponseMetadata {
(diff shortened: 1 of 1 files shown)


matwojcik:
@adamw pls take a look at this minor ;)

core/shared/src/main/scala/sttp/client/ResponseMetadata.scala
```diff
@@ -13,6 +13,8 @@ trait ResponseMetadata extends HasHeaders {
   def isRedirect: Boolean = code.isRedirect
   def isClientError: Boolean = code.isClientError
   def isServerError: Boolean = code.isServerError
+
+  override def toString: String = s"ResponseMetadata(code=$code, statusText=$statusText)"
```
ghostbuster91, line 17: If we are going to omit headers totally it would be a good idea to say that explicitly to user to avoid confusion. 
```
 override def toString: String = s"ResponseMetadata(code=$code, statusText=$statusText, headers=omitted)"
```
On the other hand, we have these `HeaderNames.sensitiveHeaders`, which we might get advantage of here, but then it won't be customizable.

matwojcik: Good catch, even on `Header` there is `toStringSafe` which I decided to use in the end

adamw:
Released in 2.0.8, thanks!

PR #494, opened 2020-04-11
Allow custom buckets when building histograms for Prometheus backend

eschizoid:
The purpose of this PR is to add support to allow custom buckets when building histograms.

### DIFF
--- a/metrics/prometheus-backend/src/main/scala/sttp/client/prometheus/PrometheusBackend.scala
+++ b/metrics/prometheus-backend/src/main/scala/sttp/client/prometheus/PrometheusBackend.scala
@@ -21,16 +21,16 @@ object PrometheusBackend {
 
   def apply[F[_], S, WS_HANDLER[_]](
       delegate: SttpBackend[F, S, WS_HANDLER],
-      requestToHistogramNameMapper: Request[_, _] => Option[CollectorNameWithLabels] = (_: Request[_, _]) =>
-        Some(CollectorNameWithLabels(DefaultHistogramName)),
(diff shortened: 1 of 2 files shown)


metrics/prometheus-backend/src/main/scala/sttp/client/prometheus/PrometheusBackend.scala
```diff
@@ -184,7 +184,8 @@ class PrometheusListener(
   * Represents the name of a collector, together with label names and values.
   * The same labels must be always returned, and in the same order.
   */
-case class CollectorNameWithLabels(name: String, labels: List[(String, String)] = Nil) {
+case class CollectorNameWithLabels(name: String, labels: List[(String, String)] = Nil, extraBuckets: List[Double] = Nil) {
```
adamw, line 187: I'd change two things here:

1. we should make the whole list of buckets overridable. The default would be as currently (`List(.005, ...)`), but you should be able to specify your custom set
2. the name `CollectorNameWithLabels` becomes incorrect, and in fact only valid for histograms. Maybe we should have a separate `HistogramCollectorConfig` and symmetrically `CollectorConfig` for other types?

eschizoid: @adamw Thanks for reviewing this bad boy. I will take care of both comments today :)

eschizoid: @adamw PR comments addressed, let me know what you think

eschizoid:
any chance to merge this bad boy?

adamw:
Released in 2.0.8, thanks! (with slight changes :) )

PR #500, opened 2020-04-17
Make vals in ops classes private

jhnsmth:
This vals pollute auto-complete scope in IDE. So when you type, this useless vals always pop-up, stealing space from some useful methods.

Also finalize classes for better inlining chances


adamw:
Released in 2.0.8, thanks!

PR #508, opened 2020-04-23
WIP:Add json auth

tampler:
Hello 

I need to support JSON Authorization in headers like this:
`curl -X POST "**LINK***" -H "accept: application/json" -H "Authorization: Basic UUID"`

How do I add `Circe` support to the `sttp.client` ? Can't make work in your sbt. Pls help

@adamw


tampler:
Okay, I found a better option. I need a `Basic` Auth with a token. So I'd better do this:
```scala
def basicToken(token: String): RequestT[U, T, S] =
    rt.header(hn, s"Basic $token")
```

tampler:
Improved solution in #509

PR #509, opened 2020-04-23
WIP: Add basicToken auth

tampler:
Improvement over #508 
```scala
def basicToken(token: String): RequestT[U, T, S] =
    rt.header(hn, s"Basic $token")
```

### DIFF
--- a/core/shared/src/main/scala/sttp/client/RequestT.scala
+++ b/core/shared/src/main/scala/sttp/client/RequestT.scala
@@ -356,6 +356,9 @@ class SpecifyAuthScheme[U[_], T, +S](hn: String, rt: RequestT[U, T, S], digestTa
     rt.header(hn, s"Basic $c")
   }
 
+  def basicToken(token: String): RequestT[U, T, S] =
+    rt.header(hn, s"Basic $token")
+
   def bearer(token: String): RequestT[U, T, S] =
     rt.header(hn, s"Bearer $token")
 


tampler:
@adamw

tampler:
Should this be a separate Auth entity?

adamw:
Released in 2.1.0-RC1

tampler:
Thanks Adam.

Still no cigar... My backend requires `json` serialization on header, but this is not yet supported:
`curl -X POST "mylink" -H "accept: application/json" -H "Authorization: Basic MyUUID"`

Seems like, I need to add `HeaderSerializer`, similar to your `BodySerializer`. I tried the following:
```scala
implicit class jsonReq[F[_], A, +B](req: RequestT[F, A, B]) {
    def header(json: Json): RequestT[F, A, B] = {
      val item = json.as[AuthExec].map(r => (r.action, r.cmd))
      req.header(Header("Authorization", "Basic ae3c8f40-4276-11ea-ab32-f389c0c628ef"))
    }
  }
val exec  = AuthExec("Authorization", "Basic ae3c8f40-4276-11ea-ab32-f389c0c628ef")

  val newReq: RequestT[Empty, String, Nothing] = basicRequest.response(asStringAlways).header(exec.asJson)
```
But that won't work, since it still sends String/String header, instead of JSON.

Any advice on that ? 

Thanks,
Boris

adamw:
Hm why wouldn't this work? You're not using `item` in your example so I'm not sure where's the problem :)

tampler:
Hi Adam

The problem is that I'm still using a `(String, String)` header for auth, which is rejected from our server. Pls see below:

actual request:
```bash
curl -L --max-redirs 32 -X POST -H 'Authorization: Basic UUID' -H 'Content-Type: application/json; charset=UTF-8' --data '{
  "reason" : "bad media",
  "ubuse_id" : "UUID"
}' 'mylink'
```
expected reqeust:
```bash
curl -X POST "mylink" -H "accept: application/json" -H "Authorization: Basic MyUUID"
```
actual server response
```bash
Got response code: 405
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
```

tampler:
Uhhh Sorry, my bad.

All works now, I just missed the correct link !
Thanks for a great task and support Adam!

Do you want to talk on sttp and tapir on our online conference ?

adamw:
Good to hear :)

What's the conference & when is it?

PR #521, opened 2020-04-30
ZIO opentracing - Improvements

Fristi:
- Allows you to use the complete OpenTracing API of ZIO in the `ZioTelemetryOpenTracingTracer` which is by default empy
- The whole operation should be wrapped by `.span` otherwise it will be be part of the parent span


adamw:
Released in 2.1.1, thanks!

Fristi:
Thank you very much ❤️

Issue #515, opened 2020-04-28
SttpJson4s - Removal of DefaultFormats


Executioner1939:
Hi, I would like to know if it's at all possible to remove the default parameter for the `formats` in the [SttpJson4sApi](https://github.com/softwaremill/sttp/blob/e616268c222148bd2e327b5d06310b2b21e7e513/json/json4s/src/main/scala/sttp/client/json4s/SttpJson4sApi.scala) `trait`:

```scala
def asJson[B: Manifest](
      implicit formats: Formats = DefaultFormats,
      serialization: Serialization
  )
```
I feel this causes a lot of confusion, at least for myself and led to me not understanding why my code was failing at runtime.

Our import of our custom `Formats` was accidentally removed, and the `DefaultFormats` were used without any indication. I would personally prefer the compiler letting me know that there are no implicit formats in scope in this case.

I find the `DefaultFormats` to be a good starting point, but don't handle enough types like `UUID`, `java.time.*`, [threeten-extra](https://www.threeten.org/threeten-extra/) and especially serializers for `ADT`'s which is an extremely common pattern in Scala.

I can make this contribution if it makes sense and others agree.

adamw:
I understand the rationale. Though this is a major change, we should probably wait for 2.2. If you'd prepare a PR, I'll keep it around until that time.

Quite possibly, other json integrations might need similar changes, so that the behavior is consitent.

Executioner1939:
Awesome, I can make a PR so long then :-). I am not familiar with the other Json Integrations, will update this so long for Json4s and maybe have a look at the others.

adamw:
Done in 2.2.0

PR #524, opened 2020-05-01
Removed DefaultFormats from Json4s Package

Executioner1939:
Address #515 by removing the default parameter for `formats` in the `SttpJson4sApi` functions.


adamw:
Released in 2.2.0, thanks!

PR #535, opened 2020-05-11
FinagleBackend: use correct default port (443) for https requests.

zamzterz:


adamw:
Released in 2.1.2. Thanks!

PR #554, opened 2020-05-18
Unify custom encoding handling

ghostbuster91:


async-http-client-backend/monix/src/test/scala/sttp/client/asynchttpclient/monix/AsyncHttpClientMonixHttpTest.scala
```diff
@@ -20,4 +20,6 @@ class AsyncHttpClientMonixHttpTest extends HttpTest[Task] with CancelTest[Task,
       .onErrorRecover {
         case _: TimeoutException => None
       }
+
+  override def throwsExceptionOnUnsupportedEncoding = false
```
adamw, line 24: shouldn't this be added to all async-http-client tests?

ghostbuster91: It should, fixed :)

adamw:
Reverted - sorry, but still a lot of tests fail

PR #564, opened 2020-05-22
Fail openWebsocket for JDK HttpClient backends when request fails with 404

LGLO:
Now it's ready for review.
Problem was that we missed handling exception in CompletableFuture and never used callback, because listener was never called

### DIFF
--- a/core/src/test/scalajvm/sttp/client/testing/websocket/HighLevelWebsocketTest.scala
+++ b/core/src/test/scalajvm/sttp/client/testing/websocket/HighLevelWebsocketTest.scala
@@ -1,24 +1,28 @@
 package sttp.client.testing.websocket
 
 import org.scalatest.{Assertion, BeforeAndAfterAll, SuiteMixin}
+import org.scalatest.concurrent.{Signaler, ThreadSignaler, TimeLimits}
+import org.scalatest.flatspec.AsyncFlatSpecLike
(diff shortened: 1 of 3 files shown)


core/src/test/scalajvm/sttp/client/testing/websocket/HighLevelWebsocketTest.scala
```diff
@@ -67,6 +71,20 @@ abstract class HighLevelWebsocketTest[F[_], WS_HANDLER[_]]
       .toFuture()
   }
 
+  it should "fail correctly when can't open WebSocket" in {
+    implicit val signaler: Signaler = ThreadSignaler
+    failAfter(Span(15, Seconds)) {
```
adamw, line 76: what happens when there's no time limit? it hangs? and why 15 seconds, seems a lot ;)

LGLO: It hangs without fix added in this PR.
So if there is such bug again, perhaps in other backend or some regression, this test will fail after 15 seconds.
If implementation is correct this test passes immediately.
15 seconds seems safe for CI.

adamw: 👍

adamw:
Released in 2.1.4, thanks!

PR #565, opened 2020-05-22
Add docs about OAuth2

magdzikk:


docs/requests/authentication.md
```diff
@@ -46,3 +46,22 @@ Also keep in mind that there are some limitations with the current implementatio
...
+
+## OAuth2
+
+You can use sttp with OAuth2. Looking at the [OAuth2 protocol flow](https://tools.ietf.org/html/rfc6749#section-1.2), sttp might be helpful in the second and third step of the process:
+
+1. (A)/(B) - Your UI needs to enable the user to authenticate. Your application will then receive a callback from the authentication server, which will include an authentication code.
+
+2. (C)/(D) - You need to send a request to the authorisation server, passing in the authentication code from step 1. You'll receive an access token in response (and optionally a refresh token). For example, assuming that you're using github as your authentication server, that you've taken the values of `clientId` an `clientSecret` from the github settings, and that you have received `authCode` in step 1 above:
+```scala
+      case class TokenResponse(access_token: String, scope: String, token_type: String, refresh_token: Option[String])
```
adamw, line 58: maybe you can remove the indentation?

docs/requests/authentication.md
```diff
@@ -46,3 +46,22 @@ Also keep in mind that there are some limitations with the current implementatio
...
 * authorizationInfo is not supported
 * scalajs supports only md5 algorithm
+
+## OAuth2
+
+You can use sttp with OAuth2. Looking at the [OAuth2 protocol flow](https://tools.ietf.org/html/rfc6749#section-1.2), sttp might be helpful in the second and third step of the process:
+
+1. (A)/(B) - Your UI needs to enable the user to authenticate. Your application will then receive a callback from the authentication server, which will include an authentication code.
+
+2. (C)/(D) - You need to send a request to the authorisation server, passing in the authentication code from step 1. You'll receive an access token in response (and optionally a refresh token). For example, assuming that you're using github as your authentication server, that you've taken the values of `clientId` an `clientSecret` from the github settings, and that you have received `authCode` in step 1 above:
```
LGLO, line 56: `clientId` an `clientSecret` => `clientId` and `clientSecret` ?
"github settings" are mentioned here, maybe on short sentence above that we are giving GH based example?

magdzikk: That's what I wanted to achieve by starting this sentence with "For example, assuming that you're using github as your authentication server" - but maybe it's still not clear enough 🤔

magdzikk: but maybe I can say that if you're using a different auth server, you'll gave a different url and so on

LGLO: I'm sorry, it's good.
You use snake case in the code and camel case above. AFAIR RFC was using snake, so maybe we could change here as well?

magdzikk: I'm only using snake case where I want circe to work automatically, i.e. when parsing the response from the auth server as `TokenResponse`

PR #576, opened 2020-05-28
[WIP] ChromeDriver should be downloaded automatically 

ghostbuster91:
It shouldn't require any manual setup from developers

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -107,7 +107,7 @@ val browserTestSettings = Seq(
         ) ++ (if (debugging) Seq.empty else Seq("headless"))
         options.addArguments(args: _*)
         val capabilities = org.openqa.selenium.remote.DesiredCapabilities.chrome()
-        capabilities.setCapability("chrome.binary","target/chromedriver");
+        capabilities.setCapability("chrome.binary", "target/chromedriver");
         capabilities.setCapability(org.openqa.selenium.chrome.ChromeOptions.CAPABILITY, options)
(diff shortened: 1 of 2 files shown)


adamw:
Thanks!

build.sbt
```diff
@@ -110,7 +103,37 @@ val browserTestSettings = Seq(
       },
       org.scalajs.jsenv.selenium.SeleniumJSEnv.Config().withKeepAlive(debugging)
     )
-  }
+  },
+  downloadLatestChromeDriver := {
+    if (java.nio.file.Files.notExists(new File("target", "chromedriver").toPath)) {
```
LGLO, line 108: I'm afraid this won't work on Windows, where filename is `chromedriver.exe`

build.sbt
```diff
@@ -491,7 +514,6 @@ lazy val httpClientBackend = (projectMatrix in file("httpclient-backend"))
     }
   )
   .jvmPlatform(scalaVersions = List(scala2_13, scala3))
-  .settings(onlyJava11)
```
LGLO, line 494: Won't it break build for users of JDK 8? I think they'll have red builds.

adamw: Yes, we require java11 for builds now. One of the modules required it anyway.

build.sbt
adamw: Good point, do we have anybody with windows to fix this?:)

PR #577, opened 2020-05-29
WebSocket stubbing

LGLO:
Proposal for API (and implementation) of adding web socket support.

I don't really like combination of mutable and immutable in WebSocketStub.

Also there are two methods of stubbing `openWebsocket`, one meant for cases when supplied  `WebSocketHandler`s are used and return `WebSocket` and the other one where user specifies his handler - basically Akka.

### DIFF
--- a/akka-http-backend/src/test/scala/sttp/client/akkahttp/SttpBackendStubAkkaTests.scala
+++ b/akka-http-backend/src/test/scala/sttp/client/akkahttp/SttpBackendStubAkkaTests.scala
@@ -1,0 +1,89 @@
+package sttp.client.akkahttp
+
+import akka.actor.ActorSystem
+import akka.http.scaladsl.model.ws.{Message, TextMessage}
+import akka.stream.OverflowStrategy
+import akka.stream.scaladsl.{Flow, Keep, Sink, Source}
+import org.scalatest.BeforeAndAfterAll
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sttp.client._
+import sttp.model.Headers
+
+import scala.concurrent.{Await, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration._
+
+class SttpBackendStubAkkaTests extends AnyFlatSpec with Matchers with ScalaFutures with BeforeAndAfterAll {
+
+  implicit val system: ActorSystem = ActorSystem()
+
+  override protected def afterAll(): Unit = {
+    Await.result(system.terminate().map(_ => ()), 5.seconds)
+  }
+
+  "backend stub" should "cycle through responses using a single sent request" in {
+    // given
+    implicit val backend = AkkaHttpBackend.stub
+      .whenRequestMatches(_ => true)
+      .thenRespondCyclic("a", "b", "c")
+
+    // when
+    def r = basicRequest.get(uri"http://example.org/a/b/c").send().futureValue
+
+    // then
+    r.body shouldBe Right("a")
+    r.body shouldBe Right("b")
+    r.body shouldBe Right("c")
+    r.body shouldBe Right("a")
+  }
+
+  it should "use given flow as web socket handler" in {
+    // This test is an example how can we test client flow.
+    // We check behavior of client when connected to echo server.
+    // Client responsibility was to send two messages to the server and collect received messages.
+    val useHandler: Flow[Message, Message, Future[Seq[Message]]] => Future[Seq[Message]] = clientFlow => {
+      val ((outQueue, clientReceivedMessages), inQueue) = Source
+        .queue(1, OverflowStrategy.fail)
+        .viaMat(clientFlow)(Keep.both)
+        .toMat(Sink.queue())(Keep.both)
+        .run()
+
+      def echoMsg(): Future[Unit] =
+        inQueue.pull().flatMap {
+          case None =>
+            echoMsg()
+          case Some(msg) =>
+            outQueue.offer(TextMessage(s"echo: " + msg.asTextMessage.getStrictText)).map(_ => ())
+        }
+
+      (for {
+        _ <- outQueue.offer(TextMessage("Hi!"))
+        _ <- echoMsg()
+        _ <- echoMsg()
+        _ = outQueue.complete()
+        _ <- outQueue.watchCompletion()
+      } yield ()).flatMap(_ => clientReceivedMessages)
+    }
+
+    val clientFlow: Flow[Message, Message, Future[Seq[Message]]] = {
+      Flow.fromSinkAndSourceMat(
+        Sink.seq[Message],
+        Source((1 to 2).map(i => TextMessage(s"test$i")))
+      )(Keep.left)
+    }
+
+    implicit val b = AkkaHttpBackend.stub
+      .whenRequestMatches(_ => true)
+      .thenHandleOpenWebSocket(Headers(List.empty), useHandler)
+
+    val receivedMessages = basicRequest
+      .get(uri"wss://echo.websocket.org")
+      .openWebsocket(clientFlow)
+      .flatMap(_.result)
+      .futureValue.toList
+
+    receivedMessages shouldBe List("Hi!", "echo: test1", "echo: test2").map(TextMessage(_))
+  }
+}
--- a/async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
+++ b/async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
@@ -138,5 +138,5 @@ object AsyncHttpClientCatsBackend {
     *
     * See [[SttpBackendStub]] for details on how to configure stub responses.
     */
-  def stub[F[_]: Concurrent]: SttpBackendStub[F, Nothing] = SttpBackendStub(new CatsMonadAsyncError())
+  def stub[F[_]: Concurrent]: SttpBackendStub[F, Nothing, WebSocketHandler] = SttpBackendStub(new CatsMonadAsyncError())
 }
--- a/async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
+++ b/async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
@@ -1,19 +1,22 @@
 package sttp.client.asynchttpclient.zio
 
 import org.scalatest.concurrent.ScalaFutures
-import sttp.client._
-import sttp.client.impl.zio.RIOMonadAsyncError
-import sttp.client.testing.SttpBackendStub
-import sttp.client.impl.zio._
-import zio.Task
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+import sttp.client._
+import sttp.client.testing.SttpBackendStub
+import sttp.client.impl.zio._
+import sttp.client.monad.MonadError
+import sttp.client.ws.{WebSocket, WebSocketEvent}
+import sttp.model.Headers
+import sttp.model.ws._
+import zio._
 
 class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFutures {
 
   "backend stub" should "cycle through responses using a single sent request" in {
     // given
-    implicit val b: SttpBackendStub[Task, Nothing] = SttpBackendStub(new RIOMonadAsyncError[Any])
+    implicit val b: SttpBackendStub[Task, Nothing, NothingT] = SttpBackendStub(new RIOMonadAsyncError[Any])
       .whenRequestMatches(_ => true)
       .thenRespondCyclic("a", "b", "c")
 
@@ -25,5 +28,42 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
     runtime.unsafeRun(r).body shouldBe Right("b")
     runtime.unsafeRun(r).body shouldBe Right("c")
     runtime.unsafeRun(r).body shouldBe Right("a")
+  }
+
+  it should "return given web socket response" in {
+    val rioMonad: MonadError[zio.Task] = new RIOMonadAsyncError[Any]
+    val frame1 = WebSocketFrame.text("initial frame")
+    val sentFrame = WebSocketFrame.text("sent frame")
+
+    def webSocket(queue: Queue[WebSocketFrame.Incoming]) =
+      new WebSocket[Task] {
+        override def isOpen: zio.Task[Boolean] = Task.succeed(true)
+        override def monad: MonadError[zio.Task] = rioMonad
+        override def receive: zio.Task[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]] = queue.take.map(Right(_))
+        override def send(f: WebSocketFrame, isContinuation: Boolean): zio.Task[Unit] =
+          f match {
+            case t: WebSocketFrame.Text => queue.offer(t).unit
+            case _                      => Task.unit
+          }
+      }
+
+    def makeBackend(queue: Queue[WebSocketFrame.Incoming]) =
+      AsyncHttpClientZioBackend.stub
+        .whenRequestMatches(_ => true)
+        .thenRespondWebSocket(Headers(List.empty), webSocket(queue))
+
+    val test = for {
+      queue <- Queue.unbounded[WebSocketFrame.Incoming]
+      _ <- queue.offer(frame1)
+      backend = makeBackend(queue)
+      handler <- ZioWebSocketHandler()
+      request = basicRequest.get(uri"http://example.org/a/b/c")
+      ws <- backend.openWebsocket(request, handler).map(_.result)
+      msg1 <- ws.receive
+      _ <- ws.send(sentFrame, false)
+      msg2 <- ws.receive
+    } yield (msg1, msg2)
+
+    runtime.unsafeRun(test) shouldBe ((Right(frame1), Right(sentFrame)))
   }
 }
--- a/async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
+++ b/async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
@@ -1,0 +1,137 @@
+package sttp.client.asynchttpclient.zio
+
+import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import sttp.client._
+import sttp.client.testing.WebSocketStub
+import sttp.client.impl.zio._
+import sttp.client.ws.WebSocketEvent
+import sttp.model.Headers
+import sttp.model.ws.WebSocketFrame
+
+import scala.util.{Failure, Success}
+
+class WebSocketStubZioTests extends AnyFlatSpec with Matchers with ScalaFutures {
+
+  def mkBackend(webSocketStub: WebSocketStub[_]) =
+    AsyncHttpClientZioBackend.stub
+      .whenRequestMatches(_ => true)
+      .thenRespondWebSocket(Headers(List.empty), webSocketStub)
+
+  "web socket stub" should "return initial Incoming frames on 'receive'" in {
+    val frames = List("a", "b", "c").map(WebSocketFrame.text(_))
+    val webSocketStub = WebSocketStub.withInitialIncoming(frames)
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      msg1 <- ws.receive
+      msg2 <- ws.receive
+      msg3 <- ws.receive
+    } yield List(msg1, msg2, msg3)
+
+    runtime.unsafeRun(test) shouldBe frames.map(Right(_))
+  }
+
+  it should "return initial responses on 'receive'" in {
+    val okFrame = WebSocketFrame.text("abc")
+    val exception = new Exception("boom")
+    val webSocketStub = WebSocketStub.withInitialResponses(List(Success(Right(okFrame)), Failure(exception)))
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      msg <- ws.receive
+      err <- ws.receive.either
+    } yield (msg, err)
+
+    runtime.unsafeRun(test) shouldBe ((Right(okFrame), Left(exception)))
+  }
+
+  it should "add next Incoming frames as reaction to send" in {
+    val firstFrame = WebSocketFrame.text("No. 1")
+    val secondFrame = WebSocketFrame.text("Second")
+    val thirdFrame = WebSocketFrame.text("3")
+    val expectedFrame = WebSocketFrame.text("give me more!")
+    val webSocketStub = WebSocketStub
+      .withInitialIncoming(List(firstFrame))
+      .thenRespond {
+        case `expectedFrame` => List(secondFrame, thirdFrame)
+        case _               => List.empty
+      }
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      msg1 <- ws.receive
+      err1 <- ws.receive.either
+      _ <- ws.send(WebSocketFrame.text("not expected"))
+      err2 <- ws.receive.either
+      _ <- ws.send(expectedFrame)
+      msg2 <- ws.receive
+      msg3 <- ws.receive
+    } yield (msg1, err1.isLeft, err2.isLeft, msg2, msg3)
+
+    runtime.unsafeRun(test) shouldBe ((Right(firstFrame), true, true, Right(secondFrame), Right(thirdFrame)))
+  }
+
+  it should "add next responses as reaction to send" in {
+    val ok = WebSocketFrame.text("ok")
+    val exception = new Exception("some error")
+
+    val webSocketStub = WebSocketStub.withNoInitialResponses
+      .thenRespondWith {
+        case _ => List(Success(Right(ok)), Failure(exception))
+      }
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      _ <- ws.send(WebSocketFrame.text("let's add responses"))
+      msg <- ws.receive
+      err1 <- ws.receive.either
+      err2 <- ws.receive.either
+    } yield (msg, err1, err2.isLeft)
+
+    runtime.unsafeRun(test) shouldBe ((Right(ok), Left(exception), true))
+  }
+
+  it should "be closed after sending Close event" in {
+    val ok = WebSocketFrame.text("ok")
+    val closeEvent = WebSocketEvent.Close(500, "internal error")
+    val webSocketStub = WebSocketStub.withInitialResponses(List(Success(Left(closeEvent)), Success(Right(ok))))
+
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      _ <- ws.send(WebSocketFrame.text("let's add responses"))
+      close <- ws.receive
+      isOpen <- ws.isOpen
+      err <- ws.receive.either
+    } yield (close, isOpen, err.isLeft)
+
+    runtime.unsafeRun(test) shouldBe ((Left(closeEvent), false, true))
+  }
+
+  it should "use state to add next responses" in {
+    val webSocketStub = WebSocketStub.withNoInitialResponses
+      .thenRespondWithS(0) {
+        case (counter, _) => (counter + 1, List(Success(Right(WebSocketFrame.text(s"No. $counter")))))
+      }
+    implicit val b = mkBackend(webSocketStub)
+    val test = for {
+      handler <- ZioWebSocketHandler()
+      ws <- basicRequest.get(uri"http://example.org/a/b/c").openWebsocket(handler).map(_.result)
+      _ <- ws.send(WebSocketFrame.text("a"))
+      _ <- ws.send(WebSocketFrame.text("b"))
+      _ <- ws.send(WebSocketFrame.text("c"))
+      msg1 <- ws.receive
+      msg2 <- ws.receive
+      msg3 <- ws.receive
+    } yield List(msg1, msg2, msg3)
+
+    runtime.unsafeRun(test) shouldBe List("No. 0", "No. 1", "No. 2").map(s => Right(WebSocketFrame.text(s)))
+  }
+}
--- a/core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
+++ b/core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
@@ -4,12 +4,15 @@ import java.io.InputStream
 
 import sttp.client._
 import sttp.client.internal._
+
 import sttp.client.monad.{FutureMonad, IdMonad, MonadError}
 import sttp.client.testing.SttpBackendStub._
 import sttp.client.internal.SttpFile
-import sttp.client.ws.WebSocketResponse
+import sttp.client.ws.{WebSocket, WebSocketEvent, WebSocketResponse}
 import sttp.client.{IgnoreResponse, ResponseAs, ResponseAsByteArray, SttpBackend}
-import sttp.model.StatusCode
+import sttp.client.SttpClientException.ReadException
+import sttp.model.{Headers, StatusCode}
+import sttp.model.ws.WebSocketFrame
 
 import scala.concurrent.Future
 import scala.language.higherKinds
@@ -32,11 +35,12 @@ import scala.util.{Failure, Success, Try}
   * or headers. A [[ClassCastException]] might occur if for a given request,
   * a response is specified with the incorrect or inconvertible body type.
   */
-class SttpBackendStub[F[_], S](
+class SttpBackendStub[F[_], S, WS_HANDLER[_]](
     monad: MonadError[F],
     matchers: PartialFunction[Request[_, _], F[Response[_]]],
-    fallback: Option[SttpBackend[F, S, NothingT]]
-) extends SttpBackend[F, S, NothingT] { // TODO
+    wsMatchers: PartialFunction[Request[_, _], WhenOpenWebsocket[F, WS_HANDLER]],
+    fallback: Option[SttpBackend[F, S, WS_HANDLER]]
+) extends SttpBackend[F, S, WS_HANDLER] {
 
   /**
     * Specify how the stub backend should respond to requests matching the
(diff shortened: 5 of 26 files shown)


async-http-client-backend/cats/src/main/scala/sttp/client/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
```diff
@@ -138,5 +138,5 @@ object AsyncHttpClientCatsBackend {
     *
     * See [[SttpBackendStub]] for details on how to configure stub responses.
     */
-  def stub[F[_]: Concurrent]: SttpBackendStub[F, Nothing] = SttpBackendStub(new CatsMonadAsyncError())
```
LGLO, line 141: This change is basically binding stub to web socket handler type for given backend. I happens a lot in this PR.

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -26,4 +28,29 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
     runtime.unsafeRun(r).body shouldBe Right("c")
     runtime.unsafeRun(r).body shouldBe Right("a")
   }
+
+  it should "return given web socket response" in {
```
LGLO, line 32: This is example of usage of `WebSocketStub` and also stubbing `openWebsocket`. I ought to add similar thing for akka and put both in docs.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -86,8 +93,24 @@ class SttpBackendStub[F[_], S](
...
+            monad.unit(WebSocketResponse(headers, uz(handler)))
+          case ReturnWebsocketResponse(headers, response) =>
+            monad.map(response) { r => WebSocketResponse(headers, r.asInstanceOf[WR]) }
+        }
+      case Success(None) =>
+        fallback match {
+          case None     => monad.error(new ReadException(new Exception("request not matched any stub path")))
+          case Some(fb) => fb.openWebsocket(request, handler)
+        }
+      case Failure(e) => monad.error(e)
```
magdzikk, line 111: Is it possible to flatten it this way?
```
case Success(Some(UseHandler(headers, useHandler))) =>
  val uz = useHandler.asInstanceOf[WS_HANDLER[WR] => WR]
  monad.unit(WebSocketResponse(headers, uz(handler)))
case Success(Some(ReturnWebsocketResponse(headers, response))) =>
  monad.map(response) { r => WebSocketResponse(headers, r.asInstanceOf[WR]) }
case Success(None) =>
  fallback match {
    case None     => monad.error(new ReadException(new Exception("request not matched any stub path")))
    case Some(fb) => fb.openWebsocket(request, handler)
  }
case Failure(e) => monad.error(e)
```

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -97,49 +120,75 @@ class SttpBackendStub[F[_], S](
...
-      new SttpBackendStub(monad, matchers.orElse(m), fallback)
+      new SttpBackendStub[F, S, WS_HANDLER](monad, matchers.orElse(m), wsMatchers, fallback)
+    }
+
+    /**
+      * When [[openWebsocket()]] is called, it will ignore handler and return given result.
+      * It is intended to be used when [[openWebsocket()]] is called with sttp supplied handlers
+      * that return wrapped WebSocket and WebSocketStub as WS_RESULT.
+      * */
+    def thenWebsocketResponse[WS_RESULT](headers: Headers, result: WS_RESULT) = {
```
magdzikk, line 146: Should we add an explicit return type to these public methods?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
     }
   }
 }
+
+/**
+  * Websockets use can differ, for akka backend crucial part is the handler.
+  * For other backends
```
magdzikk, line 288: You probably know it, but just a reminder that this looks unfinished

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+) extends WebSocket[F] {
+
+  private var state: S = initialState
+
+  override def isOpen: F[Boolean] = monad.unit(true)
+
+  override def receive: F[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]] =
+    events.dequeueFirst(_ => true) match {
+      case Some(event) => event
+      case None        => monad.error(new Exception("Unexpected 'receive', no more prepared events."))
```
magdzikk, line 317: 👍

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+private[testing] case class UseHandler[F[_], WS_RESULT, WS_HANDLER[_]](
+    headers: Headers,
+    useHandler: Any => WS_RESULT
+) extends WhenOpenWebsocket[F, WS_HANDLER]
+
+private[testing] case class ReturnWebsocketResponse[F[_], WS_RESULT](headers: Headers, response: F[WS_RESULT])
+    extends WhenOpenWebsocket[F, NothingT]
+
+/**
+  * Simple stub for web sockets that uses queue of events for receive.
```
magdzikk, line 300:
```suggestion
  * A simple stub for web sockets that uses a queue of events for receive.
```

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    }
+
+  override def send(frame: WebSocketFrame, isContinuation: Boolean): F[Unit] =
+    monad.eval {
+      val (newState, newEvents) = makeNewEvents(state, frame)
+      newEvents.foreach(events.enqueue(_))
+      state = newState
+    }
+
+  /** Returns stub that has same prepared messages but adds new messages to receive in reaction to send.
```
magdzikk, line 327:
```suggestion
  /** Returns a stub that has same prepared messages but adds new messages to receive in reaction to send.
```

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    *
+    * @param addReceived
```
magdzikk, line 329: Just a reminder that this looks unfinished

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    */
+  def respondUsing(addReceived: WebSocketFrame => List[WebSocketFrame.Incoming]): WebSocketStub[F, Unit] =
+    respondUsingF(
+      addReceived.andThen(_.map(m => monad.unit(Right(m): Either[WebSocketEvent.Close, WebSocketFrame.Incoming])))
+    )
+
+  /** Returns stub that has same prepared messages but adds new messages to receive in reaction to send.
```
magdzikk, line 336:
```suggestion
  /** Returns a stub that has same prepared messages but adds new messages to receive in reaction to send.
```

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+
+  def respondUsingStateF[S2](initial: S2)(
+      onSend: (S2, WebSocketFrame) => (S2, List[F[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]]])
+  ): WebSocketStub[F, S2] = new WebSocketStub(mutable.Queue.concat(events), initial, onSend)
+
+}
+
+object WebSocketStub {
+
+  /** Creates stub that has given incoming frames prepared for 'receive' and doesn't add messages on 'send'.
```
magdzikk, line 365:
```suggestion
  /** Creates a stub that has given incoming frames prepared for 'receive' and doesn't add messages on 'send'.
```

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    * There is more powerful version that takes list of effects to return.
```
magdzikk, line 366: Do you mean that this here is the more powerful version, or that somewhere else is a more powerful version?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    */
+  def respondWithMessages[F[_]](
+      messages: List[WebSocketFrame.Incoming]
+  )(implicit monad: MonadError[F]): WebSocketStub[F, Unit] = {
+    respondWithMessagesF(messages.map(m => monad.unit(Right(m): Either[WebSocketEvent.Close, WebSocketFrame.Incoming])))
+  }
+
+  /** Creates stub that has given effects prepared for 'receive' and doesn't add messages on 'send' */
```
magdzikk, line 374:
```suggestion
  /** Creates a stub that has given effects prepared for 'receive' and doesn't add messages on 'send' */
```

core/src/test/scala/sttp/client/testing/SttpBackendStubTests.scala
```diff
@@ -227,8 +226,32 @@ class SttpBackendStubTests extends AnyFlatSpec with Matchers with ScalaFutures {
     basicRequest.get(uri"http://example.org").response(asStringAlways).send().body shouldBe "Internal server error"
   }
 
+  it should "return given response when openWebsocket" in {
+    val anythingCanBeWebSocketStub: Int = 42
```
magdzikk, line 230: 😄

core/src/test/scala/sttp/client/testing/SttpBackendStubTests.scala
```diff
@@ -227,8 +226,32 @@ class SttpBackendStubTests extends AnyFlatSpec with Matchers with ScalaFutures {
...
+
+  it should "use given handler to open websocket" in {
+    val dummyHandler: Identity[String] = "dummy handler"
+    val useHandler: Identity[String] => String = str => str + " was used"
+    val headers = Headers(List(Header.apply("h", "v")))
+    implicit val s: SttpBackend[Identity, Nothing, Identity] = SttpBackendStub(IdMonad).whenAnyRequest
+      .thenUseWebsocketHandler(headers, useHandler)
+    basicRequest.get(uri"http://example.org").openWebsocket(dummyHandler) shouldBe WebSocketResponse(
+      headers,
+      "dummy handler was used"
```
magdzikk, line 249: ❤️

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -26,4 +28,29 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
...
+    val monad = new RIOMonadAsyncError[Any]
+    val frame1 = WebSocketFrame.text("initial frame")
+    val sentFrame = WebSocketFrame.text("sent frame")
+    val webSocketStub =
+      WebSocketStub
+        .respondWithMessages(List(frame1))(monad)
```
adamw, line 38: would it be possible not to require the `monad` instance here? I.e. make `WebSocketStub` "lazy", parametrised with the monad instance, which would be given later by the `SttpBackendStub` (which already knows it). That would make this more user-friendly :)

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -26,4 +28,29 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
...
+        .respondUsing(frame => List(frame.asInstanceOf[WebSocketFrame.Text]))
```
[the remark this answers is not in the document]
adamw: Or maybe it would be more practical to accept a `PartialFunction` here, allowing to write:

```
.respondUsing {
  case tm: WebSocketFrame.Text => tm 
}
```
?

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -26,4 +28,29 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
...
+    implicit val b = AsyncHttpClientZioBackend.stub
+      .whenRequestMatches(_ => true)
+      .thenWebsocketResponse(
```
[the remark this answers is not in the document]
adamw: We have a family of `.thenRespondXxx()`, so we probably should have something symmetrical. Maybe simply `.thenRespondWebSocket`?

Or maybe use different terminology, `.thenOpenWebSocket`? Though it doesn't *really* open a WS, it just returns a stub ;)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
adamw, line 146: maybe: `.thenWebSocketResult` and `.thenWebSocketResultFromHandler`? To keep the result/handler terminology

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+      (newState, messages.map(m => monad.unit(Right(m): Either[WebSocketEvent.Close, WebSocketFrame.Incoming])))
+    })
+
+  def respondUsingStateF[S2](initial: S2)(
+      onSend: (S2, WebSocketFrame) => (S2, List[F[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]]])
+  ): WebSocketStub[F, S2] = new WebSocketStub(mutable.Queue.concat(events), initial, onSend)
+
+}
+
+object WebSocketStub {
```
adamw, line 363: shouldn't we have a top-level `respondUsingState` as well?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+    useHandler: Any => WS_RESULT
+) extends WhenOpenWebsocket[F, WS_HANDLER]
+
+private[testing] case class ReturnWebsocketResponse[F[_], WS_RESULT](headers: Headers, response: F[WS_RESULT])
+    extends WhenOpenWebsocket[F, NothingT]
+
+/**
+  * Simple stub for web sockets that uses queue of events for receive.
+  */
+class WebSocketStub[F[_], S](
```
adamw, line 302: I think to cover all use-cases we should also expose a method to enqueue events to be sent. This would be needed e.g. to simulate a websocket which periodically sends out a message.

Also, maybe we could make this thread-safe? Even in a "primitive" way, using an `AtomicReference` or maybe even a synchronized method.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +282,100 @@ object SttpBackendStub {
...
+  def respondUsing(addReceived: WebSocketFrame => List[WebSocketFrame.Incoming]): WebSocketStub[F, Unit] =
+    respondUsingF(
+      addReceived.andThen(_.map(m => monad.unit(Right(m): Either[WebSocketEvent.Close, WebSocketFrame.Incoming])))
+    )
+
+  /** Returns stub that has same prepared messages but adds new messages to receive in reaction to send.
+    *
+    * @param addReceived
+    */
+  def respondUsingF(
```
adamw, line 340: it's also probably important to mention (in the docs) that this *replaces* the current event-producing logic. (And the enqueued events remain as-is)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
adamw: I retract the message-enqueueing method, this would require `receive` to get suspended whil waiting for new messages, and could cause tests to hang. That's I think outside the scope of a stub :)

But still would be nice to have things thread-safe :)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
adamw: Ah I see how this works now :)
Maybe we could try to make this read nicer, e.g.:

```
WebSocketStub.respondWithMessages(...).thenRespondUsing(...)
```

to emphasize that first the initial messages will be used, and *then* the stub logic

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
LGLO: The price is either loosing power to set "failed response" or requiring monad instance when setting responses (which doesn't make it better).

adamw: hm can you give an example of setting a failed response? What would be the use-case?

LGLO: `.respondUsingF(_ => List(monad.error(new Exception("boom!")), monad.unit(Right(frame1))))`
Would err on the first `receive` and succeed on the second.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
LGLO: Somewhere else (postfix F)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
LGLO: Will do! Also this `uz` was meant to be temporary name.

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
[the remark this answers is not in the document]
LGLO: I propose `thenRespondWebSocket` and `thenHandleOpenWebSocket` depending on parameter.

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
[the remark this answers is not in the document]
LGLO:
```
Is
.respondUsing {
  case t: WebSocketFrame.Text => List(t)
  case _                      => List.empty
}
```
that bad?

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
adamw: hm but when you create a WebSocketStub, you already know the concrete effect type, so the user can just pass in the failed effect? Without having to provide the sttp monad

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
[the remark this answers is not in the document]
adamw: 👍

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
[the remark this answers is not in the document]
adamw: yeah, that's ok :)

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/SttpBackendStubZioTests.scala
LGLO: `.respondUsingF(_ => List(RIO.fail(new Exception("boom!")), ZIO.succeed(Right(frame1))))`
Is valid as well.

Let's assume we don't have an instance of monad available when setting stub.
`type Message = Either[WebSocketEvent.Close, WebSocketFrame]`
Choices I see:
1. we store `List[Message]` and user have no power to set error
1. we store `List[F[Message]]` and user cannot use convenient API (for happy-paths) that takes frames without `F`, because we can't store it without `monad.unit`
1. we store `List[Try[Message]]` we have both APIs

I'll give the last option a shot

adamw: 4. we store `List[SttpMonad => F[Message]]`, which would allow us to provide the convenient API as well?

LGLO: I don't see much improvement over `Try` because I can simply translate `Success(a)` => `monad.unit(a)`, `Failure(t)` => `monad.error(t)`>
Would it be more convenient for users to supply `Message => List[Try[Message]]` or `Message => MonadError => List[F[Message]]`?

Still I think that for most complicated cases, users should fallback to implementing `WebSocket` themselves.

adamw: ok, let's try `Try` :)

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
```diff
@@ -0,0 +1,119 @@
...
+      err <- ws.receive.either
+    } yield (msg, err)
+
+    runtime.unsafeRun(test) shouldBe ((Right(okFrame), Left(exception)))
+  }
+
+  it should "add next Incoming frames as reaction to send" in {
+    val firstFrame = WebSocketFrame.text("No. 1")
+    val secondFrame = WebSocketFrame.text("Second")
+    val thridFrame = WebSocketFrame.text("3")
```
magdzikk, line 56: typo: `thrid`

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
magdzikk: Which method is it?

LGLO: `withMessages` but I'm going to rename it to `withResponses`. I'm going one more round through this PR, focusing on docs.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -97,49 +117,84 @@ class SttpBackendStub[F[_], S](
...
+      }
+      new SttpBackendStub[F, S, WS_HANDLER](monad, matchers, wsMatchers.orElse(m), fallback)
+    }
+
+    /**
+      * When [[openWebsocket()]] is called, it will ignore handler and return given result.
+      * It is intended to be used when [[openWebsocket()]] is called with sttp supplied handlers
+      * It returns [[WebSocket]] built by [[WebSocketStub]] wrapped as WS_RESULT.
+      * */
+    def thenRespondWebSocket[WS_RESULT](headers: Headers, result: WebSocketStub[_]): SttpBackendStub[F, S, WS_HANDLER] = {
```
adamw, line 156: WS_RESULT can be removed?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -97,49 +117,84 @@ class SttpBackendStub[F[_], S](
...
+      thenRespondWebSocket(headers, result.build(monad))
+    }
+
+    /**
+      * When [[openWebsocket()]] is called it uses given headers and handler to create result.
+      * It is intended to be used when [[openWebsocket()]] is called with user supplied handler that
+      * doesn't return WebSocket object to act on.
+      * */
+    def thenHandleOpenWebSocket[WS_RESULT](headers: Headers, useHandler: WS_HANDLER[WS_RESULT] => WS_RESULT) = {
+      val m: PartialFunction[Request[_, _], WhenOpenWebsocket[F, WS_HANDLER]] = {
```
adamw, line 166: would it be possible to somehow expose a variant which would expose the partial function? So that the returned websocket/handler could depend (dynamically) on the request

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +288,109 @@ object SttpBackendStub {
...
+
+/**
+  * A simple stub for web sockets that uses a queue of events for receive.
+  * New messages can be added to queue when `send` is invoked.
+  * For more complex cases, please provide own implementation of [[WebSocket]].
+  */
+class WebSocketStub[S](
+    initialResponses: List[Try[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]]],
+    initialState: S,
+    makeNewEvents: (S, WebSocketFrame) => (S, List[Try[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]]])
```
adamw, line 315: terminology: either `Responses` or `Events`? :)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +288,109 @@ object SttpBackendStub {
...
+
+  /** Returns stub that has same initial messages but replaces funtion that adds messages to receive when `sent` is called. */
+  def thenRespond(addReceived: WebSocketFrame => List[WebSocketFrame.Incoming]): WebSocketStub[Unit] =
+    thenRespondWith(
+      addReceived.andThen(_.map(m => Success(Right(m): Either[WebSocketEvent.Close, WebSocketFrame.Incoming])))
+    )
+
+  /** More powerful version of [[thenRespond()]]. Allows to use function that calls WebSocket to close or fails. */
+  def thenRespondWith(
+      addReceived: WebSocketFrame => List[Try[Either[WebSocketEvent.Close, WebSocketFrame.Incoming]]]
```
adamw, line 326: maybe this should be: `Try[Either[Close, List[Incoming]]]`?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
```diff
@@ -211,3 +288,109 @@ object SttpBackendStub {
...
+        }
+    }
+}
+
+object WebSocketStub {
+
+  /** Creates a stub that has given incoming frames prepared for 'receive' and doesn't add messages on 'send'.
+    * There is more powerful version [[withInitialMessages()]] that takes list of effects to return.
+    */
+  def withInitialIncoming(
```
adamw, line 380: 👍

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
```diff
@@ -0,0 +1,117 @@
...
+    runtime.unsafeRun(test) shouldBe ((Right(firstFrame), true, true, Right(secondFrame), Right(thridFrame)))
+  }
+
+  it should "add next responses as reaction to send" in {
+    val exception = new Exception("transient error")
+    val ok = WebSocketFrame.text("ok")
+
+    val webSocketStub = WebSocketStub.withNoInitialMessages
+      .thenRespondWith {
+        case _ => List(Failure(exception), Success(Right(ok)))
```
adamw, line 84: shouldn't the websocket get closed after the exception or a `Close` event? That is, a websocket which already produced an exception or a close should be unusable later on

adamw:
Shapes up nicely, thanks @LGLO :)

akka-http-backend/src/test/scala/sttp/client/akkahttp/SttpBackendStubAkkaTests.scala
```diff
@@ -0,0 +1,83 @@
...
+      )
+
+    basicRequest
+      .get(uri"wss://echo.websocket.org")
+      .openWebsocket(clientFlow)
+      .flatMap(_.result)
+      .onComplete {
+        case Failure(exception) => println(s"WebSocket failure: $exception")
+        case Success(results) =>
+          println(results)
```
adamw, line 78: can't we assert on the results that it contains the expected messages?

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
LGLO: So, there is no such thing as transient error in WebSocket?

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
LGLO: Do you mean 1 to 1 or just using other imports to shorten?
We don't offer full flexibility in stub, but I think having possibility to add 0 to many new responses for on 1 input message for price of having `List` in this signature is a good tradeoff.

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
adamw: not that I'm aware of, if the TCP conneciton gets broken, it's broken, I wouldn't use it again ;)

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
adamw: I mean reordering the effects :) The outer effect - Try - as everything can fail. Then, we can either close the websocket or receive 0-n messages. There's no sense receiving messages after a close, for example :)

adamw: (and the stub should work like that - once a `Close` has been received by the client, the websocket (here: stub) becomes closed, and any further `.receive` attempts end with an exception)

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
LGLO: Then I'll make WebSocketStub stay in last failure and also fail after it returned closed.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
LGLO: Done

async-http-client-backend/zio/src/test/scala/sttp/client/asynchttpclient/zio/WebSocketStubZioTests.scala
LGLO: Thanks! My dyslexia is very well. I had to read it 3 times before I've finally seen it's `thrid`.

core/src/main/scala/sttp/client/testing/SttpBackendStub.scala
LGLO: done

akka-http-backend/src/test/scala/sttp/client/akkahttp/SttpBackendStubAkkaTests.scala
LGLO: I've reworked it but forgot to push after commit, sorry.

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+`Flow[Message, Message, Mat]` to define WebSocket client behavior instead of working with `WebSocket` instance.
+`thenHandleOpenWebSocket` method allows to use given flow to simulate server side in tests.
+
+Example:
+```scala
+// it should say Hi! and 42
+val behaviorToTest: Flow[Message, Message, Future[Done]] = ???
+// setup test double
+var received = List.empty[String]
+val testFlow: Flow[Message, Message, Future[Done]] => Future[Done] = clientFlow => {
```
adamw, line 259: `serverFlow`?

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
 
 val response2 = basicRequest.post(uri"http://api.internal/b").send()
 // response2 will be whatever a "real" network call to api.internal/b returns
+```
+
+## Testing WebSockets
+
+Stub methods `whenRequestMatches` and `whenAnyRequest` also allow to specify behavior for `openWebsocket` call when used together with `thenRespondWebSocket` or `thenHandleOpenWebSocket`. `thenRespondWebSocket` is intended for cases when sttp supplied
```
magdzikk, line 204: Unneeded line break at the end of this line (i.e. lines 204 and 205 should be one line)

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+`WebSocketHandler` is used, that are cases where users code manipulates `WebSocket[F]` to implement logic.
```
magdzikk, line 205:
```suggestion
`WebSocketHandler` is used, i.e. the cases where the user's code manipulates `WebSocket[F]` to implement logic.
```

magdzikk: (if I understand it correctly)

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+  handler <- ZioWebSocketHandler() // using sttp supplied handler
+  openResponse <- testingBackend.openWebsocket(basicRequest.get(uri"wss://some.uri"), handler)
+  webSocket = openResponse.result // interaction with WebSocket is interesting in this case
+  message <- webSocket.receive
+} yield message
+```
+
+## WebSocketStub
+
+Implementation of `WebSocket` stub is available. It wraps state machine that can be used to simulate simple WebSockets interactions.
```
magdzikk, line 227:
```suggestion
There is an implementation of `WebSocket` available. It wraps a state machine that can be used to simulate simple WebSockets interactions.
```

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+It allows to set initial responses for `receive` calls as well as logic to add further messages in reaction to `send` calls.
+`SttpBackendStub` has special API that accepts `WebSocketStub` and builds it to `WebSocket` instance.
```
magdzikk, line 229:
```suggestion
`SttpBackendStub` has a special API that accepts a `WebSocketStub` and builds a `WebSocket` instance out of it.
```

magdzikk: (if I understand that correctly)

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+    List(WebSocketFrame.text("Hello from the server!"))
+  )
+  .thenRespondS(0) {
+    case (counter, tf: WebSocketFrame.Text) => (counter + 1, List(WebSocketFrame.text(s"echo: ${tf.payload}")))
+    case (counter, _)                       => (counter, List.empty)
+  }
+backend.whenAnyRequest.thenRespondWebSocket(Headers(List.empty), webSocketStub)
+```
+There is a possiblity to add error responses as well, if it's not enough,
+then using own implementation of `WebSocket` is recommended.
```
magdzikk, line 245:
```suggestion
then using a custom implementation of `WebSocket` is recommended.
```

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+
+## Using handler when testing WebSockets
+
+Akka backend works a bit differently in regards to WebSockets. In it's case user defines
```
magdzikk, line 249:
```suggestion
Akka backend works a bit differently in regard to WebSockets. In this case, the user defines
```

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+  .withInitialIncoming(
+    List(WebSocketFrame.text("Hello from the server!"))
+  )
+  .thenRespondS(0) {
+    case (counter, tf: WebSocketFrame.Text) => (counter + 1, List(WebSocketFrame.text(s"echo: ${tf.payload}")))
+    case (counter, _)                       => (counter, List.empty)
+  }
+backend.whenAnyRequest.thenRespondWebSocket(Headers(List.empty), webSocketStub)
+```
+There is a possiblity to add error responses as well, if it's not enough,
```
magdzikk, line 244: There's a lot of extra line breaks in the middle of the sentences, throughout this file - they should be removed.

docs/testing.md
```diff
@@ -197,4 +197,95 @@ val response1 = basicRequest.get(uri"http://api.internal/a").send()
...
+then using own implementation of `WebSocket` is recommended.
+
+## Using handler when testing WebSockets
+
+Akka backend works a bit differently in regards to WebSockets. In it's case user defines
+`Flow[Message, Message, Mat]` to define WebSocket client behavior instead of working with `WebSocket` instance.
```
magdzikk, line 250:
```suggestion
a `Flow[Message, Message, Mat]` to define WebSocket client behavior instead of working with `WebSocket` instance.
```

docs/testing.md
LGLO: :+1: I've edited my changes to apply line breaks before 121 column. Markdown should render single white space in same way, regardless it's line break or space.

magdzikk: Excellent, I didn't know that

PR #582, opened 2020-06-02
Add HttpClientZioBackend

magdzikk:
Implement ZIO backend for Java 11+ HttpClient. No streams for now.

### DIFF
--- a/httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/HttpClientZioBackend.scala
+++ b/httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/HttpClientZioBackend.scala
@@ -1,0 +1,62 @@
+package sttp.client.httpclient.zio
+
+import java.net.http.{HttpClient, HttpRequest}
+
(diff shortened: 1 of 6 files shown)


httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/HttpClientZioBackend.scala
```diff
@@ -0,0 +1,62 @@
...
+      customEncodingHandler
+    )
+
+object HttpClientZioBackend {
+  private def apply(
+      client: HttpClient,
+      closeClient: Boolean,
+      customizeRequest: HttpRequest => HttpRequest,
+      customEncodingHandler: EncodingHandler
+  ): SttpBackend[Task, Nothing, Nothing] =
```
LGLO, line 30: Why don't we have WebSockets, afaik they don't require zio-streams. Do they?

magdzikk: I wanted to make the PR as small as possible, and add web sockets in the next one

PR #585, opened 2020-06-04
OpenTracing: possiblity of modifing SpanBuilder before starting a Span

LGLO:
I've used that same method that is used for modifying Span itself, passing function in tags.
Cons are the same as well: not that safe, abusing (?) tags.

### DIFF
--- a/metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
+++ b/metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
@@ -3,11 +3,12 @@ package sttp.client.opentracing
 import io.opentracing.tag.Tags
 import io.opentracing.{Span, Tracer}
 import io.opentracing.propagation.Format
+import io.opentracing.Tracer.SpanBuilder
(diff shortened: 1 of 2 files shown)


metrics/open-tracing-backend/src/main/scala/sttp/client/opentracing/OpenTracingBackend.scala
```diff
@@ -19,14 +20,22 @@ class OpenTracingBackend[F[_], S] private (delegate: SttpBackend[F, S, NothingT]
...
-              .tag(OpenTracingBackend.OperationIdRequestTag)
-              .getOrElse("default-operation-id")
-              .toString
-          )
-          .withTag(Tags.SPAN_KIND, Tags.SPAN_KIND_CLIENT)
+        val spanBuilderTransformer: SpanBuilderTransformer =
+          request
+            .tag(OpenTracingBackend.SpanBuilderTransformerRequestTag)
+            .collectFirst {
+              case f: SpanBuilderTransformer => f
```
LGLO, line 27: This is suboptimal, because any `Function1[_, _]` will match here, but we do same for `SpanTransformer` itself.

PR #589, opened 2020-06-08
Streams for HttpClientZioBackend

LGLO:
Streams implementation for HttpClientZioBackend.
Blocking is required because we use GZIPInputStream for decompression. Unfortunately Blocking has to pollute F of this backend.


build.sbt
```diff
@@ -553,7 +554,11 @@ lazy val httpClientZioBackend =
   httpClientBackendProject("zio")
     .settings(
       libraryDependencies ++=
-        Seq("dev.zio" %% "zio" % zioVersion)
+        Seq(
+          "dev.zio" %% "zio" % zioVersion,
+          "dev.zio" %% "zio-streams" % zioVersion,
+          "dev.zio" %% "zio-interop-reactivestreams" % "1.0.3.5-RC10"
```
adamw, line 560: isn't the same used for `asyncHttpClientZioStreamsBackend`? if so, we should extract the version number to a constant

httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/HttpClientZioBackend.scala
```diff
@@ -1,62 +1,147 @@
...
       client: HttpClient,
       closeClient: Boolean,
       customizeRequest: HttpRequest => HttpRequest,
-      customEncodingHandler: EncodingHandler
-  ): SttpBackend[Task, Nothing, Nothing] =
-    new FollowRedirectsBackend[Task, Nothing, Nothing](
-      new HttpClientZioBackend(client, closeClient, customizeRequest, customEncodingHandler)
+      customEncodingHandler: EncodingHandler,
+      chunkSize: Int
+  ): SttpBackend[BlockingTask, Nothing, Nothing] =
```
adamw, line 60: The result type should include support for streams :)

adamw: That is it should be sth like `SttpBackend[Task, Stream[Task, Nothing], NothingT]` (also the last param should be a type constructor, it's weird the compiler doesn't complain here)

httpclient-backend/zio/src/test/scala/sttp/client/httpclient/zio/HttpClientZioHttpTest.scala
```diff
@@ -1,11 +1,11 @@
...
+import sttp.client.httpclient.zio.HttpClientZioBackend.BlockingTask
 import sttp.client.{NothingT, SttpBackend}
 import sttp.client.impl.zio._
 import sttp.client.testing.{ConvertToFuture, HttpTest}
-import zio._
 
-class HttpClientZioHttpTest extends HttpTest[Task]{
-  override implicit val backend: SttpBackend[Task, Nothing, NothingT] = runtime.unsafeRun(HttpClientZioBackend())
-  override implicit val convertToFuture: ConvertToFuture[Task] = convertZioIoToFuture
+class HttpClientZioHttpTest extends HttpTest[BlockingTask]{
```
adamw, line 8: we'll also need a subclass of `StreamingTest` to test that streaming works

httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/HttpClientZioBackend.scala
```diff
@@ -1,62 +1,147 @@
...
+    }
+  }
+
+  override def responseBodyToStream(responseBody: InputStream): Try[ZStream[Blocking, Throwable, Byte]] = {
+    Success(Stream.fromInputStreamEffect(ZIO.succeed(responseBody), chunkSize))
+  }
+}
 
 object HttpClientZioBackend {
+  type BlockingTask[A] = ZIO[Blocking, Throwable, A]
```
adamw, line 50: Do we need blocking in the effect type? Or only for streams?

LGLO: Yes, because we use InputStream for reading response body for HttpClient backend.

adamw: Ah yes, that's not only for streams :)

PR #596, opened 2020-06-11
Make OkHttpBackend compatible with Java 8 again

jordiolivares:
With the rewrite for using an InputStream the OkHttp backend used the `InputStream.readAllBytes()` which doesn't exist in Java 8.

Unfortunately for many users, Java 8 is still the norm, so this should make the library compatible with them again.


adamw:
Released in 2.2.0, thanks!

PR #599, opened 2020-06-16
Attach mdoc to sttp release process

ghostbuster91:


docs-generation/src/examples.md
```diff
@@ -0,0 +1,365 @@
...
+
+Required dependencies:
+
+```scala            
+libraryDependencies ++= List("com.softwaremill.sttp.client" %% "core" % "2.2.0")
+```
+
+Example code:
+
+```scala mdoc:compile-only
```
ghostbuster91, line 15: I only added mdoc annotation here, just to show that it works. You can try it by yourself, just break this code and try to regenerate docs

build.sbt
```diff
@@ -707,3 +710,18 @@ lazy val examples = (projectMatrix in file("examples"))
     circe,
     slf4jBackend
   )
+
+lazy val docs: ProjectMatrix = (projectMatrix in file("docs-generation")) // important: it must not be docs/
```
adamw, line 714: why?

PR #603, opened 2020-06-18
WebSocket support for HttpClientZioBackend

magdzikk:

### DIFF
--- a/httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/ZioWebSocketHandler.scala
+++ b/httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/ZioWebSocketHandler.scala
@@ -12,7 +12,7 @@ object ZioWebSocketHandler {
   private class ZioAsyncQueue[A](queue: Queue[A], runtime: Runtime[Any]) extends AsyncQueue[BlockingTask, A] {
     override def offer(t: A): Unit = {
       if (!runtime.unsafeRun(queue.offer(t))) {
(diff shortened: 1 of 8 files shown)


httpclient-backend/zio/src/main/scala/sttp/client/httpclient/zio/ZioWebSocketHandler.scala
```diff
@@ -0,0 +1,36 @@
...
+import sttp.client.ws.{WebSocket, WebSocketEvent}
+import sttp.client.ws.internal.AsyncQueue
+import sttp.model.ws.WebSocketBufferFull
+import zio.{Queue, Runtime, Task, ZIO}
+
+object ZioWebSocketHandler {
+  private class ZioAsyncQueue[A](queue: Queue[A], runtime: Runtime[Any]) extends AsyncQueue[BlockingTask, A] {
+    override def offer(t: A): Unit = {
+      if (!runtime.unsafeRun(queue.offer(t))) {
+        throw new WebSocketBufferFull()
```
adamw, line 15: maybe this should be in fact an `IllegalStateException` or sth similar indicating a bug - we shouldn't ever get into a situation where the buffer is full?

magdzikk: That's true, the name of the exception isn't right

adamw:
Released in 2.2.1, thanks!

adamw:
& fixed in 2.2.2 :)

PR #604, opened 2020-06-19
Don't use buffer in HttpClient monix backend

magdzikk:
This simplifies MonixWebSocketHandler by letting it use `NativeWebSocketHandler` instead of a custom handler, which included unneeded buffering.

### DIFF
--- a/httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
+++ b/httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
@@ -8,6 +8,7 @@ import sttp.client.impl.monix.{MonixAsyncQueue, TaskMonadAsyncError}
 import sttp.client.ws.{WebSocket, WebSocketEvent}
 
 object MonixWebSocketHandler {
+  val IncomingBufferCapacity = 2 // has to be at least 2 (opening frame + one data frame)
 
   /**
     * Returns an effect, which creates a new [[WebSocketHandler]]. The handler should be used *once* to send and
@@ -17,7 +18,8 @@ object MonixWebSocketHandler {
       s: Scheduler
   ): Task[WebSocketHandler[WebSocket[Task]]] = {
     Task {
-      val queue = new MonixAsyncQueue[WebSocketEvent](Some(2))
+      val queue =
+        new MonixAsyncQueue[WebSocketEvent](Some(IncomingBufferCapacity))
       NativeWebSocketHandler(queue, TaskMonadAsyncError)
     }
   }
--- a/httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
+++ b/httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
@@ -1,149 +1,36 @@
 package sttp.client.httpclient.monix
 
-import java.net.http.WebSocket.Listener
-import java.net.http.{WebSocket => JWebSocket}
-import java.nio.ByteBuffer
-import java.util.concurrent.atomic.AtomicBoolean
-import java.util.concurrent.{CompletableFuture, CompletionStage}
-import java.util.function.BiConsumer
-
+import monix.catnap.MVar
 import monix.eval.Task
 import monix.execution.Scheduler
 import sttp.client.httpclient.WebSocketHandler
-import sttp.client.impl.monix.{MonixAsyncQueue, TaskMonadAsyncError}
-import sttp.client.monad.syntax._
-import sttp.client.monad.{Canceler, MonadAsyncError, MonadError}
+import sttp.client.httpclient.internal.NativeWebSocketHandler
+import sttp.client.impl.monix.TaskMonadAsyncError
 import sttp.client.ws.internal.AsyncQueue
 import sttp.client.ws.{WebSocket, WebSocketEvent}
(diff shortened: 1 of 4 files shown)


httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
```diff
@@ -1,149 +1,36 @@
...
-      "Incoming buffer capacity has to be at least 2 (opening frame + one data frame)"
-    )
     Task {
-      val isOpen: AtomicBoolean = new AtomicBoolean(false)
-      val queue = new MonixAsyncQueue[WebSocketEvent](Some(incomingBufferCapacity))
-      WebSocketHandler(
-        new AddToQueueListener(queue, isOpen, incomingBufferCapacity),
-        httpClientWebSocketToWebSocket(_, queue, isOpen, TaskMonadAsyncError)
-      )
+      val queue = new SingleElementQueue[WebSocketEvent]
```
adamw, line 22: thinking about this, we probably do need a 2-element queue (that is, we can still use `NativeWebSocketHandler`, just with a different queue).

That's because of error and close events. If a data frame is waiting to be received, and the ws is closed or there's a network error, despite there being no frames requested (as the data frame is not yet received, and we haven't called `ws.request(1)`), a `Close`/`Error` event will be enqueued. So in this case, we'll have two elements.

And when using an `MVar`, `.put` will block the network thread, which shouldn't happen.

magdzikk: Good point... I wonder if we could write a test for it

adamw: No idea :) we would have to test that the network thread isn't blocked? 

Maybe ... a server which sends a "welcome message" when the WS is opened - this should block, then sending should maybe block as well? - that is, the task would never complete.

Maybe we could try with the `/ws/send_and_wait` test endpoint - it sends two messages when the ws is open

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
```diff
@@ -1,149 +1,24 @@
...
-                cb(Left(error))
-              } else {
-                cb(Right(()))
-              }
-            }
-          })
-          Canceler(() => cf.cancel(true))
-        }
-      }
+      val queue = new MonixAsyncQueue[WebSocketEvent](Some(2))
```
adamw, line 20: maybe we could retain the comment here where the 2 comes from? (`"Incoming buffer capacity has to be at least 2 (opening frame + one data frame)"`)

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/HttpClientHighLevelMonixWebsocketTest.scala
```diff
@@ -22,12 +22,12 @@ class HttpClientHighLevelMonixWebsocketTest extends HighLevelWebsocketTest[Task,
   implicit val convertToFuture: ConvertToFuture[Task] = convertMonixTaskToFuture
   implicit val monad: MonadError[Task] = TaskMonadAsyncError
 
-  def createHandler: Option[Int] => Task[WebSocketHandler[WebSocket[Task]]] = _ => MonixWebSocketHandler(5)
+  def createHandler: Option[Int] => Task[WebSocketHandler[WebSocket[Task]]] = _ => MonixWebSocketHandler()
```
adamw, line 25: I guess the argument doesn't make sense? :)

httpclient-backend/monix/src/main/scala/sttp/client/httpclient/monix/MonixWebSocketHandler.scala
magdzikk: good idea

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/HttpClientHighLevelMonixWebsocketTest.scala
magdzikk: no it doesn't :P

magdzikk: But it's an overridden method from `HighLevelWebSocketTest`. I'll add the `override` keyword

magdzikk: the alternative would be to provide an empty implementation, and e.g. create the handler inline:
```  override def createHandler: Option[Int] => Task[WebSocketHandler[WebSocket[Task]]] = ???

  it should "handle backpressure correctly" in {
    basicRequest
      .get(uri"$wsEndpoint/ws/echo")
      .openWebsocketF(MonixWebSocketHandler())
//(...)
```
Which one do you think would be better?

adamw: Ah, if it's an override, don't bother, it's good as it is :)

adamw:
Released in 2.2.1, thanks!

Issue #618, opened 2020-06-29
Extend ReponseError to allow arbitrary types for error case


ghostbuster91:
This is a summary and continuation of two discussions:
- https://github.com/softwaremill/sttp-model/issues/5
- https://github.com/OpenAPITools/openapi-generator/pull/6684#discussion_r443485605

As in openapi specification we are aware of possible error responses there should be an easy way to express that in sttp.

That would require adding another generic parameter to `ResponseError `, while default `ResponseError` would be still parametrized by `String` type.

Additional `ResponseError` type is needed in order to represent unmapped error codes.

All existing functions which deals with `ResponseError` (like `asJsonAlwaysUnsafe`) need to be refined.

adamw:
Fixed in v3

PR #620, opened 2020-06-30
Add support for typed http errors

ghostbuster91:
This relates to #618 

### DIFF
--- a/core/src/main/scala/sttp/client/ResponseAs.scala
+++ b/core/src/main/scala/sttp/client/ResponseAs.scala
@@ -122,7 +122,7 @@ object ResponseAs {
     */
   def deserializeRightWithError[E: ShowError, T](
       doDeserialize: String => Either[E, T]
-  ): (Either[String, String], ResponseMetadata) => Either[ResponseError[E], T] = {
+  ): (Either[String, String], ResponseMetadata) => Either[ResponseErrorTyped[String, E], T] = {
     case (Left(s), meta) => Left(HttpError(s, meta.code))
     case (Right(s), _)   => deserializeWithError(doDeserialize)(implicitly[ShowError[E]])(s)
   }
@@ -162,11 +162,11 @@ object ResponseAs {
     }
 }
 
-sealed abstract class ResponseError[+T](error: String) extends Exception(error)
-case class HttpError(body: String, statusCode: StatusCode)
-    extends ResponseError[Nothing](s"statusCode: $statusCode, response: $body")
-case class DeserializationError[T: ShowError](body: String, error: T)
-    extends ResponseError[T](implicitly[ShowError[T]].show(error))
+sealed abstract class ResponseErrorTyped[+HE, +DE](error: String) extends Exception(error)
+case class HttpError[HE](body: HE, statusCode: StatusCode)
+    extends ResponseErrorTyped[HE, Nothing](s"statusCode: $statusCode, response: $body")
+case class DeserializationError[DE: ShowError](body: String, error: DE)
+    extends ResponseErrorTyped[Nothing, DE](implicitly[ShowError[DE]].show(error))
 
 trait ShowError[-T] {
   def show(t: T): String
--- a/core/src/main/scala/sttp/client/SttpApi.scala
+++ b/core/src/main/scala/sttp/client/SttpApi.scala
@@ -263,7 +263,7 @@ trait SttpApi extends SttpExtensions with UriInterpolator {
       } else {
         onError.map {
           case Left(a)  => Left(a)
-          case Right(b) => Left(HttpError(b))
+          case Right(b) => Left(HttpError(b, meta.code))
         }
       }
     }
--- a/core/src/main/scala/sttp/client/SttpApi.scala
+++ b/core/src/main/scala/sttp/client/SttpApi.scala
@@ -244,4 +244,56 @@ trait SttpApi extends SttpExtensions with UriInterpolator {
     */
   def multipart[B: BodySerializer](name: String, b: B): Part[BasicRequestBody] =
     Part(name, implicitly[BodySerializer[B]].apply(b), contentType = Some(MediaType.ApplicationXWwwFormUrlencoded))
+
+  /**
+    * Allows to provide different mappings for success and error responses
+    * @param onError transformation which will be applied if the response is unsuccessful (non 2xx)
+    * @param onSuccess transformation which will be applied if the response is successful (2xx)
+    * @tparam DE underlying type of deserialization failure
+    * @tparam HE type which represents known error cases
+    * @tparam T type of the successful response
+    */
+  def either[DE, HE, T](
+      onError: ResponseAs[Either[DeserializationError[DE], HE], Nothing],
+      onSuccess: ResponseAs[Either[DeserializationError[DE], T], Nothing]
+  ): ResponseAs[Either[ResponseErrorTyped[HE, DE], T], Nothing] = {
+    fromMetadata { meta =>
+      if (meta.isSuccess) {
+        onSuccess
+      } else {
+        onError.map {
+          case Left(a)  => Left(a)
+          case Right(b) => Left(HttpError(b))
+        }
+      }
+    }
+  }
+
+  def fromStatusCode[DE, T](
+      mapping: StatusCode => ResponseAs[Either[DeserializationError[DE], T], Nothing]
+  ): ResponseAs[Either[DeserializationError[DE], T], Nothing] = {
+    fromMetadata { meta =>
+      mapping(meta.code)
+    }
+  }
+
+  /**
+    * Same as #either but throws deserialization error if either success or error deserialization fails
+    * @param onSuccess transformation which will be applied if the response is successful (2xx)
+    * @param onError transformation which will be applied if the response is unsuccessful (non 2xx)
+    * @tparam E type which represents known error cases
+    * @tparam T type of the successful response
+    * @return
+    */
+  def eitherUnsafe[E, T](
+      onSuccess: ResponseAs[T, Nothing]
+  )(onError: ResponseAs[E, Nothing]): ResponseAs[Either[E, T], Nothing] = {
+    fromMetadata { meta =>
+      if (meta.isSuccess) {
+        onSuccess.map(Right(_))
+      } else {
+        onError.map(Left(_))
+      }
+    }
+  }
 }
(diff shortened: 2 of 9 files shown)


json/circe/src/main/scala/sttp/client/circe/SttpCirceApi.scala
```diff
@@ -20,9 +21,27 @@ trait SttpCirceApi {
...
     asString.map(ResponseAs.deserializeRightWithError(deserializeJson))
 
+  /**
+    * If the response is successful (2xx) or there the decoder is defined for given http code, tries to deserialize the body from a string into JSON. Returns:
+    * - `Right(b)` if the parsing was successful
+    * - `Left(HttpError(b))` if the response code was other than 2xx, but a decoder was defined
+    * - `Left(UnknownHttpError(String))` if the response code was other than 2xx and there was no decoder (deserialization is not attempted)
+    * - `Left(DeserializationError)` if there's an error during deserialization
+    */
+  def asJsonWithError[E:IsOption,B: Decoder: IsOption](knownErrorCodes:PartialFunction[StatusCode,Decoder[E]]): ResponseAs[Either[ResponseError[E,io.circe.Error], B], Nothing] = {
```
adamw, line 34: Another option would be to re-use existing combinators:

```
def asJsonTypedErrors[T, E, D](
  success: ResponseAs[Either[ResponseError[String, D], T])( 
  errors: PartialFunction[StatusCode, ResponseAs[Either[ResponseError[String, D], E]]]): ResponseAs[Either[ResponseError[E, D], T]] = 
  fromMetadata { meta =>
    if (meta.isSuccess) success.map(Right(_)) else (...)
  }
```

which can be later used sth like:

```
val myResponse = asJsonTypedErrors(asJson[Success]) {
  case StatusCode.BadRequest => asJson[MyBadRequest]
  case StatusCode.Forbidden => asJson[MyForbidden]
}
```

One thing I don't really like is that the signatures become really complicated. I suspect most of the time you'd want errors to be represented as `String`s. Maybe we could use some type aliases here? E.g.:

```
abstract class ResponseErrorTyped[E, D] extends Exception

type ResponseError[E] = ResponseErrorTyped[E, String]
```

and maybe even json-lib-specific aliases, e.g. `type CirceResponseError[D] = ResponseError[io.circe.Error, D]`

adamw: In fact this way we create a description which can be applied not only to json, so instead of `asJsonTypedErrors` we would have `asTypedErrors`.

And this can come in two variants:

```
asTypedErrors: ResponseAs[Either[ResponseError[E, D], T]] // the most general
asTypedErrorsUnsafe: ResponseAs[Either[E, T]]
```

core/src/main/scala/sttp/client/ResponseAs.scala
```diff
@@ -142,8 +142,7 @@ object ResponseAs {
...
 }
 
-sealed abstract class ResponseError[+T] extends Exception {
-  def body: String
-}
-case class HttpError(body: String) extends ResponseError[Nothing]
-case class DeserializationError[T](body: String, error: T) extends ResponseError[T]
+sealed abstract class ResponseError[+HE,+DE] extends Exception 
+case class HttpError[E](body: E) extends ResponseError[E,Nothing]
+case class UnknownHttpError(body: String) extends ResponseError[Nothing, Nothing]
```
adamw, line 147: maybe we could try reducing the number of cases, removing `UnknownHttpError`. Then, the user would always have to provide total mapping for errors (e.g. using specific status codes + general case)

That's because in the default case of `ResponseError[E, String]`, we would have `HttpError(String)` which would occur on non-2xx responses, and `UnknownHttpError(String)` which would never occur

ghostbuster91: Current implementation returns `UnknownHttpError` if there is no mapping for given status code ( also if mappings are empty). 

Although I am in favor of removing `UnknownHttpError` I am not sure if it is possible because that implies that `HttpError[String]` and `HttpError[MyCustomError]` have a common denominator. 

Specifying generic type as `Any` won't work because we will lose ability to perform exhaustive match.

adamw: Yes, that would require that mappings be total (or rather, so that they include a fallback case for any non-covered status codes). If we have mappings at all, that is - by default we would just use `HttpError[String]`

generated-docs/out/responses/body.md
```diff
@@ -86,6 +86,24 @@ basicRequest
 
 A number of JSON libraries are supported out-of-the-box, see [json support](../json.md).
 
+Sometimes it might be useful to model some http error responses right away. We can do that using `asTypedErrors` method:
+```scala
+case class SuccessModel(...)
+sealed trait MyErrorModel
+case class Conflict(...) extends MyModel
```
chameleon82, line 93: Typo. Seems should be `case class Conflict(...) extends MyErrorModel`

core/src/main/scala/sttp/client/ResponseAs.scala
```diff
@@ -142,8 +142,7 @@ object ResponseAs {
...
 }
 
-sealed abstract class ResponseError[+T] extends Exception {
-  def body: String
-}
-case class HttpError(body: String) extends ResponseError[Nothing]
-case class DeserializationError[T](body: String, error: T) extends ResponseError[T]
+sealed abstract class ResponseErrorTyped[+HE, +DE] extends Exception
+case class HttpError[HE](body: HE) extends ResponseErrorTyped[HE, Nothing]
+case class UnmatchedHttpError(body: String) extends ResponseErrorTyped[Nothing, Nothing]
```
adamw, line 147: I'd still try removing this and keeping only `HttpError` - as by the default String-case you'd need to match both types of errors, which only one could ever occur.

This would also make `ErrorResponseAs` a total function

generated-docs/out/examples.md
```diff
@@ -60,7 +60,7 @@ val request = basicRequest
   .response(asJson[HttpBinResponse])
 
 implicit val backend = AkkaHttpBackend()
-val response: Future[Response[Either[ResponseError[Exception], HttpBinResponse]]] =
+val response: Future[Response[Either[ResponseErrorTyped[Exception], HttpBinResponse]]] =
```
adamw, line 63: this probably doesn't compile? ;)

ghostbuster91: who knows :P

generated-docs/out/responses/body.md
ghostbuster91: Thanks, fixed

core/src/main/scala/sttp/client/ResponseAs.scala
ghostbuster91: Hmm I think that it is more user-friendly to expose api with partialFunction but I understand your argument. Changed.

core/src/main/scala/sttp/client/SttpApi.scala
```diff
@@ -209,4 +208,47 @@ trait SttpApi extends SttpExtensions with UriInterpolator {
...
+
+  /**
+    * Allows to provide additional mappings for error cases based on statusCodes
+    * @param onSuccess transformation which will be applied if the response is successful (2xx)
+    * @param onError transformation which will be applied if the response is unsuccessful (non 2xx)
+    * @tparam DE underlying type of deserialization failure
+    * @tparam HE type which represents known error cases
+    * @tparam T type of the successful response
+    */
+  def asTypedError[DE, HE, T](onSuccess: ResponseAs[Either[DeserializationError[DE], T], Nothing])(
```
adamw, line 220: as this is more of a decorator, maybe a name like `withTypedErrors` would be better?

docs/responses/body.md
```diff
@@ -95,6 +95,24 @@ basicRequest
...
+case class Conflict(...) extends MyErrorModel
+case class ServerError(...) extends MyErrorModel
+
+val myRequest: Request[Either[ResponseErrorTyped[MyErrorModel, io.circe.Error], MyModel], Nothing] =
+  basicRequest
+    .get(uri"https://example.com")
+    .response(asTypedError(asJsonAlways[SuccessModel]){
+      case StatusCode.Conflict => asJsonAlways[Conflict]
+      case StatusCode.ServerError => asJsonAlways[ServerError]
+    })
```
adamw, line 111: now I'm thinking, that while it's good to have the logic centralised in `asTypedError`, maybe it would be more user-friendly to have a `asJsonEither` method as well.

Variant which uses different models for 2xx and different for all others:

```
basicRequest
     .get(uri"https://example.com")
     .response(asJsonEither[MyErrorModel, MyModel])
```

And the more general:

```
basicRequest
     .get(uri"https://example.com")
     .response(asJsonEitherFromStatusCode[MyErrorModel, MyModel]({
       case StatusCode.Conflict => asJsonAlways[Conflict]
       case StatusCode.ServerError => asJsonAlways[ServerError]
     })
```

what do you think>

ghostbuster91: How about:
```
basicRequest
     .get(uri"https://example.com")
     .response(either(asJson[MyErrorModel], asJson[MyModel]))
```
and 
```
basicRequest
     .get(uri"https://example.com")
     .response(either(fromStatusCodes[MyErrorModel]{
       case StatusCode.Conflict => asJsonAlways[Conflict]
       case StatusCode.ServerError => asJsonAlways[ServerError]
     }), asJson[MyModel])
```

While I haven't thought even for a second whether it is possible, I like this structure and it might give additional flexibility since with it we should be able to parse errors from different formats than the main case.

adamw: Yes, that's better :)

ghostbuster91:
Would it be possible to get rid of `unsafe` variant of `responseAs` constructos?
Maybe we should always create a safe variant i.e. variant with type `Either[ResponseErrorTyped[X,Y],T]` and allow for modifications which will eliminate particular errors from it?

Example:
```
asJson[MyModel] // ResponseAs[Either[ResponseErrorTyped[X,Y] MyModel]
.unsafeDeserialize // ResponseAs[Either[HttpError[X], MyModel]
.unsafeHttpError // ResponseAs[MyModel]
```
also it should be possible to perform such elimination in any order:
```
asJson[MyModel] // ResponseAs[Either[ResponseErrorTyped[X,Y] MyModel]
.unsafeHttpError // ResponseAs[Either[DeserializationError[Y], MyModel]
.unsafeDeserialize // ResponseAs[MyModel]
```

PR #635, opened 2020-07-16
Let HttpUrlConnectionBackend allow custom encodings

eugeniyk:
During checking backends for supporting custom encodings found out that we have special logic in `HttpUrlConnectionBackend` which prevents passing custom encoded bodies back to the client.

For consistency with other backends (every of them just let unsupported encoding pass through) here is the fix for `HttpUrlConnectionBackend`

------


eugeniyk:
Closed in favour of #636

PR #636, opened 2020-07-16
Let HttpUrlConnectionBackend allow custom encodings

eugeniyk:
During checking backends for supporting custom encodings found out that we have special logic in HttpUrlConnectionBackend which prevents passing custom encoded bodies back to the client.

For consistency with other backends (every of them just let unsupported encoding pass through) here is the fix for HttpUrlConnectionBackend

-----

### DIFF
--- a/core/src/main/scala/sttp/client/internal/ToCurlConverter.scala
+++ b/core/src/main/scala/sttp/client/internal/ToCurlConverter.scala
@@ -21,7 +21,7 @@ class ToCurlConverter[R <: RequestT[Identity, _, _]] {
 
   private def extractHeaders(r: R): String = {
     r.headers
-    // filtering out compression headers so that the results are human-readable, if possible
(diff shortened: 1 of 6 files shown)


adamw:
Thanks!

However, the mechanism is different in other backends. They allow specifying a `customEncodingHandler: EncodingHandler` (you can try searching for this), which specifies how to deal with encodings. 

By default, all backends `throw new UnsupportedEncodingException(s"Unsupported encoding: $ce")`

Maybe you could try making `HttpURLConnectionBackend` follow suit and support an `EncodingHandler` parameter as well?

eugeniyk:
@adamw got your point, let me change that for `HttpURLConnectionBackend`

eugeniyk:
@adamw I've added test for custom encoding for HttpURLConnectionBackend, can you please help to implement rest of the tests for other backends

adamw:
Released in 2.2.3, thanks!

PR #643, opened 2020-07-24
Obey followRedirects request option

lunaticare:
When used with AsyncHttpClient backend

### DIFF
--- a/async-http-client-backend/src/main/scala/sttp/client/asynchttpclient/AsyncHttpClientBackend.scala
+++ b/async-http-client-backend/src/main/scala/sttp/client/asynchttpclient/AsyncHttpClientBackend.scala
@@ -228,6 +228,7 @@ abstract class AsyncHttpClientBackend[F[_], S](
       .setUrl(r.uri.toString)
       .setReadTimeout(if (readTimeout.isFinite) readTimeout.toMillis.toInt else -1)
       .setRequestTimeout(if (readTimeout.isFinite) readTimeout.toMillis.toInt else -1)
+      .setFollowRedirect(r.options.followRedirects)
     r.headers.foreach { case Header(k, v) => rb.setHeader(k, v) }
     setBody(r, r.body, rb)
     rb.build()


async-http-client-backend/src/main/scala/sttp/client/asynchttpclient/AsyncHttpClientBackend.scala
```diff
@@ -228,6 +228,7 @@ abstract class AsyncHttpClientBackend[F[_], S](
       .setUrl(r.uri.toString)
       .setReadTimeout(if (readTimeout.isFinite) readTimeout.toMillis.toInt else -1)
       .setRequestTimeout(if (readTimeout.isFinite) readTimeout.toMillis.toInt else -1)
+      .setFollowRedirect(r.options.followRedirects)
```
adamw, line 231: This should be handled by `FollowRedirectsBackend` wrapper. Do you see some problems with redirection?

lunaticare: @adamw Yes, encountered a problem, redirect happens regardless of `Request#followRedirects` value. This might be caused by how I create the client. Thank you for clarifying the logic behind request processing, will check my code.

lunaticare: The problem was that I set up redirect flag in native client.  Illustration of my approach (Ammonite):
```scala
import $ivy.`com.softwaremill.sttp.client::async-http-client-backend-future:2.1.4`

import org.asynchttpclient.DefaultAsyncHttpClientConfig
import sttp.client.asynchttpclient.future.AsyncHttpClientFutureBackend
import sttp.client._
import sttp.model._

import scala.concurrent.duration.DurationInt

import scala.concurrent.{Await, Future}
import scala.concurrent.ExecutionContext.Implicits

val url1 = "https://bit.ly/2WXLAEY"
val url2 = "http://google.com/"

val client1: SttpBackend[Future, Nothing, NothingT] = AsyncHttpClientFutureBackend()

val client2: SttpBackend[Future, Nothing, NothingT] = AsyncHttpClientFutureBackend(customizeRequest = { builder =>
  builder.setFollowRedirect(true)
})

val client3: SttpBackend[Future, Nothing, NothingT] =
  new FollowRedirectsBackend[Future, Nothing, NothingT](
    AsyncHttpClientFutureBackend.usingConfigBuilder(
    { cb: DefaultAsyncHttpClientConfig.Builder =>
      cb.setFollowRedirect(true)
    }
    )(Implicits.global)
  )

def getUrl(url: String, client: SttpBackend[Future, Nothing, NothingT]): Int = {
  val request: RequestT[Empty, Either[String, String], Nothing] = sttp.client.basicRequest
    .followRedirects(false)
  val request1 = request.head(uri"$url")
  try {
    val resp = Await.result(client.send(request1), 10.seconds)
    resp.code.code
  } catch {
    case t: Throwable =>
      println(t)
      throw t
  }
}

for {
  url <- Seq(url1, url2)
  (clientName, client) <- Seq(
    ("client1_good_noLowLevelRedirect", client1), 
    ("client2_bad_redirectInLowLevelClient", client2), 
    ("client3_bad_redirectInLowLevelClient", client3))
} yield {
  (clientName, url, getUrl(url, client))
}
```

```
res14: Seq[(String, String, Int)] = List(
  ("client1_good_noLowLevelRedirect", "https://bit.ly/2WXLAEY", 301),
  ("client2_bad_redirectInLowLevelClient", "https://bit.ly/2WXLAEY", 200),
  ("client3_bad_redirectInLowLevelClient", "https://bit.ly/2WXLAEY", 200),
  ("client1_good_noLowLevelRedirect", "http://google.com/", 301),
  ("client2_bad_redirectInLowLevelClient", "http://google.com/", 200),
  ("client3_bad_redirectInLowLevelClient", "http://google.com/", 200)
)
```

adamw: Ah ... well I suppose since we have request-level and backend-level redirect flags one of them has to take priority - in sttp's case it's the request-level one.

PR #648, opened 2020-08-04
Upgrade ZIO dependencies to 1.0.0

ghostdogpr:
ZIO 1.0.0 is out! 🎉 
I disabled dotty because ZIO upgraded to 0.26 but sttp uses 0.23.


adamw:
Released in 2.2.4, thanks!

Issue #656, opened 2020-08-12
It is difficult to determine the ultimate URI of a resource when using `FollowRedirectsBackend`


stephenjudkins:
`FollowRedirectsBackend` works well to transparently follow a chain of 30x redirects to a new resource. However, when at that resource the ultimate URI is difficult to determine. This is problematic because the resource may contain relative URIs, and knowledge of the resource's URI is necessary to canonicalize them. One must follow the chain of redirects in `history` and reproduce the same logic for this to work.

adamw:
Will be released in 3.0.0-RC2

PR #657, opened 2020-08-12
Include URI in `Response`

stephenjudkins:
Fixes https://github.com/softwaremill/sttp/issues/656. Feedback and suggestions welcome.


adamw:
Thanks!

First of all, I would consider if we want to have just the original request's URI on the response - or maybe the whole request? Or the requests's meta-data at least (headers + uri + method, w/o the body)?

Second, that's a breaking change, so would have to go to at least a minor release. But as `v3` isn't that far away (well, the `v3` branch doesn't yet fully compile, but we are getting there), maybe it would make most sense to make the changes against that version.

adamw:
I've added a slightly different implementation in the v3 branch, let me know what you think. I'll close this PR as we can't make the change in v2 anyway.

Issue #666, opened 2020-08-23
Incorrect utf-8 handling in Scala Native


sujeet:
Repro code: https://github.com/sujeet/sttp-unicode-bugreport
Affected version: 2.2.5
Summary below:

When server serves `"ह"`,the JVM sttp client returns it correctly as
```
"ह" (string) or as ByteArray(34, -32, -92, -71, 34)
```
The Scala Native sttp client, however returns it incorrectly as
```
"9" (string) or as ByteArray(34, 57, 34)
```

Following are the repro steps:

### Run the python server that serves a single unicode letter `ह` enclosed in double quotes.
```
python unicodeserver.py
```

### Compile and run the scala native binary
```
sbt build
./target/scala-2.11/unicode-out
```

The following is the output:

```
default: '"9"', length: 3
utf8str: '"9"', length: 3
bytearray: List(34, 57, 34)
```

### Run the equivalent scala script with ammonite
```
amm jvm-version.sc
```

The following is the outut:

```
default: '"ह"', length: 3
utf8str: '"ह"', length: 3
bytearray: List(34, -32, -92, -71, 34)
```

### Expectation
Native code should produce the same output as the JVM version

### Details
```
ammonite: 1.6.9, scala(2.13.0), java(1.8.0_144)
native: sbt(0.13.15), scala(2.11.11), scala-native(0.4.0-M2)
```

### Revelant file content replicated below for quick glance

`unicodeserver.py`
```python
import http.server
import socketserver

PORT = 9988

class Handler(http.server.SimpleHTTPRequestHandler):
  def do_GET(self):
    self.send_response(200)

    self.send_header('Content-type', 'application/json; charset=UTF-8')
    self.send_header('Cache-Control', 'no-store, must-revalidate')
    self.end_headers() 

    bytes_to_send = '"ह"'.encode('utf-8')

    print(bytes_to_send)
    self.wfile.write(bytes_to_send)


def serve(port=PORT):
  with socketserver.TCPServer(('', port), Handler) as httpd:
    httpd.serve_forever()


serve()
```

`jvm-version.sc` (Scala JVM code)
```scala
import $ivy.`com.softwaremill.sttp.client::core:2.2.5`
import sttp.client.quick._

def req = quickRequest.get(uri"http://localhost:9988")

val default = req.send().body
println(s"default: '$default', length: ${default.length}")

val utf8str = req.response(asString("UTF-8")).send().body.right.get
[...]

PR #669, opened 2020-08-26
Fix handling of utf-8 responses in scala-native

sujeet:
#### Problem:
`str.toCharArray.map(_.toByte)`, takes only one byte per
character, thus producing an incorrect `Array[Byte]` for
any string that contains multi-byte characters.

#### Solution:
Use `str.getBytes` instead.

#### Demo:
```scala
val smile = "😀"

smile.toCharArray.map(_.toByte)
// Array[Byte] = Array(61, 0)

smile.getBytes
// Array[Byte] = Array(-16, -97, -104, -128)
```

#### Testing:
This patch also adds an unicode character to the test
string in `SyncHttpTest`. The modified test fails without
the patch, and passes with the patch.

#### Future Work:
Currently `AbstractCurlBackend` calls the `fromCString`
function from scala-native, without the encoding parameter.
Thus, the encoding parameter gets defaulted to UTF-8.
Testing needs to be done in cases where the encoding
of the server response is not UTF-8, which is out of the
scope for this patch.

#### Checklist:

### DIFF
--- a/core/src/main/scalanative/sttp/client/AbstractCurlBackend.scala
+++ b/core/src/main/scalanative/sttp/client/AbstractCurlBackend.scala
@@ -213,7 +213,7 @@ abstract class AbstractCurlBackend[F[_], S](monad: MonadError[F], verbose: Boole
       }
   }
 
-  private def toByteArray(str: String): F[Array[Byte]] = responseMonad.unit(str.toCharArray.map(_.toByte))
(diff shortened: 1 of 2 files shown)


sujeet:
Fixes #666

adamw:
Thanks! Released in [2.2.6](https://github.com/softwaremill/sttp/releases/tag/v2.2.6).

If you'd have further improvements to the native backend, that would be most welcome!

PR #674, opened 2020-09-03
Add more ergonomic zio stubbing

paulpdaniels:

This PR adds a more ergonomic subbing syntax to the ZIO backend. This allows the stubbing to be composed in a for comprehension, example in the tests.

### DIFF
--- a/async-http-client-backend/zio/src/main/scala/sttp/client/asynchttpclient/zio/AsyncHttpClientZioBackend.scala
+++ b/async-http-client-backend/zio/src/main/scala/sttp/client/asynchttpclient/zio/AsyncHttpClientZioBackend.scala
@@ -73,16 +73,16 @@ object AsyncHttpClientZioBackend {
   ): Task[SttpBackend[Task, Stream[Throwable, Byte], WebSocketHandler]] =
     ZIO
       .runtime[Any]
-      .flatMap(
-        runtime =>
-          Task.effect(
-            AsyncHttpClientZioBackend(
-              runtime,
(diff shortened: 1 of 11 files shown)


adamw:
Thanks! Looks nice 👍 

I've tried making the code more generic and moving it to the `zio` module so that similar functionality could be added to the httpclient implementation fo the zio backend, but so far failed. As a backup plan there's always the possibility to copy code, but I'd like to avoid that if possible.

Could you maybe add some short mention in the docs that this is available? Over here: https://github.com/softwaremill/sttp/blob/master/docs/backends/zio.md

Also I see that following the ZIO convention the `send` method should be a package method, not wrapped in the `SttpClient` object. Probably something to fix in v3, if I'm right :)

adamw:
Ok, I think I managed to add the same code to httpclient as well. Let me know what you think :)

paulpdaniels:
Ah looks nice! Tbh I didn't realize there was a second backend implementation of it because we always use the `AsyncHttpBackend` implementations.

I'll add some tests for the httpclient one as well and update the docs

> Also I see that following the ZIO convention the send method should be a package method

Yeah but not a deal breaker since `send` is a little ambiguous in code anyway, but yeah something for v3. Speaking of which, with the upcoming release already in RC will you forward port this into it? Or will it need a redesign at that point?

adamw:
Released in [2.2.8](https://github.com/softwaremill/sttp/releases/tag/v2.2.8), thanks!

I'll migrate to v3

PR #677, opened 2020-09-07
Wrap mutable httpclient builder in monad.eval, so that the request is properly constructed when the effect is re-used

adamw:


adamw:
Released in [2.2.8](https://github.com/softwaremill/sttp/releases/tag/v2.2.8)

PR #679, opened 2020-09-07
Streaming text accumulator

ghostbuster91:

### DIFF
--- a/core/src/test/scalajvm/sttp/client/testing/websocket/WebSocketStreamingTest.scala
+++ b/core/src/test/scalajvm/sttp/client/testing/websocket/WebSocketStreamingTest.scala
@@ -24,120 +24,108 @@ trait WebSocketStreamingTest[F[_], S] extends ToFutureWrapper { outer: Suite wit
   implicit val monad: MonadError[F]
   implicit val convertToFuture: ConvertToFuture[F]
 
-  it should "use pipe to process websocket messages - server-terminated" in {
-    val received = new LinkedBlockingQueue[String]()
+  def webSocketPipeTerminatedByServerTest(
+      postfix: String
+  )(pipe: LinkedBlockingQueue[String] => streams.Pipe[WebSocketFrame.Data[_], WebSocketFrame]) = {
+    it should s"use pipe to process websocket messages - server-terminated - $postfix" in {
+      val received = new LinkedBlockingQueue[String]()
+      basicRequest
+        .get(uri"$wsEndpoint/ws/send_and_expect_echo")
+        .response(
+          asWebSocketStreamAlways(streams)(pipe(received))
+        )
+        .send(backend)
+        .map { _ =>
+          received.asScala.toList shouldBe List("test1", "test2", "test3")
+        }
+        .toFuture()
+    }
+  }
+
+  def webSocketPipeClientTerminated(
+      postfix: String
+  )(pipe: LinkedBlockingQueue[String] => streams.Pipe[WebSocketFrame.Data[_], WebSocketFrame]) = {
+    it should s"use pipe to process websocket messages - client-terminated - $postfix" in {
+      val received = new LinkedBlockingQueue[String]()
+      basicRequest
+        .get(uri"$wsEndpoint/ws/echo")
+        .response(
+          asWebSocketStreamAlways(streams)(pipe(received))
+        )
+        .send(backend)
+        .map { _ =>
+          received.asScala.toList shouldBe List("echo: 1", "echo: 2", "echo: 3", "echo: 4", "echo: 5")
+        }
+        .toFuture()
+    }
+  }
+
+  webSocketPipeTerminatedByServerTest("raw") { received =>
     val buffer = new AtomicReference[String]("")
-    basicRequest
-      .get(uri"$wsEndpoint/ws/send_and_expect_echo")
-      .response(
-        asWebSocketStreamAlways(streams)(
-          functionToPipe(
-            Nil,
-            {
-              case WebSocketFrame.Text(payload, false, _) =>
-                buffer.accumulateAndGet(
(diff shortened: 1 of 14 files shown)


implementations/monix/src/main/scala/sttp/client/impl/monix/MonixWebSockets.scala
```diff
@@ -30,4 +31,28 @@ object MonixWebSockets {
         .guarantee(ws.close())
     }
   }
+  def textAccumulator: (String => WebSocketFrame) => MonixStreams.Pipe[WebSocketFrame, WebSocketFrame] =
```
adamw, line 34: `stringPipe`?

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/HttpClientMonixWebSocketTest.scala
```diff
@@ -24,4 +31,48 @@ class HttpClientMonixWebSocketTest extends WebSocketTest[Task] with WebSocketStr
...
+        )
+      )
+      .send(backend)
+      .map { _ =>
+        received.asScala.toList shouldBe List("test1", "test2", "test3")
+      }
+      .toFuture()
+  }
+
+  it should "use pipe to process websocket messages - client-terminated - text accumulator" in {
```
adamw, line 54: does copying these whole tests add any value? Maybe a single test for the text pipe + some additional unit tests for the conversion itself provide better coverage?

implementations/monix/src/main/scala/sttp/client/impl/monix/MonixWebSockets.scala
ghostbuster91: technically that is rather a pipe transformation than a pipe

adamw: `fromTextPipe`?

httpclient-backend/monix/src/test/scala/sttp/client/httpclient/monix/HttpClientMonixWebSocketTest.scala
ghostbuster91: hmm I am not sure because to be honest I was totally uncertain whether such a complex combination of observable's operators won't break anything. Also I wanted to see how hard/easily would it be to send additional message in the beginning as in the `client terminated` test.

adamw: ok, maybe we should then extract the common parts of these tests and just pass in the stream?

ghostbuster91: yes, I think so

docs/websockets.md
```diff
@@ -54,3 +54,6 @@ def asWebSocketStreamAlways[S](s: Streams[S])(p: s.Pipe[WebSocketFrame.Data[_],
 ```
 
 Using streaming websockets requires the backend to support the given streaming capability (see also [streaming](requests/streaming.md)). Streaming capabilities are described as implementations of `Streams[S]`, and are provided by backend implementations, e.g. `AkkaStreams` or `Fs2Streams[F]`.
+
+When working with streams of websocket frames keep in mind that a text payload maybe fragmented into multiple frames.
+STTP provides two useful methods (`fromTextPipe`, `fromTextPipeF`) for each backend to aggregate these fragments back into complete messages.
```
adamw, line 59: `sttp` (lower case) & it would probably be good to mention where these functions are (in MonixWebSockets etc.), as finding thme might be non-trivial

implementations/monix/src/main/scala/sttp/client/impl/monix/MonixWebSockets.scala
```diff
@@ -30,4 +31,26 @@ object MonixWebSockets {
...
+  def fromTextPipeF: MonixStreams.Pipe[String, WebSocketFrame] => MonixStreams.Pipe[WebSocketFrame, WebSocketFrame] =
+    p => p.compose(combinedTextFrames)
+
+  def combinedTextFrames: MonixStreams.Pipe[WebSocketFrame, String] = { input =>
+    Observable
+      .fromTask(
+        input
+          .takeWhileInclusive { case WebSocketFrame.Text(_, finalFragment, _) => !finalFragment }
+          .collect { case text: WebSocketFrame.Text => text }
+          .toListL
```
adamw, line 46: maybe we can rewrite this as stream operations, without compiling the stream over and over again? I suspect this might work for websocket streams as they are backend by a mutable data structure (the queue of incoming messages), but would break e.g. for strict streams

core/src/test/scalajvm/sttp/client/testing/websocket/WebSocketStreamingTest.scala
```diff
@@ -24,76 +24,112 @@ trait WebSocketStreamingTest[F[_], S] extends ToFutureWrapper { outer: Suite wit
...
+  }
+
+  def prepend(item: WebSocketFrame.Text)(
+      to: streams.Pipe[WebSocketFrame.Data[_], WebSocketFrame]
+  ): streams.Pipe[WebSocketFrame.Data[_], WebSocketFrame]
+
+  def fromTextPipe(function: String => WebSocketFrame): streams.Pipe[WebSocketFrame.Data[_], WebSocketFrame]
+
   def functionToPipe(
       initial: List[WebSocketFrame.Data[_]],
```
adamw, line 134: maybe we don't need this `initial` parameter and could use `prepend` instead?

PR #713, opened 2020-10-07
Implement non-blocking gunzip and deflate in HttpClientMonixBackend

ghostbuster91:

This build uses a snapshot version of monix (since compression hasn't been released yet), just to verify that all the things work together. Once confirmed we might want to trigger a monix release. 

It seems that we'll  need to revisit our decision of using `ByteBuffer` as an internal representation of chunk in monix streams, because both monix and monix-nio take different approach - they use `Array[Byte]`. How one is better than another?

Side note: This snapshot version also contains `bufferWhileInclusive` operator which could greatly simplify `MonixWebSocket` code, but it doesn't work for some reason. A further debugging will be needed.

### DIFF
--- a/httpclient-backend/monix/src/main/scala/sttp/client3/httpclient/monix/HttpClientMonixBackend.scala
+++ b/httpclient-backend/monix/src/main/scala/sttp/client3/httpclient/monix/HttpClientMonixBackend.scala
@@ -62,7 +62,7 @@ class HttpClientMonixBackend private (
   override protected def publisherToBody(p: Publisher[util.List[ByteBuffer]]): Observable[ByteBuffer] = {
     Observable
       .fromReactivePublisher(p)
-      .map(list => ByteBuffer.wrap(list.asScala.toList.flatMap(_.safeRead()).toArray))
+      .flatMapIterable(list => list.asScala.toList)
   }
 
   override protected def emptyBody(): Observable[ByteBuffer] = Observable.empty
--- a/httpclient-backend/monix/src/main/scala/sttp/client3/httpclient/monix/HttpClientMonixBackend.scala
+++ b/httpclient-backend/monix/src/main/scala/sttp/client3/httpclient/monix/HttpClientMonixBackend.scala
@@ -62,7 +62,7 @@ class HttpClientMonixBackend private (
   override protected def publisherToBody(p: Publisher[util.List[ByteBuffer]]): Observable[ByteBuffer] = {
     Observable
       .fromReactivePublisher(p)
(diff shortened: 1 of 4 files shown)


adamw:
👍 So we're waiting for a monix release?

adamw:
The rationale behind using `ByteBuffers` I think was that it's the internal representation used by async-http-client, so if possible we'd like to minimise conversions. But if that's not possible, we can equally well use byte arrays.

httpclient-backend/monix/src/main/scala/sttp/client3/httpclient/monix/HttpClientMonixBackend.scala
```diff
@@ -49,47 +50,32 @@ class HttpClientMonixBackend private (
...
     Task.eval(new MonixSimpleQueue[T](None))
 
-  override protected def publisherToBody(p: Publisher[util.List[ByteBuffer]]): InputStream = {
-    val subscriber = new InputStreamSubscriber
-    p.subscribe(subscriber)
-    subscriber.inputStream
+  override protected def publisherToBody(p: Publisher[util.List[ByteBuffer]]): Observable[ByteBuffer] = {
+    Observable
+      .fromReactivePublisher(p)
+      .map(list => ByteBuffer.wrap(list.asScala.toList.flatMap(_.safeRead()).toArray))
```
adamw, line 65: this can probably be optimized, so that e.g. for a single-element list we don't convert it to array and back to a buffer.

Also, for multiple buffers, we could probably create an observable which emits all of these buffers, instead of emitting one combined array

ghostbuster91:
> The rationale behind using ByteBuffers I think was that it's the internal representation used by async-http-client, so if possible we'd like to minimise conversions. But if that's not possible, we can equally well use byte arrays.

The biggest problem with it is during the decoding phase.
```
  override protected def standardEncoding: (Observable[ByteBuffer], String) => Observable[ByteBuffer] = {
    case (body, "gzip")    => body.map(_.safeRead()).transform(gunzip()).map(ByteBuffer.wrap)
    case (body, "deflate") => body.map(_.safeRead()).transform(inflate()).map(ByteBuffer.wrap)
    case (_, ce)           => throw new UnsupportedEncodingException(s"Unsupported encoding: $ce")
  }
```
The only way to minimize this conversions is to have a `gunzip` operator which acts on `ByteBuffer`

adamw:
Right, I suppose most requests are compressed. Well, if byte arrays are the "standard" way of chunking in monix, let's switch to that

adamw:
Please reopen against `master` after the monix release

PR #746, opened 2020-11-08
Fix handling of host header in finagle-backend

sloanesturz:
Prior to this change, the finagle backend did not handle the `Host:` HTTP header correctly.

Invoking a request to a URL (for example, directly at an IP address) and setting a `Host:` header would result in two hosts being sent. This is because of a weird quirk in the FinagleRequestBuilder.

Example:

```
basicRequest
  .get(uri"127.0.0.1/my-resource")
  .header("Host", "my-service-name")
  .send()
```

Would create an invalid HTTP request with two host headers, a la:

```
  curl 127.0.0.1/my-resource \
    -H "Host: 127.0.0.1" \
    -H "Host: my-service-name"
```

This isn't what we want, nor is it actually a valid HTTP request at all.

For context, my datacenter has a mesh proxy setup where each service connects
to other services via a proxy, who delegates to the right service via the HTTP
`Host` header. Without this change, I cannot use STTP for my service-to-service
communication, since the Finagle backend is doing this unintuitive thing.

### DIFF
--- a/finagle-backend/src/main/scala/sttp/client3/finagle/FinagleBackend.scala
+++ b/finagle-backend/src/main/scala/sttp/client3/finagle/FinagleBackend.scala
@@ -35,7 +35,7 @@ import sttp.client3.{
   SttpClientException,
   WebSocketResponseAs
 }
-import sttp.model.{Header, Method, Part, StatusCode, Uri}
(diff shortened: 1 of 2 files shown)


adamw:
Released in [3.0.0-RC9](https://github.com/softwaremill/sttp/releases/tag/v3.0.0-RC9), thanks!

PR #763, opened 2020-11-22
Don't explicitly set Content-Type on multipart/form-data requests

sbrunk:
This came up in https://github.com/softwaremill/tapir/pull/860

### DIFF
--- a/core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
+++ b/core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
@@ -25,7 +25,7 @@ import sttp.client3.internal.{SttpFile, _}
 import sttp.client3.ws.{NotAWebSocketException, GotAWebSocketException}
 import sttp.monad.MonadError
 import sttp.monad.syntax._
-import sttp.model.{Header, StatusCode}
+import sttp.model.{Header, HeaderNames, MediaType, StatusCode}
 
 import scala.collection.immutable.Seq
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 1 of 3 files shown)


core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
```diff
@@ -80,8 +80,10 @@ abstract class AbstractFetchBackend[F[_], S <: Streams[S], P](
     }
 
     val rheaders = new JSHeaders()
-    request.headers.foreach { case Header(name, value) =>
-      rheaders.set(name, value)
+    request.headers.foreach { case header @ Header(name, value) =>
+      // for multipart/form-data requests dom.FormData is responsible for setting the Content-Type header
+      // as it will also compute and set the boundary for the different parts, so we have to leave it out here
+      if (header != Header.contentType(MediaType.MultipartFormData)) rheaders.set(name, value)
```
adamw, line 86: 1. There's a `Headers.is(String)` method for a case-insensitive check of the header's name - we probably should use it here
2. I was wondering why no tests failed before :) And looking closer, there's a `send a multipart message with custom content type` test which is disabled for Scala.JS (see `FetchHttpTest`). Maybe we should add another test, which explicitly sets the content-type header to `MultipartFormData` (+ a multipart body)?
3. Shouldn't we ignore all multipart content-type headers for scala.js? That is, just looking at the value's prefix, if it's not `multipart/` Or maybe if the content-type is set to something different than `multipart/form-data` (which we are able to support), we should throw an exception - as we can't send a request which has e.g. `multipart/mixed` set. An exception is not nice, but maybe better than silently filtering things out.

sbrunk: I've just pushed an update that should address your remarks. I went with your suggestion of throwing an exception for multipart types other than `multipart/form-data` This is also consistent with the behaviour in case of unsupported streaming bodies.

adamw:
Thanks! Released in v3.0.0-RC10

PR #780, opened 2020-12-02
Add new test server implementation to make tests independent of particular implementation

adamsmo:
Still WIP as there are few failing tests


adamw:
Thanks, but we'll stay with the current implementation - it's good enough

PR #781, opened 2020-12-03
Add support for server sent events via response asStream

adamsmo:

docs are still WIP, but the implementation is finished


akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpServerSentEvents.scala
```diff
@@ -0,0 +1,17 @@
...
+import akka.stream.scaladsl.{Flow, Framing}
+import akka.util.ByteString
+import sttp.client3.sse.ServerSentEvent
+
+object AkkaHttpServerSentEvents {
+
+  def decodeSSE: Flow[ByteString, ServerSentEvent, NotUsed] =
+    //todo decide on max frame length
+    Framing
+      .delimiter(ByteString("\n\n"), maximumFrameLength = Int.MaxValue, allowTruncation = true)
```
adamsmo, line 13: `maximumFrameLength` can be parameter accepted by `decodeSSE` method

adamw: Is there anything specified by the SSE standard? Maximum message size? If not, we should make this configurable, with some reasonable default - maybe taken from other impls?

httpclient-backend/zio/src/main/scala/sttp/client3/httpclient/zio/ZioServerSentEvents.scala
```diff
@@ -0,0 +1,15 @@
+package sttp.client3.httpclient.zio
+
+import sttp.capabilities.zio.ZioStreams
+import sttp.client3.sse.ServerSentEvent
+import zio.stream.ZTransducer
+
+object ZioServerSentEvents {
```
adamw, line 7: this should go to `implementations/zio` as well, like with `fs2` (thus avoiding duplication)

testing/server/src/main/scala/sttp/client3/testing/server/HttpServer.scala
```diff
@@ -105,6 +106,29 @@ private class HttpServer(port: Int, info: String => Unit) extends AutoCloseable
           parameterMap { _ => entity(as[String]) { (body: String) => complete(body) } }
         }
       }
+    } ~ pathPrefix("sse") {
+      path("echo") {
```
adamw, line 110: it's not really an `echo` :) but rather `replicate3` or sth like that

adamw: or an `echo3`

core/src/main/scala/sttp/client3/sse/ServerSentEvent.scala
```diff
@@ -0,0 +1,31 @@
...
+
+case class ServerSentEvent(
+    data: Option[String] = None,
+    eventType: Option[String] = None,
+    id: Option[String] = None,
+    retry: Option[Int] = None
+)
+
+object ServerSentEvent {
+  def parseEvent(event: List[String]): ServerSentEvent = {
```
adamw, line 13: maybe add some tests?
it's not possible for parsing to fail - there are no invalid inputs?

adamw:
@adamsmo Are you planning to continue working on this PR, or should I take over? :)

adamsmo:
Please take over, responding to your last comment

>maybe add some tests?
it's not possible for parsing to fail - there are no invalid inputs?

It should not be possible for parsing to fail, as it ignores any value with not recognized label. If the line does not contain any `:` it is treated as a line with label and no value, and if the label is unknown, it is ignored.

I think I do not have push permission to this branch after I was removed from the SML organization

adamw:
Released in [3.0.0-RC13](https://github.com/softwaremill/sttp/releases/tag/v3.0.0-RC13)

PR #787, opened 2020-12-23
Separate RequestT into two classes: Request and PartialRequest

adamw:
This is a draft PR mainly to gather feedback on this change.

Currently, there's a single `RequestT[U, T, R]` case class which represents both partial requests (where the method & uri are unknown), and full requests (where the method & uri are specified). This is differentiated using the `U` type parameter which can either be `Empty = None.type` or `Identity`. There are type aliases e.g. `type PartialRequest[T, R] = Request[Empty, T, R]`.

An alternative design would be to have two top-level case classes: `Request` and `PartialRequest`, the first with method, uri, headers, response spec, options, and the second with only headers, response spec and options. Common operations are grouped into a `ResponseOps` trait, which is extended by both classes, and defines a `ThisType` type which is the resulting type for common operations. See the `Request` and `RequestT` files in the diff for details.

Both solutions have pros and cons, hence I'm not sure if introducing this change to sttp client3 makes sense. Feedback welcome! :)

Pros of separate classes:
* proper types instead of type aliases
* no `RequestT` type, which might be confusing initially
* no need to understand what `RequestT[Empty, ...]` means, where `Empty = None.type`
* less implicits - no need to have `isIdInRequest` witnesses so that `send` can be called on on full requests. Here, `send()` is simply only available on the `Request` class, and absent on the `PartialRequest` class
* less type parameters
* just as `Response` implements `ResponseMetadata`, `Request` can now implement `RequestMetadata`, which can be a common trait for requests in tapir & client (not sure if that's useful though, yet)

Cons of separate classes:
* some duplication of data fields
* duplication of request/partial request extension methods (but maybe this can be fixed)
* the common methods in `RequestOps` need to use abstract `withBody`, `withHeaders` etc. methods, where implementations are simple copy calls on case classes
* in `RequestOps`, method calls can't be chained - as each one returns a `ThisType[T, R]`, so calling two methods results in `ThisType[T, R]#ThisType[T, R]`. Currently this is solved using casting, but a better solution would be to introduce dedicated abstract methods, e.g. `withBodyAndMetadata(...)` - there aren't that many of them


adamw:
Closing as neither me nor nobody else seems to be convinces that this is a good idea

PR #804, opened 2021-01-06
add setOpenTracingParentSpanContext to OpenTracingBackend

arun0009:

Adding `setOpenTracingParentSpanContext` to `OpenTracingBackend` as we don't usually have the current span available and have to build span context from B3 headers (e.g. in the case of Zipkin). 

If we create a new Span just so we can pass it to `OpenTracingBackend.setOpenTracingParentSpan` we have unnecessary additional trace/noise in tracing output. 

Also, there's no way to get Span out from SpanContext per this issue https://github.com/opentracing/specification/issues/81.


adamw:
Released in 3.0.0-RC15, thanks!

PR #853, opened 2021-02-08
Issue/619

kubinio123:


armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -0,0 +1,198 @@
...
+  type PE = Any with capabilities.Effect[Future]
+
+  override def send[T, R >: PE](request: Request[T, R]): Future[Response[T]] =
+    adjustExceptions(request)(execute(request))
+
+  private def execute[T, R >: PE](request: Request[T, R]): Future[Response[T]] =
+    WebClient
+      .builder()
+      .responseTimeout(Duration.ofMillis(request.options.readTimeout.toMillis))
+      .build()
```
adamw, line 64: shouldn't we build the `WebClient` once and re-use it for each request?

We might in fact parametrise the backend with the client and allow the user to provide a custom one (in the factory methods)

build.sbt
```diff
@@ -555,6 +555,16 @@ lazy val finagleBackend = (projectMatrix in file("finagle-backend"))
   .jvmPlatform(scalaVersions = scala2)
   .dependsOn(core % compileAndTest)
 
+lazy val armeriaBackend = (projectMatrix in file("armeria-backend"))
```
adamw, line 558: the project also needs to be added to root aggregates, so that it's included when running tests from root

build.sbt
```diff
@@ -555,6 +555,16 @@ lazy val finagleBackend = (projectMatrix in file("finagle-backend"))
...
+  .settings(commonJvmSettings)
+  .settings(testServerSettings)
+  .settings(
+    name := "armeria-backend",
+    libraryDependencies += "com.linecorp.armeria" % "armeria" % "1.4.0"
+  )
+  .jvmPlatform(scalaVersions = List(scala2_13))
```
adamw, line 565: I suppose we can support all scala 2 & scala 3 versions?

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -0,0 +1,198 @@
...
+
+  override implicit def responseMonad: MonadError[Future] = new FutureMonad()(ec)
+}
+
+object ArmeriaBackend {
+  def apply()(implicit ec: ExecutionContext = ExecutionContext.global): SttpBackend[Future, Any] =
+    new FollowRedirectsBackend[Future, Any](new ArmeriaBackend()(ec))
+
+  private val PseudoHeaderPrefix = ":"
+  private val MultipartNotSupportedException = new IllegalArgumentException("Multipart body is not supported")
```
adamw, line 196: I'd create instances of the exceptions at the use-site, then at least some part of the stack trace is useful

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
kubinio123: good point

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -0,0 +1,198 @@
...
+      Future.failed(StreamingNotSupportedException)
+
+    override protected def handleWS[T](
+        responseAs: WebSocketResponseAs[T, _],
+        meta: ResponseMetadata,
+        ws: Nothing
+    ): Future[T] = ws
+
+    override protected def cleanupWhenNotAWebSocket(response: InputStream, e: NotAWebSocketException): Future[Unit] =
+      Future(response.close())
```
adamw, line 180: I suspect that we should close the web client, if we created it

kubinio123: WebClient from Armeria has no close() like methods, it only holds http related config. I haven't found any mentions about closing it in Armenia docs neither.

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
kubinio123: ok

build.sbt
kubinio123: added

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
kubinio123: I parameterized backend with client, but when it's not provided we have to build a new one since it's immutable and we want to adjust timeout based on options from our domain Request object

build.sbt
kubinio123: I used `FutureConverters`, and since it's located in different packages in versions prior to 2.13 i didn't add them here

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -51,22 +47,28 @@ import scala.io.Source
...
       .execute(requestToArmeria(request))
       .aggregate()
       .asScala
       .flatMap(responseFromArmeria(request, _))
 
+  private def getClient(request: Request[_, _]): WebClient =
+    client.getOrElse(
+      WebClient
+        .builder()
+        .writeTimeout(Duration.ofMillis(request.options.readTimeout.toMillis))
```
adamw: So to override the read timeout we need to build a new client? This won't allocate any new thread pools / connections - how are connections shared?

It wouldn't be good to allocate a new connection pool for each request

kubinio123: Clients are build using default factory which shares a common connection pool with them so it's ok.
However there's a "hack" which allows to reuse instance of WebClient using Armeria delegates mechanism, but the drawback is that the timeout value would have to be stored in headers. Otherwise delegates cannot access it.

docs/backends/armeria.md
```diff
@@ -0,0 +1,21 @@
+# Armeria backend
```
adamw, line 1: I'd add a section to `Future-based backends` instead. There's nothing that is *that* special abour armeria that it would need a separate section

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -0,0 +1,201 @@
...
+    val method = methodToArmeria(request.method)
+    val path = request.uri.toString()
+    val headers = headersToArmeria(request.headers, method, path)
+
+    (request.body match {
+      case NoBody                 => HttpRequest.of(method, path)
+      case StringBody(s, _, _)    => HttpRequest.of(method, path, PLAIN_TEXT, s)
+      case FileBody(f, _)         => HttpRequest.of(method, path, OCTET_STREAM, Files.readAllBytes(f.toPath))
+      case ByteArrayBody(b, _)    => HttpRequest.of(method, path, OCTET_STREAM, b)
+      case InputStreamBody(is, _) => HttpRequest.of(method, path, OCTET_STREAM, is.readAllBytes())
```
adamw, line 89: there's no option to pass an input stream?

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
```diff
@@ -0,0 +1,201 @@
...
+      case ByteBufferBody(b, _)   => HttpRequest.of(method, path, OCTET_STREAM, UTF_8.decode(b).toString)
```
adamw, line 90: we're still doing a conversion to string here. It we have to convert at all, let's covert to a byte array

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
kubinio123: no, only bytes can be passed here

armeria-backend/src/main/scala/sttp/client3/armeria/ArmeriaBackend.scala
kubinio123: In case of `ByteBufferBody` since `ByteBuffer` b can be readonly we cannot access it using `b.array()` or `b.duplicate().array()` since both of this calls end up with `ReadOnlyBufferException`. Decoding is the only solution i have found for that case.

adamw: There's an util for that it turns out :) See `sttp.client3.internal.RichByteBuffer.safeRead`

kubinio123: nice, it works :)

ikhoon:
I missed this PR. Armeria supports Reactive Streams that could send a `File` or `InputStream` or `StreamBody` with backpressure.
Let me send a f/u PR for that.

PR #857, opened 2021-02-12
updated docs with info about supporting host headers overrides

kubinio123:
removed unused import

### DIFF
--- a/docs/backends/finagle.md
+++ b/docs/backends/finagle.md
@@ -15,8 +15,6 @@ val backend = FinagleBackend()
 
 This backend depends on [finagle](https://twitter.github.io/finagle/), and offers an asynchronous backend, which wraps results in Twitter's `Future`.
 
-This backend supports host header override.
-
 Please note that: 
 
 * the backend does not support `SttpBackendOptions`, that is specifying proxy settings (proxies are not implemented in http4s, see [this issue](https://github.com/http4s/http4s/issues/251)), as well as configuring the connect timeout 
--- a/docs/backends/finagle.md
+++ b/docs/backends/finagle.md
@@ -15,6 +15,8 @@ val backend = FinagleBackend()
 
 This backend depends on [finagle](https://twitter.github.io/finagle/), and offers an asynchronous backend, which wraps results in Twitter's `Future`.
 
+This backend supports host header override.
+
 Please note that: 
 
 * the backend does not support `SttpBackendOptions`, that is specifying proxy settings (proxies are not implemented in http4s, see [this issue](https://github.com/http4s/http4s/issues/251)), as well as configuring the connect timeout 
--- a/generated-docs/out/backends/summary.md
+++ b/generated-docs/out/backends/summary.md
@@ -19,29 +19,30 @@ Each backend has two type parameters:
 Below is a summary of all the JVM backends; see the sections on individual backend implementations for more information:
 
 ```eval_rst
-==================================== ================================ ================================================= ========================== ===================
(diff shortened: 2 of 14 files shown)


kubinio123:
Should I add it also to docs in previous versions?

generated-docs/out/backends/summary.md
```diff
@@ -19,29 +19,30 @@ Each backend has two type parameters:
...
-``HttpClientFutureBackend``          ``scala.concurrent.Future``      n/a                                               yes (regular)              no
-``HttpClientMonixBackend``           ``monix.eval.Task``              ``monix.reactive.Observable[ByteBuffer]``         yes (regular & streaming)  yes
-``HttpClientFs2Backend``             ``F[_]: cats.effect.Concurrent`` ``fs2.Stream[F, Byte]``                           yes (regular & streaming)  yes
-``HttpClientZioBackend``             ``zio.Task``                     ``zio.stream.Stream[Throwable, Byte]``            yes (regular & streaming)  yes
-``FinagleBackend``                   ``com.twitter.util.Future``      n/a                                               no                         no
-==================================== ================================ ================================================= ========================== ===================
+==================================== ================================ ================================================= ========================== =================== ===============================
```
adamw, line 22: you need to modify the `summary.md` in `docs` - this is the directory with the generated source (after processing by mdoc). So this will be overwritten on the next release.

Also, I would add the "support host overrides" note as a side-note for particular backends, or a side-note somewhere later in this page. This is a summary table which aims to help people choose the right one. I doubt supporting host overrides would be a decisive factor most of the time :)

adamw:
> Should I add it also to docs in previous versions?

No need, just latest is fine

generated-docs/out/backends/summary.md
```diff
@@ -19,29 +19,30 @@ Each backend has two type parameters:
...
+Class                                Effect type                      Supported stream type                             Supports websockets        Fully non-blocking  Supports host header overrides 
+==================================== ================================ ================================================= ========================== =================== ===============================
+``HttpURLConnectionBackend``         None (``Identity``)              n/a                                               no                         no                  no
```
adamw, line 25: also for the built-in-java backends, a note on how to override this behavior would be good (there's some system property), so that the documentation is helpful

generated-docs/out/backends/summary.md
kubinio123: I moved this info to specific backend sections

generated-docs/out/backends/summary.md
kubinio123: I Added note about system property for HttpURLConnection and HttpClient backend sections

adamw:
Tests failed:

```
[info] - should not send the URL's hostname as the host header *** FAILED ***
421
[info]   "Timeout-Access-><function1>,Accept-Encoding->gzip, deflate,User-Agent->Java/11.0.10,Host->localhost:51823,Accept->text/html, image/gif, image/jpeg, */*;q=0.2, */*;q=0.2,Connection->keep-alive" did not include substring "Host->test.com" (HttpTest.scala:273)
```

docs/backends/finagle.md
```diff
@@ -15,6 +15,8 @@ val backend = FinagleBackend()
 
 This backend depends on [finagle](https://twitter.github.io/finagle/), and offers an asynchronous backend, which wraps results in Twitter's `Future`.
 
+This backend supports host header override.
```
adamw, line 18: As every backend supports it in some way, let's remove this - it doesn't really add any new information. The suprising behavior is when this might *not* be the case, so let's document that.

PR #861, opened 2021-02-16
SSL example for HttpUrlConnection

kubinio123:

### DIFF
--- a/docs/conf/ssl.md
+++ b/docs/conf/ssl.md
@@ -44,7 +44,6 @@ Created `ssl` instance will be used in examples below.
 
 Using `SSLContext` from [first section](#ssl-context) define a function to customize connection.
 
-Implement `HostnameVerifier`. Accepting all hosts instance is only for demonstration purposes.
 ```scala
 import sttp.client3._
 import java.net.HttpURLConnection
--- a/docs/conf/ssl.md
+++ b/docs/conf/ssl.md
@@ -28,26 +28,26 @@ import javax.net.ssl._
 
 Initialize `KeyManagerFactory`:
 ```scala
-val keyStore = KeyStore.getInstance(KeyStore.getDefaultType)
-keyStore.load(new FileInputStream("/path/to/your_cert.p12"), "pass".toCharArray)
+val ks: KeyStore = KeyStore.getInstance(KeyStore.getDefaultType)
+ks.load(new FileInputStream("/path/to/your_cert.p12"), "pass".toCharArray)
 
-val keyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
-keyManagerFactory.init(keyStore, "pass".toCharArray)
+val kmf: KeyManagerFactory = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm)
+kmf.init(ks, "pass".toCharArray)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/conf/ssl.md
```diff
@@ -2,9 +2,121 @@
...
 
-* `HttpUrlConnectionBackend`: when creating the backend, specify the `customizeConnection: HttpURLConnection => Unit` parameter, and set the hostname verifier & SSL socket factory as required
-* akka-http: when creating the backend, specify the `customHttpsContext: Option[HttpsConnectionContext]` parameter. See [akka-http docs](http://doc.akka.io/docs/akka-http/current/scala/http/server-side/server-https-support.html)
-* async-http-client: create a custom client and use the `setSSLContext` method
-* OkHttp: create a custom client modifying the SSL settings as described [on the wiki](https://github.com/square/okhttp/wiki/HTTPS)
+## SSL Context
+
+Common requirement for handling SSL is creating `SSLContext`. It's required by several backends.
+
+Example assumes that you have your client key store in `.p12` format and the server certificate imported to your trust store.
```
adamw, line 11: Looks great! Couple of questions:
* what is described here is importing certificates for a server, and hence implementing *one way SSL*, right? If so, it would be good to explicitly mention that - I suspect people might look for docs on "how to implement one way SSL" or "how to implement mutual SSL"
* sometimes you need to disable SSL - only during development of course ;) - I think a section on that would be helpful as well

kubinio123: - It's actually a mutual SSL example since we have both client key store (`.p12` file) and trust store with server certificate imported. So server identifies client and vice versa. I can add an example without trust store (one way)
- I could add a code sample on how to disable SSL

adamw: great :) I think it would be best to have sections on one-way / mutual / disabled SSL - I'm guessing that's what people will be looking for in the docs in the first place

kubinio123: Done, but i didn't add explicit section about disabled SSL because that depends on server config as well.
On the client side we can disable validation of server certificates and that's included in "One way SSL" section.

docs/conf/ssl.md
```diff
@@ -2,9 +2,134 @@
...
+Common requirement for handling SSL is creating `SSLContext`. It's required by several backends.
+
+### One way SSL
+
+Example assumes that you have your client key store in `.p12` format. If you have your credentials in `.pem` format covert them using:
+
+`openssl pkcs12 -export -inkey your_key.pem -in your_cert.pem -out your_cert.p12`
+
+Sample code might look like this:
+```scala
```
adamw, line 18: can we enable mdoc here? (using scala mdoc:silent)

that would verify that the code compiles (at least :) ) see the readme how to run docs generation localy

docs/conf/ssl.md
adamw: looks good, thank you!

docs/conf/ssl.md
kubinio123: I can use mdoc:compile-only here but variables are out of scope for mdoc in further code samples so I haven't used it at first

adamw: According to the docs: https://scalameta.org/mdoc/docs/modifiers.html, `mdoc:silent` accumulates the definitions. Would be good to enable it as much as possible, it really helps in keeping the docs up-to-date :)

PR #877, opened 2021-02-23
add charset to content-type headers in multipart body

dkarwacki:

### DIFF
--- a/core/src/test/scala/sttp/client3/testing/HttpTest.scala
+++ b/core/src/test/scala/sttp/client3/testing/HttpTest.scala
@@ -1,16 +1,16 @@
 package sttp.client3.testing
 
-import java.io.{ByteArrayInputStream, UnsupportedEncodingException}
-import java.nio.ByteBuffer
-
 import org.scalatest._
 import org.scalatest.freespec.AsyncFreeSpecLike
 import org.scalatest.matchers.should.Matchers
+import sttp.client3.internal.{Iso88591, Utf8}
 import sttp.client3.testing.HttpTest.endpoint
 import sttp.client3.{Response, ResponseAs, SttpBackend, _}
 import sttp.model.StatusCode
 import sttp.monad.MonadError
 
+import java.io.{ByteArrayInputStream, UnsupportedEncodingException}
+import java.nio.ByteBuffer
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
@@ -48,6 +48,7 @@ trait HttpTest[F[_]]
   protected def supportsSttpExceptions = true
   protected def supportsMultipart = true
   protected def supportsCustomMultipartContentType = true
+  protected def supportsCustomMultipartEncoding = true
   protected def supportsCustomContentEncoding = false
   protected def throwsExceptionOnUnsupportedEncoding = true
   protected def supportsHostHeaderOverride = true
@@ -406,6 +407,29 @@ trait HttpTest[F[_]]
         req.send(backend).toFuture().map { resp =>
           resp.body should include("f1")
           resp.body should include("v1")
+        }
+      }
+
+      if (supportsCustomMultipartEncoding) {
+        "send a multipart message containing body parts with explicitly set encodings" in {
+
+          val utf8Char = "\u00f6" //ö
+          val replacementChar = "\ufffd" //�
+
+          val req = basicRequest
+            .post(uri"$endpoint/multipart/content_type")
+            .response(asStringAlways)
+            .multipartBody(
+              multipart("p1", utf8Char, Utf8),
+              multipart("p2", utf8Char, Iso88591)
+            )
+
+          req.send(backend).toFuture().map { resp =>
+            resp.body should be(
+              //replacementChar should be used in place of utf8Char for the part with Iso88591 encoding
+              s"p1=$utf8Char$defaultFileName content-type: text/plain; charset=UTF-8, p2=$replacementChar$defaultFileName content-type: text/plain; charset=ISO-8859-1"
+            )
+          }
         }
       }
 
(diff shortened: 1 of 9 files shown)


core/src/test/scalajs/sttp/client3/testing/AbstractFetchHttpTest.scala
```diff
@@ -27,6 +27,8 @@ abstract class AbstractFetchHttpTest[F[_], +P] extends HttpTest[F] {
 
   override protected def supportsHostHeaderOverride = false
 
+  override protected def supportsCharsetControl = false
```
adamw, line 30: This should probably say sth about multiparts, e.g. `supportsCharsetsInMultipartParts`

core/src/main/scala/sttp/client3/internal/package.scala
```diff
@@ -30,6 +34,18 @@ package object internal {
     transfer()
   }
 
+  private[client3] def withCharset(headers: Map[String, String], encoding: String): Map[String, String] = {
+    val contentTypeOpt = for {
+      charset <- Try(Charset.forName(encoding)).toOption
```
adamw, line 39: do we need this validation? if the user wants to specify an invalid charset, I'd let them do so. Or throw an exception, but that might be too restrictive

core/src/main/scalajvm/sttp/client3/HttpURLConnectionBackend.scala
```diff
@@ -150,7 +150,13 @@ class HttpURLConnectionBackend private (
     // inspired by: https://github.com/scalaj/scalaj-http/blob/master/src/main/scala/scalaj/http/Http.scala#L542
     val partsWithHeaders = mp.parts.map { p =>
       val contentDisposition = s"${HeaderNames.ContentDisposition}: ${p.contentDispositionHeaderValue}"
-      val otherHeaders = p.headers.map(h => s"${h.name}: ${h.value}")
+      val otherHeaders = p.body match {
+        case StringBody(_, encoding, _) =>
+          withCharset(p.headers.map(h => h.name -> h.value).toMap, encoding).map { case (name, value) =>
```
adamw, line 155: ah, in `Part` you already have a `Seq[Header]` so you can use that to create a `Headers` instance

core/src/main/scala/sttp/client3/internal/package.scala
```diff
@@ -30,6 +34,18 @@ package object internal {
...
+      contentType <- headers.get(HeaderNames.ContentType) if !contentType.contains("charset")
```
adamw, line 40: headers need to be compare using case-insensitive checks. There's a `Header.is` helper method, as well as a helper `Headers` wrapper, which even has a `.contentType` method on it. Maybe we can use it here.

core/src/main/scala/sttp/client3/internal/package.scala
```diff
@@ -30,6 +34,18 @@ package object internal {
     transfer()
   }
 
+  private[client3] def withCharset(headers: Map[String, String], encoding: String): Map[String, String] = {
```
adamw, line 37: maybe we could add an `addCharsetToContentType(charset: String): Part` method on `Part`, that could nicely encapsulate that logic

core/src/main/scala/sttp/client3/internal/package.scala
dkarwacki: but the user specifies the encoding, and we are changing it to the charset behind the scenes, so it's maybe better to keep content-type untouched if the encoding is incorrect?

adamw: I think silent errors might be very confusing

dkarwacki: ok, so i will get rid of it

finagle-backend/src/main/scala/sttp/client3/finagle/FinagleBackend.scala
```diff
@@ -128,11 +128,7 @@ class FinagleBackend(client: Option[Client] = None) extends SttpBackend[TFuture,
       case MultipartBody(_)       => throw new IllegalArgumentException("Nested multipart bodies are not supported")
     }
 
-    part.fileName match {
-      case Some(_) => FileElement(part.name, ByteArray(content.getBytes: _*), part.contentType, part.fileName)
-      case None    => SimpleElement(part.name, content)
-    }
-
+    FileElement(part.name, ByteArray(content.getBytes: _*), part.contentType, part.fileName)
```
adamw, line 131: I don' think we always want to return a file element?

core/src/test/scala/sttp/client3/testing/HttpTest.scala
```diff
@@ -409,6 +410,28 @@ trait HttpTest[F[_]]
         }
       }
 
+      if (supportsCustomMultipartEncoding) {
+        "send a multipart message containing body parts with explicitly set encodings" in {
+
+          val utf8Char = "\u00f6"
+          val replacementChar = "\ufffd"
```
adamw, line 417: I think I would add a comment where this replacement char originates from (and maybe render both in comment, if they have some proper visualisations in utf-8?)

finagle-backend/src/main/scala/sttp/client3/finagle/FinagleBackend.scala
dkarwacki: corrected

core/src/test/scala/sttp/client3/testing/HttpTest.scala
dkarwacki: corrected

PR #880, opened 2021-02-25
web sockets support in JS backends

kubinio123:


core/src/main/scala/sttp/client3/ConvertFromFuture.scala
```diff
@@ -0,0 +1,13 @@
+package sttp.client3
+
+import scala.concurrent.Future
+
+trait ConvertFromFuture[F[_]] {
```
adamw, line 5: this is used only in the JS impl? I'd move it to JS sources & to an internal package (that's not the public API)

effects/cats/src/test/scala/sttp/client3/impl/cats/CatsTestBase.scala
```diff
@@ -1,21 +1,20 @@
...
 import scala.concurrent.duration.DurationInt
 
 trait CatsTestBase {
+
   implicit def executionContext: ExecutionContext
 
   implicit lazy val monad: MonadError[IO] = new CatsMonadAsyncError[IO]
   implicit val contextShift: ContextShift[IO] = IO.contextShift(implicitly)
-  implicit lazy val timer: Timer[IO] = IO.timer(scala.concurrent.ExecutionContext.global)
+  implicit lazy val timer: Timer[IO] = IO.timer(implicitly)
```
adamw, line 17: probably you can leave out the `implicitly` :)

core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
```diff
@@ -240,63 +249,104 @@ abstract class AbstractFetchBackend[F[_], S <: Streams[S], P](
...
-      ): F[FetchResponse] = {
-        val bytes = replayableBody match {
-          case Left(byteArray) => byteArray
-          case Right(file)     => throw new IllegalArgumentException("Replayable file bodies are not supported")
-        }
-        new FetchResponse(bytes.toTypedArray.asInstanceOf[BodyInit], response.asInstanceOf[ResponseInit]).unit
-      }
+    ws.onopen = (_: Event) => queue.offer(WebSocketEvent.Open())
+    ws.onmessage = (event: MessageEvent) => queue.offer(toWebSocketEvent(event))
+    ws.onerror = (_: Event) => queue.offer(WebSocketEvent.Error(new WebSocketErrorException))
```
adamw, line 259: can't we extract some more information from the event about the error?

Would be best to just throw the original exception. The `WebSocketErrorException` is used only here, so I'd define it locally, or just use `RuntimeException`

core/src/main/scalajs/sttp/client3/JSSimpleQueue.scala
```diff
@@ -0,0 +1,27 @@
...
+  private val queue = new ConcurrentLinkedQueue[T]()
+
+  override def offer(t: T): Unit =
+    if (!queue.offer(t)) throw WebSocketBufferFull(queue.size)
+
+  override def poll: F[T] =
+    fromFuture {
+      JSTimeout[T] {
+        val t = queue.poll()
+        Either.cond(t != null, t, new WebSocketTimeoutException)
```
adamw, line 24: if we `poll` without a timeout, can we get a timeout?

core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
```diff
@@ -240,63 +249,104 @@ abstract class AbstractFetchBackend[F[_], S <: Streams[S], P](
...
+    fromFuture(isOpen).flatMap { _ =>
+      bodyFromResponseAs
+        .apply(request.response, ResponseMetadata(StatusCode.Ok, "", request.headers), Right(webSocket))
+        .map(Response.ok)
+    }
+  }
 
-      override protected def regularAsStream(response: FetchResponse): F[(streams.BinaryStream, () => F[Unit])] =
-        handleResponseAsStream(response)
+  def webSocketTimeout: FiniteDuration = 1.second
```
adamw, line 275: does this mean that if a message isn't received every second you'll get an error?

effects/cats/src/test/scala/sttp/client3/impl/cats/CatsTestBase.scala
kubinio123: compiler complains, I can only replace `implicitly` with `executionContext` from above

core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
kubinio123: There's no info about error in `Event` object, I will use `RuntimeException

core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
kubinio123: No it defines timeout for waiting for web socket to open, and timeout for each `ws.receive()` call

core/src/main/scalajs/sttp/client3/JSSimpleQueue.scala
```diff
@@ -0,0 +1,41 @@
...
+import sttp.client3.internal.ws.SimpleQueue
+
+import scala.concurrent.{Future, Promise}
+
+private[client3] class JSSimpleQueue[F[_], T](implicit fromFuture: ConvertFromFuture[F]) extends SimpleQueue[F, T] {
+
+  var state: Either[List[Promise[T]], List[T]] = Right(List())
+
+  def offer(t: T): Unit = state match {
+    case Left(promises) =>
```
adamw, line 13: maybe a bit better keeping everything at top-level:
```
case Left(p :: t) =>
case Left(Nil) =>
case Right(elems) =>
```

core/src/main/scalajs/sttp/client3/JSSimpleQueue.scala
```diff
@@ -0,0 +1,41 @@
+package sttp.client3
+
+import sttp.client3.internal.ConvertFromFuture
+import sttp.client3.internal.ws.SimpleQueue
+
+import scala.concurrent.{Future, Promise}
+
+private[client3] class JSSimpleQueue[F[_], T](implicit fromFuture: ConvertFromFuture[F]) extends SimpleQueue[F, T] {
+
+  var state: Either[List[Promise[T]], List[T]] = Right(List())
```
adamw, line 10: can be private

effects/monix/src/main/scala/sttp/client3/impl/monix/MonixWebSockets.scala
```diff
@@ -30,7 +31,10 @@ object MonixWebSockets {
       )
         .mapEval(ws.send(_))
         .completedL
-        .guarantee(ws.close())
+        .guaranteeCase {
```
adamw, line 34: why, isn't close() idempotent?

core/src/main/scalajs/sttp/client3/AbstractFetchBackend.scala
```diff
@@ -240,63 +245,107 @@ abstract class AbstractFetchBackend[F[_], S <: Streams[S], P](
...
 
-      override protected def regularIgnore(response: FetchResponse): F[Unit] =
-        transformPromise(response.arrayBuffer()).map(_ => ())
+    ws.onopen = (_: Event) => {
+      isOpen.success(())
+      queue.offer(WebSocketEvent.Open())
+    }
+    ws.onmessage = (event: MessageEvent) => queue.offer(toWebSocketEvent(event))
+    ws.onerror = (_: Event) => {
+      if (!isOpen.isCompleted) isOpen.failure(new ReadException(request, new WebSocketTimeoutException))
```
adamw, line 261: We don't really know if it's a timeout. Could be, could be not. I guess the best we can tell is that the WS was not opened for some reason.

Btw. there's really no way of finding out the error message? :)

core/src/test/scala/sttp/client3/testing/websocket/WebSocketStreamingTest.scala
```diff
@@ -62,20 +59,21 @@ trait WebSocketStreamingTest[F[_], S] extends ToFutureWrapper { outer: Suite wit
...
-          }
-        )
-        None
-      case WebSocketFrame.Text(payload, _, _) =>
-        val wholePayload = buffer.getAndSet("") + payload
-        received.add(wholePayload)
-        Some(WebSocketFrame.text(wholePayload + "-echo"))
-      case _ => throw new RuntimeException()
+    functionToPipe { d =>
+      println("running functionToPipe")
```
adamw, line 63: SOD

core/src/test/scala/sttp/client3/testing/websocket/WebSocketStreamingTest.scala
```diff
@@ -90,17 +88,12 @@ trait WebSocketStreamingTest[F[_], S] extends ToFutureWrapper { outer: Suite wit
...
     prepend(WebSocketFrame.text("1"))(functionToPipe {
       case WebSocketFrame.Text(payload, false, _) =>
-        buffer.accumulateAndGet(
-          payload,
-          new BinaryOperator[String] {
-            override def apply(t: String, u: String): String = t + u
-          }
-        )
+        val s = buffer.get()
+        buffer.set(s + payload)
```
adamw, line 92: while it doesn't matter here, in general there's a difference between the two operations - `buffer.accumulateAndGet` is atomic, while get+set is not. Or was it causing problems?

core/src/test/scala/sttp/client3/testing/websocket/WebSocketTest.scala
```diff
@@ -97,6 +114,7 @@ abstract class WebSocketTest[F[_]]
     implicit val signaler: Signaler = ThreadSignaler
     failAfter(Span(15, Seconds)) {
       basicRequest
+        .readTimeout(1.second)
```
adamw, line 117: do we need these timeouts now?

effects/monix/src/main/scala/sttp/client3/impl/monix/MonixWebSockets.scala
kubinio123: right, I will adjust that in `WebSocketImpl`

core/src/test/scala/sttp/client3/testing/websocket/WebSocketStreamingTest.scala
kubinio123: I know but it seems that `accumulateAndGet` is not convertible to JS:
`Referring to non-existent method scala.concurrent.impl.Promise$Link.accumulateAndGet`

adamw: Ah ok :)

PR #900, opened 2021-03-21
Update resilience to disclose that sttp does retry

jsoref:
Users shouldn't have to debug to discover that sttp does in fact do more than a 1:1 connection.

If your server happens to have an idle timeout (e.g. play at 75 seconds),
and thus the server side's connection closes prematurely,
sttp will merrily retry the connection, and if that fails, it'll merrily try again, and again, ....


adamw:
Hm it is possible that some backends have some additional behavior. Here you are using async-http-client, is that correct?

mchilderson:
Correct - I worked with @jsoref on the problem.

adamw:
Thanks - I'll reword a bit though, and move the changes to the proper docs, as you've modified the generated version. 

To configure AHC, you can provide the `org.asynchttpclient.maxRequestRetry` option (default 5), or provide custom configuration when creating the backend.

PR #913, opened 2021-03-29
CatsMonadAsyncError#async typo fix

nryanov:
Not sure, but it seems that there is a small typo


adamw:
Definitely - thanks!

PR #927, opened 2021-04-09
Exclude CE3 core dependency from `cats` module

catostrophe:
This is a tiny change that removes the dependency on `ce3-core` from sttp cats module where `ce3-kernel` is enough.


catostrophe:
Failed in an unrelated test. Please re-run jobs.

catostrophe:
@adamw please ru-run jobs.
I also suggest doing the same in tapir

adamw:
Thanks! Yeah, these tests sometimes fail randomly. One day I'll figure out why ;)

PR #936, opened 2021-04-15
add subprotocols option to httpbackend

barthorre:

Not sure how to properly test these changes.
Any suggestions to a more generic approach across multiple backends is welcome.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -114,7 +114,7 @@ val scalaTest = libraryDependencies ++= Seq("freespec", "funsuite", "flatspec", 
 val zioVersion = "1.0.6"
 val zioInteropRsVersion = "1.3.2"
 
-val sttpModelVersion = "1.4.1"
+val sttpModelVersion = "1.4.2"
 val sttpSharedVersion = "1.2.0"
 
 val logback = "ch.qos.logback" % "logback-classic" % "1.2.3"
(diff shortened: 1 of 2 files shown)


adamw:
Thanks!

I'm not too familiar with WS protocols, but isn't it enough to add a `Sec-WebSocket-Protocol` header(s) to the request? Maybe we could support it by adding the header to the request?

The `WebSocket` class also has an `upgradeHeaders` field, which should hold the chosen protocol.

barthorre:
The issue would be that HttpClient throws an IllegalArgument exception on adding that header.

adamw:
Maybe we could write some logic that in case of a WS request, if that header is present, remove it from the normally-handled payload, and instead use the HttpClient-provided methods?

barthorre:
That could work, considered it for a moment, be that felt kinda hacky, and would maybe be counter intuitive for people that work with httpclient directly.

adamw:
Well the subprotocols specification feels as if it should be part of the request description - that's not something global (which you would specify in a backend), but rather specific to a single request. The header looks like the simplest approach, but maybe there are better ones :)

I wouldn't be too concerned with the `HttpClient` client API, as it's only one of many backends. I think we should rather go with anything that fits the existing sttp API.

barthorre:
Is there a model where I could add some fixed header names we would search for and convert from header on the request to something added to the `WebSocket.Builder`?
As far as I can see `WebSocket.Builder` only supports adding:
- headers, already covered
- connection timeout, already covered
- subprotocols, would be derived from one specific header?

adamw:
Yes, the header names are here: https://github.com/softwaremill/sttp-model/blob/master/core/src/main/scala/sttp/model/HeaderNames.scala

barthorre:
Guess I can start by adding the one we need here:
https://github.com/softwaremill/sttp-model/pull/82

barthorre:
I removed all the `Sec-WebSocket` headers from the request since they are all illegal headers according to HttpClient.
Although the `WebSocket.Builder` provides no other way of passing that information along.

adamw:
Thank, released :)

barthorre:
Thanks for the quick actions on your side!

PR #937, opened 2021-04-16
Add zio-telemetry-opentelemetry-backend

duxet:

There is a backend for opentracing, but there's none for opentelemetry, so it's time to fix that 🙂

### DIFF
--- a/metrics/zio-telemetry-open-telemetry-backend/src/main/scala/sttp/client3/ziotelemetry/opentelemetry/ZioTelemetryOpenTelemetryBackend.scala
+++ b/metrics/zio-telemetry-open-telemetry-backend/src/main/scala/sttp/client3/ziotelemetry/opentelemetry/ZioTelemetryOpenTelemetryBackend.scala
@@ -1,0 +1,52 @@
+package sttp.client3.ziotelemetry.opentelemetry
+
+import io.opentelemetry.api.trace.SpanKind
+import io.opentelemetry.api.trace.propagation.W3CTraceContextPropagator
+import io.opentelemetry.context.propagation.{TextMapPropagator, TextMapSetter}
+import sttp.capabilities.Effect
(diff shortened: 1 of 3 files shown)


metrics/zio-telemetry-open-telemetry-backend/src/main/scala/sttp/client3/ziotelemetry/opentelemetry/ZioTelemetryOpenTelemetryBackend.scala
```diff
@@ -0,0 +1,52 @@
...
+import zio.telemetry.opentelemetry._
+
+import scala.collection.mutable
+
+class ZioTelemetryOpenTelemetryBackend[+P] private (
+    delegate: SttpBackend[RIO[Tracing, *], P],
+    tracer: ZioTelemetryOpenTelemetryTracer
+) extends DelegateSttpBackend[RIO[Tracing, *], P](delegate) {
+  def send[T, R >: P with Effect[RIO[Tracing, *]]](request: Request[T, R]): RIO[Tracing, Response[T]] = {
+    val carrier: mutable.Map[String, String] = mutable.Map().empty
```
adamw, line 20: is it safe to use an ordinary mutable map here? can't it be updated from multiple threads concurrently?

duxet: It's modified only in [W3CTraceContextPropagator](https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/trace/propagation/W3CTraceContextPropagator.java#L133,L147) `inject` method, which is called only once, so I think it should be safe.

PR #939, opened 2021-04-17
Cross-compile more Scala.js modules to Scala 3

nightscape:
- [X] Check if the project compiles by running `sbt compile` (some 2.11 modules don't compile on my machine, but I don't think my changes have to do anything with it)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -42,9 +42,13 @@ val commonJsSettings = commonSettings ++ Seq(
     if (isSnapshot.value) Seq.empty
     else
       Seq {
+        val mapSourcePrefix = if (ScalaArtifacts.isScala3(scalaVersion.value))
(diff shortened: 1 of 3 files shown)


nightscape:
@adamw this now includes support for Circe and is therefore feature-complete from my side.
I couldn't test it extensively, I just ran the tests for the `circeJS_3_0_0_RC2` and they were green.
Will CI catch eventual problems, or is there something I should check (except for the four items from the first comment)?

nightscape:
Not sure if that failing ZIO test for Scala 2.11 has something to do with the changes here, I would assume it doesn't.

adamw:
Thanks! Yeah those failing tests are due to timeouts, which occur sometimes and quite randomly.

PR #944, opened 2021-04-21
queue instead of bytes concatenation

kubinio123:
So it's faster now, ~ 80 to ~ 30 seconds on my computer. Also memory usage is normal, with byte concatenation it was up to 8 GB for the 50 MB file from the affected example.

### DIFF
--- a/async-http-client-backend/fs2/src/main/scala/sttp/client3/asynchttpclient/fs2/AsyncHttpClientFs2Backend.scala
+++ b/async-http-client-backend/fs2/src/main/scala/sttp/client3/asynchttpclient/fs2/AsyncHttpClientFs2Backend.scala
@@ -1,13 +1,11 @@
 package sttp.client3.asynchttpclient.fs2
 
-import java.io.File
-import java.nio.ByteBuffer
 import cats.effect.kernel._
 import cats.effect.std.{Dispatcher, Queue}
 import cats.implicits._
-import fs2.{Chunk, Pipe, Stream}
 import fs2.interop.reactivestreams._
 import fs2.io.file.Files
+import fs2.{Chunk, Pipe, Stream}
 import io.netty.buffer.{ByteBuf, Unpooled}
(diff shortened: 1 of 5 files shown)


adamw:
Is the speed of the cats+zio backend the same now as with the zio-native one?

kubinio123:
yes, both take ~30 seconds

async-http-client-backend/fs2/src/main/scala/sttp/client3/asynchttpclient/fs2/AsyncHttpClientFs2Backend.scala
```diff
@@ -49,8 +51,8 @@ class AsyncHttpClientFs2Backend[F[_]: Async] private (
       override def publisherToBytes(p: Publisher[ByteBuffer]): F[Array[Byte]] = {
         p.toStream[F]
           .compile
-          .fold(ByteBuffer.allocate(0))(concatByteBuffers)
-          .map(_.array())
+          .fold(new ConcurrentLinkedQueue[Array[Byte]]())(enqueueBytes)
```
adamw, line 54: do we need a `ConcurrentQueue` here? Wouldn't an ordinary `List`/`Vector`/`Queue` suffice? The fold is sequential, adding elements one-by-one, providing the previous value

kubinio123: right, I'll replace it with normal linked queue

adamw: or just a Scala immutable Queue? :)

kubinio123: yes

adamw:
Thanks! :)

Issue #958, opened 2021-05-05
Add option to proxy requests only to specific hosts


LazyBun:
I'm using `SttpBackendOptions` to setup a proxy, but I'd like to use this proxy only for specific hosts. Unfortunately, it seems that `SttpBackendOptions` only supports `nonProxyHosts`.

It seems that it should be relatively easy to implement a `nonProxyHosts` opposite - e.g. `hostsToProxy`.

LazyBun:
Closed by #959

PR #959, opened 2021-05-06
[#958] It is now possible to define for which hosts the requests should be proxied

LazyBun:

### DIFF
--- a/core/src/main/scala/sttp/client3/SttpBackendOptions.scala
+++ b/core/src/main/scala/sttp/client3/SttpBackendOptions.scala
@@ -67,7 +67,7 @@ object SttpBackendOptions {
       nonProxyHosts.exists(isWildCardMatch(host, _))
 
     private def doesNotMatchAnyHostToProxy(host: String) =
-      hostsToProxy != Nil && hostsToProxy.forall(!isWildCardMatch(host, _))
+      hostsToProxy != Nil && !hostsToProxy.exists(isWildCardMatch(host, _))
 
     def asJavaProxySelector: net.ProxySelector =
       new net.ProxySelector {
--- a/core/src/main/scala/sttp/client3/SttpBackendOptions.scala
+++ b/core/src/main/scala/sttp/client3/SttpBackendOptions.scala
@@ -36,7 +36,8 @@ object SttpBackendOptions {
       port: Int,
       proxyType: ProxyType,
       nonProxyHosts: List[String] = Nil,
-      auth: Option[ProxyAuth] = None
+      auth: Option[ProxyAuth] = None,
+      hostsToProxy: List[String] = Nil
   ) {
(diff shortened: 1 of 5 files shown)


core/src/main/scala/sttp/client3/SttpBackendOptions.scala
```diff
@@ -36,7 +36,8 @@ object SttpBackendOptions {
       port: Int,
       proxyType: ProxyType,
       nonProxyHosts: List[String] = Nil,
-      auth: Option[ProxyAuth] = None
+      auth: Option[ProxyAuth] = None,
+      hostsToProxy: List[String] = Nil
```
adamw, line 40: maybe to keep naming symmetry this should be called e.g. `useProxyHosts`?

LazyBun: I don't think that `useProxyHosts` reflects that _only_ those hosts will be proxied. 

How about `onlyProxyHosts`? It keeps to the symmetry and better (I think) shows what it does

core/src/main/scala/sttp/client3/SttpBackendOptions.scala
```diff
@@ -59,9 +60,14 @@ object SttpBackendOptions {
...
-    def ignoreProxy(host: String): Boolean = {
+    def ignoreProxy(host: String): Boolean =
+      matchesNonProxyHost(host) || doesNotMatchAnyHostToProxy(host)
+
+    private def matchesNonProxyHost(host: String) =
       nonProxyHosts.exists(isWildCardMatch(host, _))
-    }
+
+    private def doesNotMatchAnyHostToProxy(host: String) =
+      hostsToProxy != Nil && hostsToProxy.forall(!isWildCardMatch(host, _))
```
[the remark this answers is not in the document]
LazyBun: The `hostsToProxy != Nil` is necessary because otherwise we'd be ignoring proxy despite user not setting any `hostsToProxy`. The reason being that `forall` returns `true` by default for `Nil`, which makes sense of course, but in this case we don't want to check this condition at all if user did not set `hostsToProxy`.

Now that I've written it down, I think that wrapping `hostsToProxy` in `Option` could be best here, but that would be kinda different from `nonProxyHosts`... What do you think?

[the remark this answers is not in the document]
LazyBun: I've rewritten it as `hostsToProxy != Nil && !hostsToProxy.exists(isWildCardMatch(host, _))`, because I think it might be clearer that way.

core/src/main/scala/sttp/client3/SttpBackendOptions.scala
adamw: yeah this looks good. I'll do the rename quickly, thanks for the suggestion :)

PR #968, opened 2021-05-13
Basic Request converter to RFC 2616 format (HTTP 1.1)

rojekp:

### DIFF
--- a/core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
+++ b/core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
@@ -17,14 +17,34 @@ class ToRfc2616Converter[R <: RequestT[Identity, _, _]] {
   private def extractBody(r: R): String = {
     r.body match {
       case StringBody(text, _, _) => s"$text"
-      case ByteArrayBody(_, _)    => "" //TODO
-      case ByteBufferBody(_, _)   => "" //TODO
-      case InputStreamBody(_, _)  => "" //TODO
-      case StreamBody(_)          => "" //TODO
-      case MultipartBody(_)   => "" //TODO
+      case ByteArrayBody(_, _)    => "<PLACEHOLDER>"
+      case ByteBufferBody(_, _)   => "<PLACEHOLDER>"
+      case InputStreamBody(_, _)  => "<PLACEHOLDER>"
+      case StreamBody(_)          => "<PLACEHOLDER>"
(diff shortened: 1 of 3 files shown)


core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
```diff
@@ -0,0 +1,63 @@
...
+      case NoBody                 => ""
+    }
+  }
+
+  def handleMultipartBody(parts: Seq[Part[RequestBody[_]]]): String = {
+    parts
+      .map { p =>
+        p.body match {
+          case StringBody(s, _, _) =>
+            s"""--<PLACEHOLDER>
```
adamw, line 35: shouldn't we create a proper boundary, instead of using a fixed one? I think `HttpURLConnectionBackend` contains an implementation

core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
```diff
@@ -0,0 +1,63 @@
...
+    val headers = extractHeaders(request)
+    val resultWithHeaders = if (headers.isEmpty) result else result + s"\n$headers"
+    val body = extractBody(request)
+    if (body.isEmpty) resultWithHeaders else resultWithHeaders + s"\n\n$body"
+  }
+
+  private def extractBody(r: R): String = {
+    r.body match {
+      case StringBody(text, _, _) => s"$text"
+      case ByteArrayBody(_, _)    => "<PLACEHOLDER>"
```
adamw, line 20: so these are effectively not supported?

core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
rojekp: In `toCurl` version boundaries are not required. Here I can create random boundary.

core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
rojekp: I can't find any examples in documentation, so not. I also asked in issue about that cases - whether we need them.

core/src/main/scala/sttp/client3/internal/ToRfc2616Converter.scala
rojekp: Added :+1:

PR #998, opened 2021-05-30
Made thenRespondCyclic* functions thread-safe

ithinkicancode:


core/src/main/scala/sttp/client3/testing/SttpBackendStub.scala
```diff
@@ -118,8 +118,27 @@ class SttpBackendStub[F[_], +P](
       */
     def thenRespondCyclicResponses[T](responses: Response[T]*): SttpBackendStub[F, P] = {
       val iterator = Iterator.continually(responses).flatten
-      thenRespond(iterator.next)
+      thenRespond(iterator.next())
     }
+
+    /** Thread-safe version of `thenRespondCyclic`
+      */
+    def thenRespondCyclicAtomically[T](body1: T, bodies: T*): SttpBackendStub[F, P] = {
```
adamw, line 126: looks good :) maybe we can just improve the existing `thenRespondCyclic` functions? Once we have thread-safe versions, I don't think there's a reason to keep them around :)

ithinkicancode: Yup, will replace the existing ones instead.

ithinkicancode:
Getting some weird error when running the updated tests:

```console
[error] Referring to non-existent method java.util.concurrent.atomic.AtomicInteger.getAndUpdate(java.util.function.IntUnaryOperator)int
[error]   called from sttp.client3.testing.AtomicCyclicIterator.next()java.lang.Object
[error]   called from sttp.client3.testing.SttpBackendStub$WhenRequest.thenRespondCyclicResponses(sttp.client3.Response,scala.collection.immutable.Seq)sttp.client3.testing.SttpBackendStub
[error]   called from private sttp.client3.testing.SttpBackendStubTests.testFun$proxy18$1()org.scalatest.compatible.Assertion
[error]   called from private sttp.client3.testing.SttpBackendStubTests.$init$$$anonfun$25()java.lang.Object
[error]   called from constructor sttp.client3.testing.SttpBackendStubTests.<init>()void
[error]   called from static constructor sttp.client3.testing.SttpBackendStubTests.<stinit>()void
[error]   called from core module analyzer
```

I checked scala.js implementation. Seems both AtomicInteger and IntUnaryOperator are implemented in scala.js. Need to take a closer look.

core/src/main/scala/sttp/client3/testing/SttpBackendStub.scala
ithinkicancode: Updated. The old impl is now replaced by the atomic impl. Now I need to deal with failed tests. There are so many tests to run and I've been getting OOM towards the end. Need to figure that out too.

adamw:
You might need to increase sbt memory. But for this change, just running the tests from `core` should be enough, + there's CI :)

ithinkicancode:
It's ready now :)

ithinkicancode:
> You might need to increase sbt memory. But for this change, just running the tests from `core` should be enough, + there's CI :)

I'm using `-J-Xss4M -J-Xms4g -J-Xmx4g`. What would be your recommendation? :)

ithinkicancode:
@adamw I noticed the build failed. I'm seeing the same errors in js and native builds for those Java atomic methods. I can look into the errors if you don't have a better insight on why they happen. Thanks.

adamw:
Fixed: https://github.com/softwaremill/sttp/commit/3579395a2b2e825a439aa3967d80bcaefd06934d :)

adamw:
> I'm using -J-Xss4M -J-Xms4g -J-Xmx4g. What would be your recommendation? :)

I've got 8G specified, but that's just a magic number, I didn't do any measurments to check if it's enough. I usually don't run all tests locally anyway

PR #1001, opened 2021-06-01
Allow customizing log level for HttpErrors

pawelkaczor:
If you find this change valuable, I will continue working on this PR.

### DIFF
--- a/core/src/main/scala/sttp/client3/logging/Log.scala
+++ b/core/src/main/scala/sttp/client3/logging/Log.scala
@@ -34,7 +34,7 @@ class DefaultLog[F[_]](
     sensitiveHeaders: Set[String] = HeaderNames.SensitiveHeaders,
     beforeRequestSendLogLevel: LogLevel = LogLevel.Debug,
     responseLogLevel: StatusCode => LogLevel = DefaultLog.defaultResponseLogLevel,
-    responseExceptionLogLevel: Option[StatusCode] => LogLevel = DefaultLog.defaultResponseExceptionLogLevel
+    responseExceptionLogLevel: LogLevel = LogLevel.Error
 ) extends Log[F] {
 
   def beforeRequestSend(request: Request[_, _]): F[Unit] =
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sttp/client3/logging/Log.scala
```diff
@@ -61,13 +61,24 @@ class DefaultLog[F[_]](
       }
     )
 
-  override def requestException(request: Request[_, _], elapsed: Option[Duration], e: Exception): F[Unit] =
-    logger(responseExceptionLogLevel, s"Exception when sending request: ${request.showBasic}${took(elapsed)}", e)
+  override def requestException(request: Request[_, _], elapsed: Option[Duration], e: Exception): F[Unit] = {
+    val logLevel = e match {
+      case HttpError(_, statusCode) =>
+        responseExceptionLogLevel(Some(statusCode))
```
adamw, line 67: maybe here we should use simply `responseLogLevel`?

pawelkaczor: Yes, but what to do when  statusCode is not available?

adamw: Then we use `responseExceptionLogLevel`? as we are dealing with an exception which prevented the response from being obtained

pawelkaczor: Yes, actually this was my initial solution. But I was not sure if we can reuse responseLogLevel for handling HttpError (which is an Exception) while providing responseExceptionLogLevel. That perhaps could be confusing.

adamw: Yeah I was not sure either, but then `HttpError` is an "artificially" introduced exception to represent unwanted status codes, not something that comes from the backend, so maybe this will be less confusing than two log level configurations, both accepting status codes? :)

pawelkaczor: Not sure I follow your reasoning,  but I agree :)

pawelkaczor: I expected responseLogLevel function (that accepts StatusCode) to be used for handling HttpError. Later I was suprised it does not work as I thought. So yes, I agree with you @adamw .

pawelkaczor: I don't know why some tests failed.

adamw:
Thanks! Don't worry about the test failure, it happens ... sometimes, for a reason I wasn't able to find yet :)

PR #1004, opened 2021-06-04
Akka-Http ZIO backend (WIP)

domdorn:
Basic implementation for a zio-backend wrapping a Akka-Http Backend. Normal requests work, however I need help converting from `AkkaStreams with WebSockets` to `ZioStreams with WebSockets`. 

Any pointers highly appreciated!


adamw:
I think for streaming you can try using zio-reactivestream-interop - first converting the zio-stream to a `Publisher`, then using this to create an akka-stream. And the other way round. 

See https://github.com/softwaremill/sttp/blob/master/async-http-client-backend/zio/src/main/scala/sttp/client3/asynchttpclient/zio/AsyncHttpClientZioBackend.scala#L51 for another backend which also uses this interop library

domdorn:
closed this as we're migrating away from akka due to the licensing mess.

Issue #1005, opened 2021-06-08
Should the zio-telemetry backends absorb the extra dependency?


thiloplanz:
I have a question about the zio-telemetry backend.

As it stands, it wraps an existing backend that works on `Task` into an enhanced one that does telemetry but then no longer works with `Task` (it exposes the extra dependency on `Tracing`.

> The yielded effects are of type `RIO[OpenTracing, *]`

This means that you cannot use this backend as a drop-in replacement for all your code that needed a `SttpClient` (i.e. a `Task`-based backend) and everything has to be refactored to deal with the added dependency.

Would it be possible (or a good idea) to have the wrapped backend absorb the dependency (by taking it in at construction time), so that this gets hidden from users?

adamw:
This sounds reasonable :) That's how I would do this normally, but maybe the "ZIO-way" is different.

thiloplanz:
@duxet Thoughts on this? If there are no objections, I can prepare a PR.

PR #1010, opened 2021-06-12
Have ZioTelemetryOpenTelemetryBackend absorb the Tracing dependency

thiloplanz:
As it stands,  ZioTelemetryOpenTelemetryBackend wraps an existing backend that works on Task into an enhanced one that does telemetry but then no longer works with Task (it exposes the extra dependency on Tracing.

>  The yielded effects are of type RIO[OpenTracing, *]

This means that you cannot use this backend as a drop-in replacement for all your code that needed a SttpClient (i.e. a Task-based backend) and everything has to be refactored to deal with the added dependency.

This PR changes this to have the wrapper absorb the dependency (by taking it in at construction time), so that this gets hidden from users.

See #1005


duxet (approved):

duxet:
looks good, thanks for adding a test for it 👍

adamw:
Thanks! :)

Issue #1016, opened 2021-06-23
make WebSocketZio more compelling using asWebSocketStream


phderome:
I am proposing a minimal change but feel that `asWebSocketStream` reflects more useful usage than `asWebSocketAlways`

phderome:
Not all that important to me anymore.

PR #1017, opened 2021-06-23
#1016 replace ZIO websocket example to be in line with Fs2

phderome:
showing the same… interesting API, which is `asWebSocketStream`

If reviewer agrees to this, I'd like additionally to find a non-side-effecting way to print in both `fs2` and `zio` websocket examples as this is not "idiomatic" to use `println` as we do in the examples.


phderome:
not so important work in my opinion.

PR #1024, opened 2021-06-30
Fix component tag for opentracing backend

Krever:
I hope all of that works, change was super small so I did it without checking out the project (from github directly)
> Before submitting pull request:
> - [ ] Check if the project compiles by running `sbt compile`
> - [ ] Verify docs compilation by running `sbt compileDocs`
> - [ ] Check if tests pass by running `sbt test`
> - [ ] Format code by running `sbt scalafmt`


adamw:
Thanks :)

Issue #1031, opened 2021-07-09
java.lang.UnsupportedOperationException: Couldn't connect to the proxy server.


jurijkruske:
we are using last version of sttp with okHttpBackend 3.3.9 to call external service with proxy. Sometimes we are receiving strange error from  SttpBackendOptions, but we cannot see the real error from network layer. 

java.io.IOException: canceled due to java.lang.UnsupportedOperationException: Couldn't connect to the proxy server.
	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:530)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.base/java.lang.Thread.run(Unknown Source)
	Suppressed: java.lang.UnsupportedOperationException: Couldn't connect to the proxy server.
		at sttp.client3.SttpBackendOptions$Proxy$$anon$1.connectFailed(SttpBackendOptions.scala:86)
		at okhttp3.internal.connection.RealConnection.connectFailed$okhttp(RealConnection.kt:691)
		at okhttp3.internal.connection.RealConnection.trackFailure$okhttp(RealConnection.kt:730)
		at okhttp3.internal.connection.Exchange.trackFailure(Exchange.kt:167)
		at okhttp3.internal.connection.Exchange.readResponseHeaders(Exchange.kt:111)
		at okhttp3.internal.http.CallServerInterceptor.intercept(CallServerInterceptor.kt:79)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.connection.ConnectInterceptor.intercept(ConnectInterceptor.kt:34)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.cache.CacheInterceptor.intercept(CacheInterceptor.kt:95)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.http.BridgeInterceptor.intercept(BridgeInterceptor.kt:83)
		at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
		at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(RetryAndFollowUpInterceptor.kt:76)
[...]

PR #1034, opened 2021-07-12
extend exception message if connection via proxyServer couldn't be es…

jurijkruske:
…tablished

Issue https://github.com/softwaremill/sttp/issues/1031#issuecomment-877213332
We use sttp with okhttpbackaend to connect external service via proxy.
Sometimes SttpBackendOptions fired UnsupportedOperationException without any message from IOException.
This change could help to find the reason why the connection could not be established.


adamw:
Thanks!

PR #1042, opened 2021-07-16
Added separate instance for MonadError

matwojcik:
There was no cats instance for MonadError, and not every effect has Async instance, so I'd added separate instance for such effects.


adamw:
Thank you!

PR #1075, opened 2021-08-19
added response size summary metric

mrForest13:
I added a response size metric. Can be useful in some cases

### DIFF
--- a/metrics/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
+++ b/metrics/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
@@ -3,7 +3,7 @@ package sttp.client3.prometheus
 import java.util.concurrent.ConcurrentHashMap
 
 import sttp.client3.{FollowRedirectsBackend, Identity, Request, Response, SttpBackend}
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! There seem to be some compilation errors, though

mrForest13:
There is any reason why you import `import scala.collection.immutable.Seq` in `sttp.client3.Response`? Why not just use `Seq`? (by that i mean remove import)

> [error] /home/runner/work/sttp/sttp/metrics/prometheus-backend/src/test/scala/sttp/client3/prometheus/PrometheusBackendTest.scala:252:59: type mismatch;
[error]  found   : Seq[sttp.model.Header]
[error]  required: scala.collection.immutable.Seq[sttp.model.Header]
[error]     val response = Response("Ok", StatusCode.Ok, "Ok", Seq(Header.contentLength(10)))
[error]                                                           ^
[error] one error found

adamw:
Because in Scala 2.12 and earlier, `Seq` was an alias for `import scala.collection.Seq` (which includes mutable and immutable ones). In 2.13+, this changed to `import scala.collection.immutable.Seq`.

mrForest13:
it makes sense, thanks

adamw:
Thanks as well :)

PR #1095, opened 2021-09-04
Cross-compile json4s module to Scala 3

nryanov:
[json4s](https://mvnrepository.com/artifact/org.json4s/json4s-core_3/4.0.3) is already cross-compiled to scala3, so i think that json4s module could also support it

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -794,7 +794,7 @@ lazy val json4s = (projectMatrix in file("json/json4s"))
     ),
     scalaTest
   )
-  .jvmPlatform(scalaVersions = scala2)
+  .jvmPlatform(scalaVersions = scala2 ++ scala3)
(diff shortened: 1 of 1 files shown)


nryanov:
Probably, it isn't fully ready :(
Got `No Manifest available for sttp.client3.Json4sTests.Outer.` on `sbt json4s3 / test`.
Currently, tests in json4s for scala3 [are also disabled](https://github.com/json4s/json4s/blob/master/scala_3_test_filter.sbt), so maybe later it will be fixed

adamw:
@nryanov If you can reproduce this using json4s directly, maybe it's worth reporting in the repo? + creating an issue and linking it here

nryanov:
@adamw yes, i'll come back to this problem soon and try to investigate more deeply

PR #1102, opened 2021-09-10
Add documentation that explicitly defines Authorization header behavior during redirects and how to change it.

scottweaver:
Add documentation that explicitly defines Authorization header behavior during redirects and how to change it.


adamw:
Thank you!

PR #1113, opened 2021-09-21
Empty gzip fix

### LINKED ISSUE #1065: HttpClient fs2 backend: Fail to read empty body when Content-Encoding is set
The fs2 HttpClient (JDK11+) backend fails if a response has `Content-Encoding: gzip` set and a status code 204 (No Content) as the fs2 gunzip pipe expects the gzip header bytes even if the stream is entirely empty. A more graceful handling would help consuming an otherwise perfectly fine response.

The RFCs don't seem entirely clear whether this is legal, but it can happen when using http4s (see https://github.com/http4s/http4s/pull/5056) and has been "fixed" in other HTTP clients before (see https://github.com/request/request/issues/2177).

bartekzylinski:
Should close #1065


adamw:
Thanks!

Issue #1149, opened 2021-10-19
Allow turning off auto-decompression


adamw:
Currently data is automatically decompressed when the response's encoding is gzip/deflate/other supported scheme. We should add an option to turn this off

PR #1154, opened 2021-10-21
Turning off auto-decompression

ghost:


adamw:
@Pask423 could you merge master and verify if this keeps binary compatibility? Running `mimaReportBinaryIssues` should do this trick

Pask423:
closes #1149

async-http-client-backend/cats-ce2/src/test/scala/sttp/client3/asynchttpclient/cats/AsyncHttpClientCatsHttpTest.scala
```diff
@@ -17,4 +17,5 @@ class AsyncHttpClientCatsHttpTest extends HttpTest[IO] with CatsTestBase {
   }
 
   override def throwsExceptionOnUnsupportedEncoding = false
+  override def supportAutoDecompressionDisabling = false
```
adamw, line 20: so are there any backends that do support decompression disabling?

adamw:
I appreciate the work-around, though I think that as it's not a critical feature/fix, let's leave it for the next major version and implement it properly there :)

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -399,8 +405,8 @@ class SpecifyAuthScheme[U[_], T, -R](hn: String, rt: RequestT[U, T, R], digestTa
 }
 
 case class RequestOptions(
-    followRedirects: Boolean,
-    readTimeout: Duration,
-    maxRedirects: Int,
-    redirectToGet: Boolean
-)
+                           followRedirects: Boolean,
```
adamw, line 408: needs reformating :)

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -233,6 +233,12 @@ case class RequestT[U[_], T, -R](
 
   def tag(k: String): Option[Any] = tags.get(k)
 
+  private val disableAutoDecompressionKey = "disableAutoDecompression"
```
adamw, line 236: maybe let's add a comment that we're using tags to keep bin compat but this will be promoted to an option in the future?

core/src/test/scala/sttp/client3/testing/HttpTest.scala
```diff
@@ -53,6 +54,7 @@ trait HttpTest[F[_]]
   protected def throwsExceptionOnUnsupportedEncoding = true
   protected def supportsHostHeaderOverride = true
   protected def supportsCancellation = true
+  protected def supportAutoDecompressionDisabling = false
```
adamw, line 57: [minor] naming -> all other flags start with `supports`

core/src/test/scala/sttp/client3/testing/HttpTest.scala
```diff
@@ -53,6 +54,7 @@ trait HttpTest[F[_]]
   protected def throwsExceptionOnUnsupportedEncoding = true
   protected def supportsHostHeaderOverride = true
   protected def supportsCancellation = true
+  protected def supportsAutoDecompressionDisabling = false
```
adamw, line 57: sorry, one more comment. I think it's better to require by default that a backend passes all tests (supports all features), and only selectively mark the backends that don't support something. So I'd set this as `true` by default, and override with `false` if needed

adamw (approved):

PR #1196, opened 2021-12-01
Fix LoggingBackend when used with WebSocket

### LINKED ISSUE #1188: Logging with websockets error
I was playing around with sttp for the first time and was trying to enable logging.  In the below code, `logResponseBody = true` gives a runtime exception, `logResponseBody = false` works as expected.  

It may not make sense to try to log a websocket request this way but probably the code should handle this more clearly.  It's a low priority problem but tripped me up as a new user so figured I'd report it.

If you want to try locally, the `ws://localhost:10000` is this `docker run --detach -p 10000:8080 -e SEND_SERVER_HOSTNAME=false jmalloc/echo-server`

```

import cats.effect.{ExitCode, IO, IOApp}
import fs2._
import sttp.capabilities.fs2.Fs2Streams
import sttp.client3._
import sttp.client3.asynchttpclient.fs2.AsyncHttpClientFs2Backend
import sttp.client3.logging.slf4j.Slf4jLoggingBackend
import sttp.ws.WebSocketFrame

object WebSocketStreamFs2 extends IOApp {

  def webSocketFramePipe: Pipe[IO, WebSocketFrame.Data[_], WebSocketFrame] = { input =>
    Stream.emit(WebSocketFrame.text("1")) ++ input.flatMap {
      case WebSocketFrame.Text("10", _, _) =>
        println("Received 10 messages, sending close frame")
        Stream.emit(WebSocketFrame.close)
      case WebSocketFrame.Text(n, _, _) if n.length > 0 =>
        println(s"Received $n messages, replying with $n+1")
        Stream.emit(WebSocketFrame.text((n.toInt + 1).toString))
      case _ => Stream.empty // ignoring
    }
  }

  override def run(args: List[String]): IO[ExitCode] = {
    AsyncHttpClientFs2Backend
      .resource[IO]()
      .use { backend =>

        val loggingBackend = Slf4jLoggingBackend(backend, logResponseBody = true)
        basicRequest
          .response(asWebSocketStream(Fs2Streams[IO])(webSocketFramePipe))
          .get(uri"ws://localhost:10000")
          .send(loggingBackend)
          .void
      }.as(ExitCode.Success)
  }
}
```

micossow:

fixes #1188

### DIFF
--- a/core/src/main/scala/sttp/client3/logging/LoggingBackend.scala
+++ b/core/src/main/scala/sttp/client3/logging/LoggingBackend.scala
@@ -84,15 +84,21 @@ class LoggingWithResponseBodyBackend[F[_], S](
   override def send[T, R >: S with Effect[F]](request: Request[T, R]): F[Response[T]] = {
     log.beforeRequestSend(request).flatMap { _ =>
       val start = if (includeTiming) Some(now()) else None
-      (for {
-        response <- request.response(asBothOption(request.response, asStringAlways)).send(delegate)
-        _ <- log.response(request, response, response.body._2, elapsed(start))
-      } yield response.copy(body = response.body._1))
-        .handleError { case e: Exception =>
-          log
-            .requestException(request, elapsed(start), e)
-            .flatMap(_ => responseMonad.error(e))
-        }
+      if (request.isWebSocket) {
+        for {
+          response <- request.send(delegate)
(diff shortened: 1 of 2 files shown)


adamw:
Maybe we could add a test to `WebSocketTest`?

PR #1197, opened 2021-12-02
ToCurlConverter - hide sensitive headers

### LINKED ISSUE #1177: [BUG]: LoggingBackend with curl request does not redact sensitive headers
sttp.client3.logging.DefaultLog
```scala
  def beforeRequestSend(request: Request[_, _]): F[Unit] =
    logger(
      beforeRequestSendLogLevel, {
        s"Sending request: ${if (beforeCurlInsteadOfShow) request.toCurl
        else request.show(includeBody = logRequestBody, logRequestHeaders, sensitiveHeaders)}"
      }
    )
```

micossow:

fixes #1177

### DIFF
--- a/core/src/main/scala/sttp/client3/RequestT.scala
+++ b/core/src/main/scala/sttp/client3/RequestT.scala
@@ -292,6 +292,8 @@ case class RequestT[U[_], T, -R](
   ): F[Response[T]] = backend.send(asRequest.asInstanceOf[Request[T, P with Effect[F]]]) // as witnessed by pEffectFIsR
 
   def toCurl(implicit isIdInRequest: IsIdInRequest[U]): String = ToCurlConverter.requestToCurl(asRequest)
+  def toCurlWithSensitiveHeaders(sensitiveHeaders: Set[String])(implicit isIdInRequest: IsIdInRequest[U]): String =
+    ToCurlConverter.requestToCurlWithSensitiveHeaders(sensitiveHeaders)(asRequest)
 
   def toRfc2616Format(implicit isIdInRequest: IsIdInRequest[U]): String = ToRfc2616Converter.requestToRfc2616(asRequest)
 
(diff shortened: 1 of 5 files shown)


adamw:
Maybe we could add a test? :) There is already a `ToCurlConverterTest`

adamw:
Sorry, two more comments :)

1. maybe we could introduce a `toCurl` variant which would accept the `sensitiveHeaders` parameter. This could be passed from the log, but also users would have the option to log auth tokens if they wish to do so. One caveat here is that this needs to be done in a backwards-compatible way (try running `mimaReportBinaryIssues` in `core` and `core2_12`)
2. `ToRfc2616Converter` has the same problem

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -292,6 +292,8 @@ case class RequestT[U[_], T, -R](
   ): F[Response[T]] = backend.send(asRequest.asInstanceOf[Request[T, P with Effect[F]]]) // as witnessed by pEffectFIsR
 
   def toCurl(implicit isIdInRequest: IsIdInRequest[U]): String = ToCurlConverter.requestToCurl(asRequest)
+  def toCurlWithSensitiveHeaders(sensitiveHeaders: Set[String])(implicit isIdInRequest: IsIdInRequest[U]): String =
```
adamw, line 295: can we overload `toCurl`?

core/src/main/scala/sttp/client3/internal/ToCurlConverter.scala
```diff
@@ -6,10 +6,20 @@ import sttp.client3.NoBody
 import sttp.model.MediaType
 
 class ToCurlConverter[R <: RequestT[Identity, _, _]] {
-  def apply(request: R): String = {
-    val params = List(extractMethod(_), extractUrl(_), extractHeaders(_), extractBody(_), extractOptions(_))
+
+  def apply(request: R): String = create(request, sensitiveHeaders = None)
```
adamw, line 10: I'd understand `sensitiveHeaders = None` as "no sensitive headers" while it means "default sensitive headers". Maybe we can just a default value of `sttp.model.HeaderNames.SensitiveHeaders` instead of an `Option[Set[String]]`?

micossow: done

adamw (approved):

PR #1248, opened 2022-01-09
Add support for jsoniter Json

borissmidt:
- [x ] Check if the project compiles by running `sbt compile`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1011,7 +1011,8 @@ lazy val docs: ProjectMatrix = (projectMatrix in file("generated-docs")) // impo
     name := "docs",
     libraryDependencies ++= Seq(
       "org.json4s" %% "json4s-native" % json4sVersion,
-      "io.circe" %% "circe-generic" % "0.14.1",
+      "io.circe" %% "circe-generic" % circeVersion(None),
+      "com.github.plokhotnyuk.jsoniter-scala" %% "jsoniter-scala-macros" % jsoniterVersion,
       "commons-io" % "commons-io" % "2.11.0",
       "io.github.resilience4j" % "resilience4j-circuitbreaker" % resilience4jVersion,
       "io.github.resilience4j" % "resilience4j-ratelimiter" % resilience4jVersion,
--- a/build.sbt
+++ b/build.sbt
@@ -783,7 +783,7 @@ lazy val jsoniter = (projectMatrix in file("json/jsoniter"))
     scalaVersions = scala2,
     settings = commonJvmSettings
   )
-  .jsPlatform(scalaVersions = List(scala2_12, scala2_13) ++ scala3, settings = commonJsSettings)
+  .jsPlatform(scalaVersions = List(scala2_12, scala2_13), settings = commonJsSettings)
   .dependsOn(core, jsonCommon)
 
 lazy val zioJson = (projectMatrix in file("json/zio-json"))
--- a/build.sbt
+++ b/build.sbt
@@ -1030,6 +1030,7 @@ lazy val docs: ProjectMatrix = (projectMatrix in file("generated-docs")) // impo
     circe,
     sprayJson,
     zioJson,
+    jsoniter,
     asyncHttpClientZioBackend,
     // asyncHttpClientMonixBackend, // monix backends are commented out because they depend on cats-effect2
     asyncHttpClientFs2Backend,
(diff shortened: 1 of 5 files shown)


borissmidt:
Offtopic; I also updated the openApi generator to use sttp3. 
https://github.com/OpenAPITools/openapi-generator/pull/11260

json/jsoniter/src/main/scala/sttp/client3/jsoniter/SttpJsoniterJsonApi.scala
```diff
@@ -0,0 +1,54 @@
...
+  def asJsonEither[E: JsonValueCodec: IsOption, B: JsonValueCodec: IsOption]
+      : ResponseAs[Either[ResponseException[E, Exception], B], Any] = {
+    asJson[B].mapLeft {
+      case de @ DeserializationException(_, _) => de
+      case HttpError(e, code) => deserializeJson[E].apply(e).fold(DeserializationException(e, _), HttpError(_, code))
+    }.showAsJsonEither
+  }
+
+  def deserializeJson[B: JsonValueCodec: IsOption]: String => Either[Exception, B] = { s: String =>
+    try {
```
Pask423, line 47: maybe here you could use `Try` and`.toEither`  with`left.map` instead of try {} catch

build.sbt
```diff
@@ -767,6 +768,23 @@ lazy val circe = (projectMatrix in file("json/circe"))
   .jsPlatform(scalaVersions = List(scala2_12, scala2_13) ++ scala3, settings = commonJsSettings)
   .dependsOn(core, jsonCommon)
 
+lazy val jsoniter = (projectMatrix in file("json/jsoniter"))
```
Pask423, line 771: You should add jsoniter to `allAggregates` variable and probably also to `docs` as dependency

Pask423 (changes requested):

Pask423:
Could you merged newest master it should solve problem with js tests.
It would be also great if you could write docs for  `jsoniter` similar to one written for `zio-json`.
Please check if all files are correctly formatted with `fmt`.

borissmidt:
I see one big problem with jsoniter and sttp and that is that sttp uses either and option a lot.
The it is possible in jsonIter to handle the option with the implicit def.
But the Either isn't really that safe so the user has to manually define an implicit val for every variant of either he wants to use.

So i think jsoniter needs to be improved for our usecase.

Pask423:
That can be quite problematic from user perspective. I am not that familiar with jsoniter to propose any workaround for this.
BTW I still see some merge conflicts could you resolve them ?

borissmidt:
I could only make a workaround for 'option[t]'.
So i think changes have to be made to jsoniter before we can support it in this library.

For Either i couldn't really make a good workaround.

So we might better close the pr and maybe we can best keep the github issue open and comment these issues on there for future reference?

adamw:
Hm I don't think we need any special `Option`/`Either` support. The options/eithers that appear in signatures are used to represent cases where either JSON parsing failed, or we received a non-200 response. So we don't typically need json codecs for these types, only for whatever is the "root" type.

borissmidt:
I've removed the ambiguous Either tests and now i expect the use of `asJsonEither` so it is clear that you either have an error of type Left or a body of type Right.

json/jsoniter/src/main/scala/sttp/client3/jsoniter/SttpJsoniterJsonApi.scala
```diff
@@ -60,12 +50,13 @@ trait SttpJsoniterJsonApi {
       readFromString[B](JsonInput.sanitize[B].apply(s))
     }.toEither.left.map {
       case de: JsonReaderException => DeserializationException[JsonReaderException](s, de)
-      case e: Exception            => e
+      case NonFatal(e) => new RuntimeException(e)
```
Pask423, line 53: I think we should go back to previous version of Exception handling

Pask423:
Great just resolve conflicts and I think we can merge it

Pask423:
There is an error with downloading dependency
image

borissmidt:
scala 3 macro support still needs to be added. I'll remove scala3 from the possible builds.

Pask423 (approved):

PR #1249, opened 2022-01-10
Fix for out of memory execption in js tests

Pask423:


adamw:
🥳 would be great if this would actually solve the memory problems

build.sbt
```diff
@@ -43,6 +43,9 @@ val commonJvmSettings = commonSettings ++ Seq(
 )
 
 val commonJsSettings = commonSettings ++ Seq(
+  scalaJSLinkerConfig ~= {
+    _.withBatchMode(true).withParallel(false)
```
adamw, line 47: maybe add (directly to master) a comment explaining why we need this and how does this fix the issue?

Pask423:
We will see as for now it works as expected I tired to run CI several times and everything seemed to be ok.

build.sbt
Pask423: I will add it

Issue #1272, opened 2022-01-24
Support for disabling logging of a specific HTTP request in LoggingBackend


Kobenko:
Want to have something like this 
`req
      .response(asJson[Bar])
      .withLogSettings(
        logRequestBody = false,
        logResponseBody = false,
        logResponseHeaders = false,
        logRequestHeaders = true
      )
      .send(backend)`

Kobenko:
https://github.com/softwaremill/sttp/pull/1279

PR #1279, opened 2022-01-26
issue-1272: support for disabling logging of a specific HTTP request

Kobenko:
https://github.com/softwaremill/sttp/issues/1272


Pask423:
Thanks for implementing this feature.
Unfortunately there are problems with binary compatibility - look at failed mima job - you can try to make it in binary compatibility.
I am not sure if it is possible in reasonable fashion so it will probably have to wait before being merged.

core/src/main/scala/sttp/client3/logging/Log.scala
```diff
@@ -38,29 +38,50 @@ class DefaultLog[F[_]](
...
-  def beforeRequestSend(request: Request[_, _]): F[Unit] =
-    logger(
-      beforeRequestSendLogLevel, {
-        s"Sending request: ${if (beforeCurlInsteadOfShow) request.toCurl
-        else request.show(includeBody = logRequestBody, logRequestHeaders, sensitiveHeaders)}"
-      }
-    )
+  def beforeRequestSend(request: Request[_, _]): F[Unit] = request.options.loggingOptions match {
+    case Some(options) =>
+      logger(
```
Pask423, line 43: Maybe it is possible to move both cases to a method with some input parameters

core/src/main/scala/sttp/client3/logging/Log.scala
```diff
@@ -38,29 +38,50 @@ class DefaultLog[F[_]](
...
-      responseLogLevel(response.code), {
-        val responseAsString =
-          response
-            .copy(body = responseBody.getOrElse(""))
-            .show(responseBody.isDefined, logResponseHeaders, sensitiveHeaders)
-        s"Request: ${request.showBasic}${took(elapsed)}, response: $responseAsString"
-      }
-    )
+  ): F[Unit] = request.options.loggingOptions match {
+    case Some(options) =>
```
Pask423, line 64: same here

Pask423 (changes requested):

adamw:
@Kobenko @Pask423 while we cannot modify `RequestOption`s in a backwards-compatible way (it should have been a trait, but well, too late ;) ), there is the `.tag` method, which allows attaching arbitrary meta-data to a request. The signature is type-unsafe `def tag(k: String, v: Any): RequestT[U, T, R]` (again, type-safe attributes as in tapir would be better), but we could make this nicer via an extension methods added e.g. by an import from the logging package or companion object

Kobenko:
thanks for the help

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -233,6 +233,26 @@ case class RequestT[U[_], T, -R](
 
   def tag(k: String): Option[Any] = tags.get(k)
 
+  private val loggingOptionsTagKey = "loggingOptions"
+
+  /** Can only be applied when you use `LoggingBackend` */
```
adamw, line 238: that's not really true, it can be applied always, but it only has effect when using the logging backend ;)

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -233,6 +233,26 @@ case class RequestT[U[_], T, -R](
...
+    val loggingOptions = LoggingOptions(
+      logRequestBody = logRequestBody,
+      logResponseBody = logResponseBody,
+      logRequestHeaders = logRequestHeaders,
+      logResponseHeaders = logResponseHeaders
+    )
+    this.tag(loggingOptionsTagKey, loggingOptions)
+  }
+
+  def loggingOptions: Option[LoggingOptions] = tag(loggingOptionsTagKey).map(_.asInstanceOf[LoggingOptions])
```
adamw, line 254: we can cast without calling `.map` here, since we are casting anyway

core/src/main/scala/sttp/client3/logging/Log.scala
```diff
@@ -39,30 +39,59 @@ class DefaultLog[F[_]](
 ) extends Log[F] {
 
   def beforeRequestSend(request: Request[_, _]): F[Unit] =
+    request.loggingOptions match {
+      case Some(options) => before(request, options.logRequestBody, options.logRequestHeaders)
```
adamw, line 43: hm question is, should `options.logRequestBody` (and others) always override what's in the `DefaultLog` options? Maybe the fields in `LoggingOptions` should be optional, where a `None` means "keep default"

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -233,6 +233,26 @@ case class RequestT[U[_], T, -R](
 
   def tag(k: String): Option[Any] = tags.get(k)
 
+  private val loggingOptionsTagKey = "loggingOptions"
+
+  /** Can only be applied when you use `LoggingBackend` */
+  def withLogSettings(
```
adamw, line 239: I think I'd call this simply `.logSettings` as other methods don't have the `with` prefix

core/src/main/scala/sttp/client3/logging/Log.scala
Pask423: hmm, why not I options are optional themselves IMO there is no need for adding one more layer of optionality here.
Additionally loggingOptions have to be explicitly added by user so no surprises here

adamw: the idea would be to be able to override only a single setting (e.g. `logRequestHeaders`), while leaving others intact (as defined in the wrapper backend)

Pask423: you are right , will implement

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -233,6 +233,26 @@ case class RequestT[U[_], T, -R](
 
   def tag(k: String): Option[Any] = tags.get(k)
 
+  private val loggingOptionsTagKey = "loggingOptions"
+
+  /** Will only has effect when using the `LoggingBackend` */
```
adamw, line 238: typo: has -> have

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -408,3 +428,10 @@ case class RequestOptions(
     maxRedirects: Int,
     redirectToGet: Boolean
 )
+
+case class LoggingOptions(
```
adamw, line 432: maybe this should go into the `sttp.client3.logging` package?

core/src/main/scala/sttp/client3/logging/Log.scala
```diff
@@ -39,30 +39,64 @@ class DefaultLog[F[_]](
...
+        responseBody,
+        options.logResponseBody.getOrElse(responseBody.isDefined),
+        options.logResponseHeaders.getOrElse(logResponseHeaders),
+        elapsed
+      )
+    case None =>
+      toResponse(request.showBasic, response, responseBody, responseBody.isDefined, logResponseHeaders, elapsed)
+  }
+
+  private def toResponse(
```
adamw, line 82: [minor] `toResponse` is misleading - we aren't converting anything to a response. Maybe `handleResponse` or simply `response` with a different set of parameters :)

adamw (approved):

Pask423 (approved):

PR #1301, opened 2022-02-04
Expose Responses in PrometheusBackend where available

### LINKED ISSUE #1298: Expose Response in PrometheusBackend customization
Currently the signature of the `PrometheusBackend.apply` is
```
  def apply[F[_], P](
      delegate: SttpBackend[F, P],
      requestToHistogramNameMapper: Request[_, _] => Option[HistogramCollectorConfig] = (_: Request[_, _]) =>
        Some(HistogramCollectorConfig(DefaultHistogramName)),
      requestToInProgressGaugeNameMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
        Some(CollectorConfig(DefaultRequestsInProgressGaugeName)),
      requestToSuccessCounterMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
        Some(CollectorConfig(DefaultSuccessCounterName)),
      requestToErrorCounterMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
        Some(CollectorConfig(DefaultErrorCounterName)),
      requestToFailureCounterMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
        Some(CollectorConfig(DefaultFailureCounterName)),
      requestToSizeSummaryMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
        Some(CollectorConfig(DefaultRequestSizeName)),
      responseToSizeSummaryMapper: Response[_] => Option[CollectorConfig] = (_: Response[_]) =>
        Some(CollectorConfig(DefaultResponseSizeName)),
      collectorRegistry: CollectorRegistry = CollectorRegistry.defaultRegistry
  ): SttpBackend[F, P
  ```
  
And so the customization allows you to take data only from requests. It would be useful to have `Response` when available to extract also information from it. 
  
The exact use case I have is to put the response code as a Prometheus label. 

The change itself is rather simple and I could give it a shot but I'm not familiar with the exact compatibility guarantees of sttp. Can we just modify the signatures or do we need to create a separate constructor to keep the back-compat?

Krever:
Fixes #1298 
I hoped to keep backaward compat by introducing alternative `apply` + convert using `response.request` but its not a viable strategy - `request` is `RequestMetadata` not `Request[_,_]`. So probably we have to break the compact (or have two separate listeners, or very weird listener signatures).

### DIFF
--- a/metrics/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
+++ b/metrics/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
@@ -1,7 +1,6 @@
 package sttp.client3.prometheus
 
 import java.util.concurrent.ConcurrentHashMap
-
 import sttp.client3.{FollowRedirectsBackend, Identity, Request, Response, SttpBackend}
 import io.prometheus.client.{CollectorRegistry, Counter, Gauge, Histogram, Summary}
 import sttp.client3.listener.{ListenerBackend, RequestListener}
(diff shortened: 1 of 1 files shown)


Pask423:
Sorry that it took me so long.
I think that we should not drop requests supports from `apply` let's wait for @adamw review before merging it.

adamw:
Looks good :)

I'd just release the fixed zio-json artifacts in 3.4.2, and then this as 3.5.0, wdyt @Pask423 ?

Pask423:
Looks ok to me

PR #1310, opened 2022-02-07
updated reactive-streams to latest (2 years old) version

th-e:


th-e:
Just to avoid dependency conflict. Currently I exclude explicitly the old jar, because I get 1.0.3 from other parts of the application

Pask423:
Sorry that it took me so long.
Just to be clear `flow-adapters` was moved to core in `1.0.3` -  https://github.com/reactive-streams/reactive-streams-jvm/pull/430/files.
Previously we had implicit dependency to core via `flow-adapters` now we have it added `explicitly`

Pask423:
Thanks!

PR #1318, opened 2022-02-12
Split zio-json module for zio 1.x and 2.x

ex0ns:
As described in the zio-json releases:
```
For some time, releases targetting ZIO 1 will be using the version 0.2.x, whereas releases targetting ZIO 2 will be using the version 0.3.x
```

I think we should publish two targets:
1. `com.softwaremill.sttp.client3:zio-json_2.13`  which depends on `zio-json:0.3.X` for ZIO2
2. `com.softwaremill.sttp.client3:zio1-json_2.13`  which depends on `zio-json:0.2.X` for ZIO

The helper to get the zio-json version was not used anymore.
I think that this was not spotted because the matrix release is only for zio-2 at the moment

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -118,10 +118,6 @@ val circeVersion: Option[(Long, Long)] => String = {
 
 val jsoniterVersion = "2.13.2"
 
-val zioJsonVersion: Option[(Long, Long)] => String = {
-  case Some((3, _)) => "0.2.0-M3"
-  case _            => "0.1.5"
-}
 val playJsonVersion: Option[(Long, Long)] => String = {
   case Some((2, 11)) => "2.7.4"
   case _             => "2.9.2"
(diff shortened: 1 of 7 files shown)


adamw:
Thanks! Are there any code differences in the zio-json <-> sttp integration? If so, I think we're missing the actual code in `zio1-json`. If not, I guess it's easiest to use the `zioVersion` method to use the appropriate dependency.

ex0ns:
Yes you are right, I though I could use the same source for both versions, but we actually need to change the integration in between, I missed that when changing the build file.
I'll try to take care of those modification but I guess they will probably be what you have done https://github.com/softwaremill/sttp/commit/bd8ad49912e636f003bf7b9cecc7279d8b8ffdf9

Due to the amount of code of this package, I propose to duplicate the sources and the tests and maintain them in parallel for both version of ZIO, would that work for you ?

adamw:
Sure, duplicating both tests & sources sounds good :) thanks

adamw:
One last thing I think it would be good to mention this in the docs as well

ex0ns:
I don't understand why is `sbt -v testJS` failing.
I can't run them locally as I don't have the `google-chrome` binary.

adamw:
@ex0ns don't worry about the JS failures, unfortunately they happen ... sometimes, and we're not quite sure why. Sth with using Chrome to test the JS backend

PR #1327, opened 2022-02-18
Update future.md

VladKopanievExt:
There is no armeria artefact with "-future" postfix in its name, it looks like documentation is trying to refer to "armeria-backend" artefact instead


adamw:
Thanks!

PR #1332, opened 2022-02-22
Adding CE2 information to js fetch backends doc

TobiasRoland:
I had a bit of trouble identifying what the artifact name would be for CE2, so I figured this might help others in the future so they don't need to browse `https://repo1.maven.org/maven2/com/softwaremill/sttp/client3/` to find the dependency


adamw:
Thanks!

adamw:
Also I think `Any effect implementing the cats-effect Concurrent typeclass can be used.` is no longer valid - in CE3 it's `Async` if I remember correctly?

PR #1337, opened 2022-02-27
Added cats stub to docs

### LINKED ISSUE #1330: [DOC] "Creating a stub backend" docs could mention explicitly cats
https://sttp.softwaremill.com/en/latest/testing.html#creating-a-stub-backend contains paragraph:
```
An empty backend stub can be created using the following ways:

* by calling .stub on the “real” base backend’s companion object, e.g. AsyncHttpClientZioBackend.stub or HttpClientMonixBackend.stub

* by using one of the factory methods SttpBackendStub.synchronous or SttpBackendStub.asynchronousFuture, which return stubs which use the Identity or standard Scala’s Future effects without streaming support

* by explicitly specifying the effect and supported capabilities, e.g. SttpBackendStub[Task, MonixStreams with WebSockets](TaskMonad)

* by specifying a fallback/delegate backend, see below
```
It mentions zio, futures, monix, but not cats.  It would be nice to add point about IO with sample how to create it -> ` SttpBackendStub[IO, WebSockets](implicitly[MonadAsyncError[IO]])`

Pask423:
closes #1330

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -10,7 +10,9 @@ An empty backend stub can be created using the following ways:
 
 * by calling `.stub` on the "real" base backend's companion object, e.g. `AsyncHttpClientZioBackend.stub` or `HttpClientMonixBackend.stub`
 * by using one of the factory methods `SttpBackendStub.synchronous` or `SttpBackendStub.asynchronousFuture`, which return stubs which use the `Identity` or standard Scala's `Future` effects without streaming support
-* by explicitly specifying the effect and supported capabilities, e.g. `SttpBackendStub[Task, MonixStreams with WebSockets](TaskMonad)`
+* by explicitly specifying the effect and supported capabilities:
+  * for Monix `SttpBackendStub[Task, MonixStreams with WebSockets](TaskMonad)`
+  * for cats `SttpBackendStub[IO, WebSockets](implicitly[MonadAsyncError[IO]])`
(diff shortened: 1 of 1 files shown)


docs/testing.md
```diff
@@ -10,7 +10,9 @@ An empty backend stub can be created using the following ways:
 
 * by calling `.stub` on the "real" base backend's companion object, e.g. `AsyncHttpClientZioBackend.stub` or `HttpClientMonixBackend.stub`
 * by using one of the factory methods `SttpBackendStub.synchronous` or `SttpBackendStub.asynchronousFuture`, which return stubs which use the `Identity` or standard Scala's `Future` effects without streaming support
-* by explicitly specifying the effect and supported capabilities, e.g. `SttpBackendStub[Task, MonixStreams with WebSockets](TaskMonad)`
+* by explicitly specifying the effect and supported capabilities:
+  * for Monix `SttpBackendStub[Task, MonixStreams with WebSockets](TaskMonad)`
+  * for cats `SttpBackendStub[IO, WebSockets](implicitly[MonadAsyncError[IO]])`
```
adamw, line 15: maybe we should add ZIO as well?

adamw: The problem with "real" backends is that there's many of them for each implementation, so we won't be able to enumerate them all. I think people will get the idea that they have to call `.stub` :)

adamw (approved):

PR #1348, opened 2022-03-04
use only relative bulk get method

### LINKED ISSUE #1347: Rich byte buffer shouldn't return whole underlying array
`RichByteBuffer` returns the whole underlying array when it exists. That array can contain extra bytes that are not related to the current payload.
We should get only data from the current position in the byte buffer until the limit.

```scala
  implicit class RichByteBuffer(byteBuffer: ByteBuffer) {
    def safeRead(): Array[Byte] = {
      if (byteBuffer.hasArray) {
        byteBuffer.array()
      } else {
        val array = new Array[Byte](byteBuffer.remaining())
        byteBuffer.get(array)
        array
      }
    }
  }
```

jkobejs:
Fixes #1347


Pask423 (approved):

Pask423:
Thanks!

PR #1362, opened 2022-03-20
Attempt to fix binary downloads for scala native

### LINKED ISSUE #131: ScalaNative: Unable to download binary files
When using sttp with ScalaNative, I'm attempting to download a JAR from a URL:

```scala
sttp.get(jar).response(asFile(outputFile)).send().unsafeBody
```
The output of `outputFile` is a 5 byte file.  The output to the console is (which should be another bug as there doesn't seem to be a way to turn it off):

```
Downloading http://repo1.maven.org/maven2/com/outr/jefe-runner_2.12/1.2.4/jefe-runner_2.12-1.2.4.jar...
*   Trying 151.101.48.209...
* TCP_NODELAY set
* Connected to repo1.maven.org (151.101.48.209) port 80 (#0)
> GET /maven2/com/outr/jefe-runner_2.12/1.2.4/jefe-runner_2.12-1.2.4.jar HTTP/1.1
Host: repo1.maven.org
Accept: */*
Accept-Encoding: gzip, deflate

< HTTP/1.1 200 OK
< ETag: "be50aa1ecb13b6ec56baa9f3f0eb6fa3"
< Content-Type: application/java-archive
< Last-Modified: Mon, 23 Apr 2018 18:48:49 GMT
< X-Checksum-MD5: be50aa1ecb13b6ec56baa9f3f0eb6fa3
< X-Checksum-SHA1: 6251244a4a38c1ec1b4ccdb784abf897e27afd63
< Via: 1.1 varnish
< Content-Length: 23591
< Accept-Ranges: bytes
< Date: Tue, 24 Jul 2018 20:27:51 GMT
< Via: 1.1 varnish
< Age: 140
< Connection: keep-alive
< X-Served-By: cache-iad2128-IAD, cache-dfw18645-DFW
< X-Cache: HIT, HIT
< X-Cache-Hits: 1, 1
< X-Timer: S1532464072.507182,VS0,VE1
< 
* Connection #0 to host repo1.maven.org left intact
```

Pask423:
closes #131


core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -431,9 +431,11 @@ class SpecifyAuthScheme[U[_], T, -R](hn: String, rt: RequestT[U, T, R], digestTa
...
+// TODO replace with tag
 case class RequestOptions(
     followRedirects: Boolean,
     readTimeout: Duration,
     maxRedirects: Int,
-    redirectToGet: Boolean
-)
\ No newline at end of file
+    redirectToGet: Boolean,
+    binaryFile: Option[String] = None
```
adamw, line 440: I'm not sure what is the purpose of this - do you want to determine if the response should be saved as a file? If so, maybe we can examine `ResponseAs`?

Pask423:
closes #131

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
```diff
@@ -82,12 +82,26 @@ trait SyncHttpTestExtensions {
 
   "download file" - {
 
-    "download a binary file using asFile" in {
-      withTemporaryNonExistentFile { file =>
-        val req = basicRequest.get(uri"$endpoint/download/binary").response(asFile(file))
-        val resp = req.send(backend)
-        md5FileHash(resp.body.right.get).map { _ shouldBe binaryFileMD5Hash }
-      }
+    // TODO this test do not verify actual downloaded file
```
adamw, line 85: it checks the md5, so it does?

core/src/main/scalanative/sttp/client3/AbstractCurlBackend.scala
```diff
@@ -57,45 +55,73 @@ abstract class AbstractCurlBackend[F[_]](monad: MonadError[F], verbose: Boolean)
...
+          setRequestBody(curl, request.body)
+          responseMonad.flatMap(lift(curl.perform)) { _ =>
+            curl.info(ResponseCode, spaces.httpCode)
+            val httpCode = StatusCode((!spaces.httpCode).toInt)
+            if (headers.ptr != null) headers.ptr.free()
+            multiPartHeaders.foreach(_.ptr.free())
+            free(spaces.httpCode.asInstanceOf[Ptr[CSignedChar]])
+            fclose(outputFilePtr)
+            curl.cleanup()
+            val responseMetadata = ResponseMetadata(httpCode, "", List.empty)
```
adamw, line 74: can't we get the response headers here?

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
Pask423: so maybe my description is incorrect as our binary file from test was not saved properly while tests was passing

core/src/main/scalanative/sttp/client3/internal/BinaryFileHelper.scala
```diff
@@ -0,0 +1,19 @@
+package sttp.client3.internal
+
+import sttp.client3.{MappedResponseAs, ResponseAs, ResponseAsFile, ResponseAsFromMetadata}
+
+object BinaryFileHelper {
```
adamw, line 5: FileHelper? Nothing specific to binary here :)

core/src/main/scalanative/sttp/client3/AbstractCurlBackend.scala
```diff
@@ -57,45 +55,73 @@ abstract class AbstractCurlBackend[F[_]](monad: MonadError[F], verbose: Boolean)
...
-            code = httpCode,
-            statusText = statusText,
-            headers = responseHeaders,
-            history = Nil,
-            request = request.onlyMetadata
-          )
-        }
+      BinaryFileHelper.getFilePath(request.response) match {
+        case Some(file) =>
+          val outputPath = file.toPath.toString
```
adamw, line 60: maybe extract two methods with the two implementations?

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
adamw: hm that's weird - how did the md5 match then? ;) I guess we can keep both tests

adamw: ah, maybe `resp.body.right.get` was `None`?

adamw:
Looks good :)

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
Pask423: I will check on current master give me some time

core/src/main/scalanative/sttp/client3/AbstractCurlBackend.scala
Pask423: nope as they were written with `curl.option(HeaderData, spaces.headersResp)` which we cannot set with current setting

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
Pask423: Yes you are right - to a degree, resp.body.right.get is good but `md5FileHash(resp.body.right.get)` is empty

core/src/main/scalanative/sttp/client3/AbstractCurlBackend.scala
Pask423: at least with reasonable effort it may be possible somehow with `HEADERFUNCTION` and `HEADERDATA`  but it will require creation of another `wdFunc`-like function and I am not sure if it will work

core/src/test/scalanative/sttp/client3/testing/SyncHttpTestExtensions.scala
Pask423: In fact there is some strange error with md5Hash function -> it always return empty string

Pask423: https://github.com/scala-native/scala-native/pull/2175

adamw:
Looks good! Would be great to have the headers as well, but if that's not possible, let's merge as-is :)

PR #1409, opened 2022-04-14
Moving Httpclient to core module

### LINKED ISSUE #1401: Consider moving the HttpClient backend to core/jvm
This would mean that OOTB sttp's users get a modern, asynchronous, websocket-capable backend, without any other dependencies (an important fact to emphasise in the docs!)

However, this would also require at least Java 11. But maybe it's high time to do that? Would require at least a minor version bump.

We could bundle the monix/zio/cats-effect implementations of the backend together with the `effects` artifacts, to further simplify and lower the entry barrier to using sttp. So e.g. to use the monix backend you would only need the `com.softwaremill.sttp % monix` dependency.

Pask423:
closes #1401


adamw:
Looks good so far :) I think we'll need to come up with a better package name than `sttp.client3.httpclient.HttpClientBackend` though (and is it in monix as well)

adamw:
Or maybe we should have `sttp.client3.HttpClientSyncBackend` at the top-level, and all the supporting classes inside `.internal.httpclient`?

Pask423:
Yeah I am aware. 
And packaging for effects is ok ?

adamw:
Hm I guess ... why wouldn't it be ok? :)

Pask423:
Just want to be sure

build.sbt
```diff
@@ -275,6 +269,12 @@ lazy val testServer = (projectMatrix in file("testing/server"))
 
 lazy val testServer2_13 = testServer.jvm(scala2_13)
 
+val scalajava8compactVersion: Option[(Long, Long)] => String = {
```
adamw, line 272: `compat`, not `compact`. But do we need this as we are targeting java 11?

core/src/main/scalajvm/sttp/client3/internal/httpclient/InputStreamBodyFromHttpClient.scala
```diff
@@ -1,13 +1,14 @@
...
 import sttp.client3.internal.{BodyFromResponseAs, FileHelpers, SttpFile}
 import sttp.client3.ws.{GotAWebSocketException, NotAWebSocketException}
-import sttp.client3.WebSocketResponseAs
 import sttp.model.ResponseMetadata
 import sttp.monad.syntax.MonadErrorValueOps
 import sttp.ws.WebSocket
 
+import java.io.{BufferedInputStream, ByteArrayInputStream, FileInputStream, InputStream}
+
 trait InputStreamBodyFromHttpClient[F[_], S] extends BodyFromHttpClient[F, S, InputStream] {
```
adamw, line 12: everything else is `private[client3], shouldn't this one also be private?

core/src/main/scalajvm/sttp/client3/quick.scala
```diff
@@ -1,5 +1,6 @@
 package sttp.client3
 
 object quick extends SttpApi {
-  lazy val backend: SttpBackend[Identity, Any] = HttpURLConnectionBackend()
+  lazy val backend: SttpBackend[Identity, Any] = HttpClientSyncBackend()
+  lazy val urlConnectionBackend: SttpBackend[Identity, Any] = HttpURLConnectionBackend()
```
adamw, line 5: let's remove this and keep just `backend`

adamw:
HttpClient tests are in a different package than implementations (e.g. `HttpClientSyncHttpTest`) - would be good to keep this consistent

core/src/main/scalajvm/sttp/client3/internal/httpclient/MultiPartBodyPublisher.java
```diff
@@ -1,4 +1,4 @@
-package sttp.client3.httpclient;
```
adamw, line 1: same here - maybe this can be package-private

core/src/main/scalajvm/sttp/client3/quick.scala
Pask423: but which one should be the exact backend: URL or HttpClient ?

adamw: we're keeping `HttpClient` of course :)

build.sbt
Pask423: Needed by scala 2_11 code

effects/monix/src/main/scalajvm/sttp/client3/httpclient/monix/MonixBodyFromHttpClient.scala
```diff
@@ -1,19 +1,20 @@
...
+import sttp.client3.WebSocketResponseAs
 import sttp.client3.impl.monix.MonixWebSockets
+import sttp.client3.internal.httpclient.BodyFromHttpClient
 import sttp.client3.internal.{BodyFromResponseAs, SttpFile}
 import sttp.client3.ws.{GotAWebSocketException, NotAWebSocketException}
-import sttp.client3.WebSocketResponseAs
 import sttp.model.ResponseMetadata
 import sttp.ws.{WebSocket, WebSocketFrame}
 
+import java.nio.file.StandardOpenOption
```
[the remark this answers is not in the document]
Pask423: In fact it seems that it was the only one

build.sbt
Pask423: I will have to created separate version of class only for 2.11, does not seem to be worth it

adamw: what is the problematic class?

Pask423: classes for httpclient in core
image

adamw: yeah ... well, I guess we've got to keep it unfortunately ;) until we drop 2.11 support at least

eugeniyk:
This PR breaks clients that are using java8 & `TryHttpURLConnectionBackend`:

java.lang.UnsupportedClassVersionError: sttp/client3/TryHttpURLConnectionBackend$ has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0.

if you're going this path, pls remove `TryHttpURLConnectionBackend` completely; otherwise I'd recommend to publish java8 & java11 separately (although for *.core it will be hard to maintain 2 stacks)

Pask423:
Since `3.6.0` we have fully switched to Java 11 in sttp.
If you want to be up to date then I recommend changing to Java 11 as well. 
If you cannot then use `3.5.2` as it is the last version which does not require Java 11.

PR #1411, opened 2022-04-15
General OpenTelemetry backend

### LINKED ISSUE #1402: Add a general opentelemetry backend
Just as we have the prometheus/opentracing backends. Maybe then the zio-opentelemetry one can be removed?

Pask423:
closes #1402


build.sbt
```diff
@@ -210,10 +210,7 @@ lazy val allAggregates = projectsWithOptionalNative ++
   playJson.projectRefs ++
   openTracingBackend.projectRefs ++
   prometheusBackend.projectRefs ++
-  zio1TelemetryOpenTelemetryBackend.projectRefs ++
-  zio1TelemetryOpenTracingBackend.projectRefs ++
-  zioTelemetryOpenTelemetryBackend.projectRefs ++
-  zioTelemetryOpenTracingBackend.projectRefs ++
+  openTelemetryBackend.projectRefs ++
```
adamw, line 213: we're missing openTracing here?

docs/backends/wrappers/zio-opentelemetry.md
```diff
@@ -1,53 +0,0 @@
-# zio-telemetry opentelemetry backend 
-
-To use, add the following dependency to your project (the `zio-*` modules depend on ZIO 2.x; for ZIO 1.x support, use `zio1-*`):
```
adamw, line 3: we should convert these docs to their general versions as well :)

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,79 @@
...
+        propagator.inject(Context.current(), carrier, setter)
+        val attributes = Attributes.of(
+          AttributeKey.stringKey("http.method"),
+          request.method.method,
+          AttributeKey.stringKey("http.url"),
+          request.uri.toString(),
+          AttributeKey.stringKey("component"),
+          "sttp3-client"
+        )
+        Span.current.addEvent("Starting the work.", attributes)
```
adamw, line 43: These event names look very artificial. We should come up with something better (what's the "standard" for http requests?) or make them configurable

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,79 @@
...
+        (span, ignored)
+      }
+      .flatMap { span =>
```
adamw, line 46: since this is a tuple, let's decosntruct it here with meaningful names

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,79 @@
...
+
+private class OpenTelemetryBackend[F[_], P](
+    delegate: SttpBackend[F, P],
+    tracer: Tracer
+) extends SttpBackend[F, P] {
+
+  private implicit val _monad: MonadError[F] = responseMonad
+  type PE = P with Effect[F]
+
+  def send[T, R >: PE](request: Request[T, R]): F[Response[T]] = {
```
adamw, line 23: the impl looks quite different from the zio one, but I suspect that's beacuse the zio library was doing some of the work that happens here? I'm not familiar with other opentelemetry integrations, if that's what is usually done, fine :)

build.sbt
Pask423: It is declared few lines above

adamw: ah ok sorry :)

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
Pask423: I used example from `io.opentelemetry.api.trace.Tracer` class with automatic context propagation it seemed reasonable

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,91 @@
...
+      request.method.method,
+      AttributeKey.stringKey("http.url"),
+      request.uri.toString(),
+      AttributeKey.stringKey("component"),
+      "sttp3-client"
+    )
+  }
+
+  private def requestLabel[R >: PE, T](request: Request[T, R]): String = {
+    s"${request.method.method}-${request.uri.path}"
```
adamw, line 75: `request.uri.path` is a `Seq`, we could probably render this in a nicer way. Do we need the `-` in there, as the usual http request line is sth like `GET /x/y/z`?

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,91 @@
...
+
+  def send[T, R >: PE](request: Request[T, R]): F[Response[T]] = {
+    val carrier: mutable.Map[String, String] = mutable.Map().empty
+    val propagator: TextMapPropagator = openTelemetry.getPropagators.getTextMapPropagator
+    val setter: TextMapSetter[mutable.Map[String, String]] = (carrier, key, value) => carrier.update(key, value)
+    responseMonad
+      .eval {
+        val attributes = prepareBaseAttributes(request)
+        val span: Span = tracer
+          .spanBuilder(s"HTTP ${request.method.method}")
```
adamw, line 32: I think I'd make this configurable (with sane defaults), like other metrics backend are configurable

docs/backends/wrappers/opentelemetry.md
```diff
@@ -0,0 +1,29 @@
+# Opentelemetry backend
```
adamw, line 1: I think this also needs to be changed in toc in index.md?

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+      }
+  }
+
+  private def prepareBaseAttributes[R >: PE, T](request: Request[T, R]) = {
+    Attributes.of(
+      AttributeKey.stringKey("http.method"),
+      request.method.method,
+      AttributeKey.stringKey("http.url"),
+      request.uri.toString(),
+      AttributeKey.stringKey("component"),
```
adamw, line 70: is this attribute specified somewhere in https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/ ?

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+import sttp.client3._
+import sttp.monad.MonadError
+import sttp.monad.syntax._
+
+import scala.collection.mutable
+
+private class OpenTelemetryBackend[F[_], P](
+    delegate: SttpBackend[F, P],
+    openTelemetry: OpenTelemetry,
+    spanName: Option[String]
```
adamw, line 18: this should be a function `Request => String` defaulting to `HTTP $method`

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+        val span: Span = tracer
+          .spanBuilder(spanName.getOrElse(s"HTTP ${request.method.method}"))
+          .setSpanKind(SpanKind.CLIENT)
+          .setAllAttributes(attributes)
+          .startSpan
+        val scope = span.makeCurrent()
+        propagator.inject(Context.current(), carrier, setter)
+        (span, scope)
+      }
+      .flatMap { spanAndScope =>
```
adamw, line 41: `.flatMap { case (span, scope) => ...`

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+            .end()
+          scope.close()
+          responseMonad.error(e)
+        }
+      }
+  }
+
+  private def prepareBaseAttributes[R >: PE, T](request: Request[T, R]) = {
+    Attributes.of(
+      AttributeKey.stringKey("http.method"),
```
adamw, line 66: aren't these keys available via `SemanticAttributes` ?

Also, why do we set the method & url, but not the target, scheme, host, etc?

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
```diff
@@ -0,0 +1,98 @@
...
+          .setAllAttributes(attributes)
+          .startSpan
+        val scope = span.makeCurrent()
+        propagator.inject(Context.current(), carrier, setter)
+        (span, scope)
+      }
+      .flatMap { spanAndScope =>
+        val (span, scope) = spanAndScope
+        val label = requestLabel(request)
+        Span.current.addEvent(s"Start handling request: $label")
```
adamw, line 44: hm reading through the opentelemtry docs, I don't think the events add any value above what's in the request attributes

adamw:
Reading through https://opentelemetry.io/docs/instrumentation/java/manual/, I see that there's also a metrics api.

Maybe we should have both an `OpenTelemetryTracingBackend` and `OpenTelemetryMetricsBackend`? The latter with metrics similar as in the prometheus one, for example.

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
Pask423: No, I took it from opentracing

adamw: from what I read opentracing is deprecated so might not be the best inspiration :)

metrics/open-telemetry-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryBackend.scala
adamw: no, I saw the constant in the docs, maybe it's not in the api
though the example in the docs for a http client call sets only method & url, so maybe that's standard

Pask423:
Why not I will make it in separate PR.
Maybe we should deprecate/remove `OpenTracingBackend` as Open Tracing itself is deprecated in favor of Open Telemetry

adamw:
@Pask423 looks good. Fine with a separate PR and removing open tracing :)

PR #1414, opened 2022-04-19
Test server artifact

### LINKED ISSUE #1376: Publish test HttpServer within an artifact or convert to a sbt plugin
For 3rd party backends, it will be useful to publish [testing/server](https://github.com/softwaremill/sttp/blob/master/testing/server/src/main/scala/sttp/client3/testing/server/HttpServer.scala) as an artifact (maybe within test artifact?). I had to copy its source code to test out compatibility.
Or it could be converted into a sbt plugin which will make more sense?

Pask423:
closes #1376

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -101,13 +101,13 @@ val enableMimaSettings = Seq(
 // things, we always start the test server.
 val testServerSettings = Seq(
   Test / test := (Test / test)
-    .dependsOn(testServer2_13 / startTestServer)
+    .dependsOn(testServer / startTestServer)
     .value,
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -257,26 +258,23 @@ lazy val rootProject = (project in file("."))
   )
   .aggregate(allAggregates: _*)
 
-lazy val testServer = (projectMatrix in file("testing/server"))
+lazy val testServer = (project in file("testing/server"))
```
adamw, line 261: since we are publishing this, maybe we can do so for both 2.12 and 2.13?

PR #1423, opened 2022-04-29
Fix PUT method in CurlBackend

aartigao:
I'm working on a CLI interface with Scala Native that issues request to an HTTP server, with the unfortunate luck that the first use case is a PUT request with body. Using CurlBackend for this scenario leaves the client completely frozen (I guess some C internals that I don't understand).

Looks like it's not working because deprecated flags are being used in calls to `libcurl`. Following the recommended example in [httpput-postfields](https://curl.se/libcurl/c/httpput-postfields.html) it's clear that we need to use a custom request.


aartigao:
More information about the deprecation [#6186](https://github.com/curl/curl/issues/6186)

Pask423 (approved):

Pask423:
Thanks!

PR #1428, opened 2022-05-02
Opentelemetry metrics backend

Pask423:


build.sbt
```diff
@@ -891,26 +890,27 @@ lazy val prometheusBackend = (projectMatrix in file("metrics/prometheus-backend"
   .jvmPlatform(scalaVersions = scala2 ++ scala3)
   .dependsOn(core)
 
-lazy val openTracingBackend = (projectMatrix in file("metrics/open-tracing-backend"))
+lazy val openTelemetryTracingBackend = (projectMatrix in file("metrics/open-telemetry-tracing-backend"))
```
adamw, line 893: if the dependencies are the same I think it can be a single project

docs/backends/wrappers/opentelemetrymetrics.md
```diff
@@ -0,0 +1,37 @@
+# Opentelemetry metrics backend
```
adamw, line 1: same here, I would just have a single "OpenTelemetry" page

metrics/open-telemetry/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryMetricsBackend.scala
```diff
@@ -0,0 +1,175 @@
...
+  val DefaultFailureCounterName = "sttp3_requests_failure_count"
+  val DefaultRequestHistogramName = "sttp3_request_size_bytes"
+  val DefaultResponseHistogramName = "sttp3_response_size_bytes"
+
+  def apply[F[_], P](
+      delegate: SttpBackend[F, P],
+      openTelemetry: OpenTelemetry,
+      meterConfig: Option[MeterConfig] = None,
+      requestToInProgressCounterNameMapper: Request[_, _] => Option[CollectorConfig] = (_: Request[_, _]) =>
+        Some(CollectorConfig(DefaultRequestsInProgressCounterName)),
```
adamw, line 152: shouldn't we provide some units as default?

adamw (approved):

metrics/open-telemetry/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryMetricsBackend.scala
Pask423: In fact unit itself is optional and we are using it in many contexts so I am not sure what exactly should be used as default unit

PR #1446, opened 2022-05-24
Update version in docs for v2

magdzikk:
I noticed that when browsing the docs for v2 (https://sttp.softwaremill.com/en/v2/quickstart.html), the page title in the browser is "sttp 1.0 documentation" which is a bit misleading.


adamw:
Thanks :)

PR #1474, opened 2022-06-26
Drop java8-compat dependency

Pask423:
There is a problem with conflicting cross-version for java8-compat dependency on scala-3.
In fact we only a few converters so I deiced to write our own converters for sttp and drop whole dependency.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
-import com.softwaremill.UpdateVersionInDocs
-import com.softwaremill.Publish.{updateDocs, ossPublishSettings}
+import com.softwaremill.Publish.{ossPublishSettings, updateDocs}
 import com.softwaremill.SbtSoftwareMillCommon.commonSmlBuildSettings
+import com.softwaremill.UpdateVersionInDocs
 import sbt.Keys.publishArtifact
 import sbt.Reference.display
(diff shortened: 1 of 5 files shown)


adamw:
Thanks!

loicdescotte:
Hi,
Unfortunately we are still using Java 8 for some clients, if I understand correctly, the problem is with Scala 3, would it be possible to keep java 8 support for Scala 2.13?
I know it's not offically supported anymore, but it would be really great... Thanks!

adamw:
Sorry, I doubt it, we've included the HttpClient-based client in the `core` module (this is now the default backend). And that implementation is only available with Java11+, so we need to compile using Java 11.

If you really need this, maybe you can try backporting zio2 support to 3.5.x. If you'll have a working branch, I can release it as the next 3.5.x patch release.

adamw:
I've created a 3.5 branch from the latest 3.5 release: https://github.com/softwaremill/sttp/tree/3.5

loicdescotte:
> If you really need this, maybe you can try backporting zio2 support to 3.5.x. If you'll have a working branch, I can release it as the next 3.5.x patch release.

Thanks I will try this, that would be great!

PR #1482, opened 2022-06-30
Zio package sharing

### LINKED ISSUE #851: Move ZIO package types into the shared project
Currently the two `ZIO` backends have identical package structures:

```scala
package object zio {

  /** ZIO-environment service definition, which is an SttpBackend.
    */
  type SttpClient = Has[SttpClient.Service]
  type SttpClientStubbing = Has[SttpClientStubbing.Service]

  object SttpClient {
    type Service = SttpBackend[Task, ZioStreams with WebSockets]
  }

  /** Sends the request. Only requests for which the method & URI are specified can be sent.
    *
    * @return An effect resulting in a [[Response]], containing the body, deserialized as specified by the request
    *         (see [[RequestT.response]]), if the request was successful (1xx, 2xx, 3xx response codes), or if there
    *         was a protocol-level failure (4xx, 5xx response codes).
    *
    *         A failed effect, if an exception occurred when connecting to the target host, writing the request or
    *         reading the response.
    *
    *         Known exceptions are converted to one of [[SttpClientException]]. Other exceptions are kept unchanged.
    */
  def send[T](
      request: Request[T, Effect[Task] with ZioStreams with WebSockets]
  ): ZIO[SttpClient, Throwable, Response[T]] =
    ZIO.accessM(env => env.get[SttpClient.Service].send(request))

  /** A variant of [[send]] which allows the effects that are part of the response handling specification (when
    * using websockets or resource-safe streaming) to use an `R` environment.
    */
  def sendR[T, R](
      request: Request[T, Effect[RIO[R, *]] with ZioStreams with WebSockets]
  ): ZIO[SttpClient with R, Throwable, Response[T]] =
    ZIO.accessM(env => env.get[SttpClient.Service].extendEnv[R].send(request))

  object SttpClientStubbing extends SttpClientStubbingBase[Any, ZioStreams with WebSockets] {
    override private[sttp] def serviceTag: Tag[SttpClientStubbing.Service] = implicitly
    override private[sttp] def sttpBackendTag: Tag[SttpClient.Service] = implicitly
  }

  object stubbing {
    import SttpClientStubbing.StubbingWhenRequest

    def whenRequestMatches(p: Request[_, _] => Boolean): StubbingWhenRequest =
      StubbingWhenRequest(p)

    val whenAnyRequest: StubbingWhenRequest =
      StubbingWhenRequest(_ => true)

    def whenRequestMatchesPartial(
        partial: PartialFunction[Request[_, _], Response[_]]
    ): URIO[SttpClientStubbing, Unit] =
      ZIO.accessM(_.get.whenRequestMatchesPartial(partial))
  }
}
```

It seems like this could be moved into the shared zio package so that we don't need to pull in a specific zio backend just to get access to the environmental capabilities.

Pask423:
closes #851


docs/backends/zio.md
```diff
@@ -136,35 +137,30 @@ Please visit [the official documentation](https://armeria.dev/docs/client-factor
 
 ## ZIO environment
 
-As an alternative to effectfully or resourcefully creating backend instances, ZIO environment can be used. In this case, a type alias is provided for the service definition:
+As an alternative to effectfully or resourcefully creating backend instances, ZIO environment can be used. In this case, a type aliases are provided for the service definition:
```
felix-hedenstrom, line 140: Should this be `In this case, type aliases are provided for the service definition`?

felix-hedenstrom: Maybe `the following type aliases are provided for the service definition` is a better fit.

adamw:
I'm not entirely convinced by the new naming. The "default" zio backend is the httpclient one - which has websocket capabilities. As it's the default one, its usage should be as simple as possible. Meaning, that it would be better to retain the `send` method and `SttpClient` environmental requirement. The armeria backend is then the "exception", which could require different handling. Not sure what the naming could be, though.

adamw:
Closing this one as the `SttpClient` type alias is gone for zio2 (see https://github.com/softwaremill/sttp/pull/1499), and for zio1 we'll simply keep the existing code

PR #1497, opened 2022-07-13
Drop target 1.8 for Scala 2.12

Pask423:
As `target:jvm-11` or `target:jvm-1.11` seems to not be supported for  Scala 2.11 and 2.12 I decided to drop it as we switch to Java 11 since `3.6.0`


adamw:
Seems there are some weird errors for 2.11 ... maybe we could keep the flag for 2.11 only?

build.sbt
```diff
@@ -296,7 +295,9 @@ lazy val core = (projectMatrix in file("core"))
         Test / publishArtifact := true, // allow implementations outside of this repo
         scalacOptions ++= Seq("-J--add-modules", "-Jjava.net.http"),
         scalacOptions ++= {
-          if (scalaVersion.value == scala2_13) List("-target:jvm-11") else Nil
+          if (scalaVersion.value == scala2_13) List("-target:jvm-11")
```
adamw, line 298: and for scala3 not?

Pask423: Will check

Issue #1529, opened 2022-08-18
Simple http backend without HKT


szymon-rd:
Right now, when a programmer that is not particularly familiar with concept of HKT uses sttp (e.g. with importing `sttp.client3.okhttp.quick._`, he sees a suggestion with the F[...] as the return type:
![image](https://user-images.githubusercontent.com/4761866/185377842-b45d8457-bf99-4599-ba1b-ae5b76e189f8.png)
It may lead to imposing a higher cognitive load and introducing him to a type of abstraction that he may not be familiar with. 
My proposal would be to introduce a `send` method on the `HttpClientSyncBackend`. 
```scala
val backend = HttpClientSyncBackend()
val request = basicRequest
  .body("Hello, world!")
  .post(uri"https://httpbin.org/post?hello=world")
val response = backend.send(request)
```
That way it could be possible to introduce a signature with just the Response as the return type. 

Additional note: Another idea that could smoothen the onboarding to the library is maybe creating a simplest backend named `client` instead of `backend` (with the api described above). 
```scala
val httpClient = SimpleHttpClient()
val request = basicRequest
  .body("Hello, world!")
  .post(uri"https://httpbin.org/post?hello=world")
val response = httpClient.send(request)
```
I would like to request a comment on that as it is a proposal that could lead to losing some consistency of naming. However, refering as `backend` to the logic that dispatches the request could sometimes lower the role expressiveness of code using sttp - term `backend` in http domain most often refers to the destination of the request.

adamw:
We can't add a different `send` method to the `HttpClientSyncBackend` backend, or rather: we can, but this wouldn't have the desired effect :) That's because we never return the immediate implementation from the companion object methods. Instead, backends are by default wrapped with `FollowRedirectsBackend`, which implement redirects. Plus you might want to use other wrappers (logging, metrics, retries), which would cause loosing the "simple" alternative.

However, the "recommended" (at least in the docs) way to send requests is using `request.send(backend)`, not `backend.send(request)` (though they are equivalent). So probably the above is not a problem.

There are two routes that we might take.

## Add a synchronous `send` to `Request`

We could add a synchronous variant of `send` to `RequestT`:

```scala
case class RequestT(...) {
  def send[F[_], P](backend: SttpBackend[F, P])(implicit
      isIdInRequest: IsIdInRequest[U],
      pEffectFIsR: P with Effect[F] <:< R
  ): F[Response[T]] = ???

[...]

PR #1532, opened 2022-08-19
DX: fix some issues

Fristi:
- Seems with `zio` module the alias `SttpClient` was removed, brought this back to have smoother migration experience
- The `zio1` stub does not include `WebSockets` .. this causes compilation issues 

**Check list**
- [ ] Check if the project compiles by running `sbt compile` (exited due `java.lang.OutOfMemoryError: Java heap space` .. maybe include a `.sbtopts` to mitigate this?)

### DIFF
--- a/effects/zio/src/main/scalajvm/sttp/client3/httpclient/zio/package.scala
+++ b/effects/zio/src/main/scalajvm/sttp/client3/httpclient/zio/package.scala
@@ -1,0 +1,10 @@
+package sttp.client3.httpclient
+
+import sttp.capabilities.WebSockets
+import sttp.capabilities.zio.ZioStreams
+import sttp.client3.SttpBackend
+
+
+package object zio {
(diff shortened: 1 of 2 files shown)


adamw:
Thanks! How would the `HttpClient` alias be used? Maybe you could adjust the example in https://github.com/softwaremill/sttp/blob/master/examples/src/main/scala/sttp/client3/examples/GetAndParseJsonZioCirce.scala

Fristi:
`def sendAndPrint(backend: SttpClient): Task[Unit]` in this case, but I also like the `SttpBackend[Task, Any]` .. but from a DX perspective you'll just annotate your layers and service implementaties with the types provided by what sttp gives you (`SttpBackend[Task, ZioStreams with WebSockets]`)

adamw:
Thanks, I'll merge this but remove the type alias for now, for two reasons:

* other zio backends support different capabilities
* the name might clash with #1529

PR #1542, opened 2022-08-31
Added reading username and password for http and https proxy

### LINKED ISSUE #403: async-http-client cannot detect proxy settings
HttpURLConnectionBackend works fine without any configuration. 
But when I try to use async-http-client I get `HTTP/1.1 407 Proxy Authentication Required`
So I need to set 
```
options = SttpBackendOptions.httpProxy()
```
manually
Could async-http-client work the same way as HttpURLConnectionBackend?

Pask423:
closes #403 
If I understood problem correctly


adamw:
I don't think we have a sane way of testing this so let's see if this works :)

PR #1545, opened 2022-09-02
Add a synchronous SimpleHttpClient

### LINKED ISSUE #1529: Simple http backend without HKT

adamw:
Closes #1529

### DIFF
--- a/core/src/main/scalajvm/sttp/client3/SttpClient.scala
+++ b/core/src/main/scalajvm/sttp/client3/SttpClient.scala
@@ -1,0 +1,36 @@
+package sttp.client3
+
+/** A simple, synchronous http client. Usage example:
+  *
+  * {{{
+  * import sttp.client3.{SttpClient, UriContext, basicRequest}
+  *
+  * val client = SttpClient()
+  * val response = basicRequest.get(uri"https://httpbin.org/get").send(client)
+  * println(response.body)
+  * }}}
+  *
+  * Wraps a [[SttpBackend]], which can be substituted or modified using [[wrapBackend]], adding e.g. logging.
+  *
+  * Creating a client allocates resources (a pool of HTTP connections), hence when no longer needed, the client should
+  * be closed using [[close]].
+  */
+case class SttpClient(backend: SttpBackend[Identity, Any]) {
+  def withBackend(newBackend: SttpBackend[Identity, Any]): SttpClient = copy(backend = newBackend)
+  def wrapBackend(f: SttpBackend[Identity, Any] => SttpBackend[Identity, Any]): SttpClient = copy(backend = f(backend))
+
+  def close(): Unit = backend.close()
+}
+
+object SttpClient {
+  def apply(): SttpClient = SttpClient(HttpClientSyncBackend())
+
+  /** Runs the given function `f` with a new, default instance of [[SttpClient]] and closes the client after the
+    * function compeltes, cleaning up any resources.
+    */
+  def apply[T](f: SttpClient => T): Unit = {
+    val client = SttpClient()
+    try f(client)
+    finally client.close()
+  }
+}
--- a/core/src/main/scalajvm/sttp/client3/SttpClient.scala
+++ b/core/src/main/scalajvm/sttp/client3/SttpClient.scala
@@ -13,8 +13,7 @@ package sttp.client3
   *
   * Wraps a [[SttpBackend]], which can be substituted or modified using [[wrapBackend]], adding e.g. logging.
   *
-  * Creating a client allocates resources (a pool of HTTP connections), hence when no longer needed, the client should
-  * be closed using [[close]].
+  * Creating a client allocates resources, hence when no longer needed, the client should be closed using [[close]].
   */
 case class SttpClient(backend: SttpBackend[Identity, Any]) {
 
--- a/core/src/main/scalajvm/sttp/client3/SttpClient.scala
+++ b/core/src/main/scalajvm/sttp/client3/SttpClient.scala
@@ -6,7 +6,8 @@ package sttp.client3
   * import sttp.client3.{SttpClient, UriContext, basicRequest}
   *
   * val client = SttpClient()
-  * val response = basicRequest.get(uri"https://httpbin.org/get").send(client)
+  * val request = basicRequest.get(uri"https://httpbin.org/get")
+  * val response = client.send(request)
   * println(response.body)
   * }}}
   *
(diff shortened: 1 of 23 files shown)


examples/src/main/scala/sttp/client3/examples/SimpleClientGetAndPost.scala
```diff
@@ -0,0 +1,32 @@
...
+
+    println("---\n")
+
+    //
+
+    val request2: Request[String, Any] = basicRequest
+      .header("X-Correlation-ID", UUID.randomUUID().toString)
+      .response(asStringAlways)
+      .body("Hello, world!")
+      .post(uri"https://httpbin.org/post")
```
szymon-rd, line 26: Is there a way to avoid writing  the`uri` interpolator and just pass String? E.g. additional method on Request for that. I think it could simplify the usage further for some basic cases.

core/src/main/scalajvm/sttp/client3/SttpClient.scala
```diff
@@ -0,0 +1,39 @@
...
+  * val request = basicRequest.get(uri"https://httpbin.org/get")
+  * val response = client.send(request)
+  * println(response.body)
+  * }}}
+  *
+  * Wraps a [[SttpBackend]], which can be substituted or modified using [[wrapBackend]], adding e.g. logging.
+  *
+  * Creating a client allocates resources, hence when no longer needed, the client should be closed using [[close]].
+  */
+case class SttpClient(backend: SttpBackend[Identity, Any]) {
```
szymon-rd, line 18: I have some second thoughts about the name of this client. I wonder if it could be better to somehow have the `Http` word in it.  The role expressiveness could be higher that way - The ideal situation is when somebody entirely unfamiliar with the library knows what a snippet does. `SttpClient` could indicate that it is a client for `STTP` protocol. Looking at our other options: `SttpHttpClient` is a bit hard to read and understand. `HttpClient` is taken by Java and could be misleading, but for example in os-lib there are types like File and it does not pose many problems. I am not entirely sure about it though, it's a tough call.

examples/src/main/scala/sttp/client3/examples/SimpleClientGetAndPost.scala
adamw: I would prefer not to - people will end up using string interpolation to create URIs, which quickly leads to problems with encoding URI fragments etc. The whole point of the `uri` interpolator is that it is context-aware, and treats anything that is interpolated appropriately as a path/query/etc. fragment.

core/src/main/scalajvm/sttp/client3/SttpClient.scala
adamw: Yeah naming is hard ;) There's also `SimpleHttpClient` which would indicate that the more "advanced" use cases require a different mechanism

szymon-rd: I agree, I think that we can go with the `SimpleHttpClient` or maybe even `BasicHttpClient`.

examples/src/main/scala/sttp/client3/examples/SimpleClientGetAndPost.scala
szymon-rd: I agree. However, it still leaves me wondering if there is a way to simplify that, as it is the only element that has limited visibility - all the rest can be imported via autosuggestions and has 1:1 name in code - name in import relation.

szymon-rd: Oh, ok. I had a problem when importing it with Metals, but I agree that most developers are going to use IntelliJ, and if it works great there then that's not an issue.

docs/simple_sync.md
```diff
@@ -0,0 +1,51 @@
...
+
+TODO
+
+## Relationship with backends
+
+The `SttpClient` serves as a simple starting point for sending requests in a synchronous way. For more advanced 
+use-cases, you should use an [sttp backend](backends/summary.md) directly. For example, if you'd like to send requests 
+asynchronously, getting a `Future` as the result. Or, if you manage side effects using an `IO` or `Task`.
+
+In fact, an instance of `SttpClient` is a think wrapper on top of a backend.
```
prolativ:
```suggestion
In fact, an instance of `SttpClient` is a thin wrapper on top of a backend.
```

examples/src/main/scala/sttp/client3/examples/SimpleClientGetAndPost.scala
prolativ: Could we define `uri` as an extension method instead of a method inside an implicit class? Then one would import `sttp.client3.uri` instead of `sttp.client3.UriContext`, which would feel like less magic.

docs/simple_sync.md
adamw: thx

examples/src/main/scala/sttp/client3/examples/SimpleClientGetAndPost.scala
adamw: You mean in the Scala 3 version instead of https://github.com/softwaremill/sttp-model/blob/master/core/src/main/scala/sttp/model/UriInterpolator.scala#L10 ?

I didn't know interpolators could be added as an extension methods, but I guess it makes sense. 

Anyway, wouldn't that break binary compatibility? (plus the docs are generated using Scala 2 so you'd end up having different code for Scala 2 / Scala 3)

szymon-rd: Removing the old one and adding a new one would be breaking the binary compatibility, there is a possibility of adding a new one next to the old one at the cost of possible confusion. In Scala 2 sources there would be a need to hack a bit and make a new implicit class called just `uri` that would delegate to the old code.

adamw: Ah of course, creating an `uri` class would do the trick. I like the idea, but I think we should do it in sttp4, adding it here would have small usability impact and as you write - might be confusing in the import suggestiosn.

adamw:
@szymon-rd If this looks "good enough" I'll merge & release?

szymon-rd (approved):
Yes, it looks great!

PR #1552, opened 2022-09-07
Absorb exceptions while mapping responses with impure functions

mschuwalow:

In certain cases `MappedResponseAs` will have impure functions embedded within it.
For example `getRight` is implemented in the following way:
```
ra.mapWithMetadata { case (t, meta) =>
  t match {
    case Left(a: Exception) => throw a
    case Left(a)            => throw HttpError(a, meta.code)
    case Right(b)           => b
  }
}
```
The exception will end up getting thrown inside of the map function of the target monad, which is not handled nicely by all effect types that sttp supports. For example for zio, this will result in the fiber dying instead of failing


mschuwalow:
What is the correct place to place tests for this? I had a brief look but didn't see any test suite dedicated to the various responseas data structures

adamw:
@mschuwalow Thanks; most of the testing is done on a more integration level, when testing a backend, see `HttpTest` and its usages

adamw:
Thanks :)

PR #1553, opened 2022-09-07
fix: consume response entity on response processing failure

### LINKED ISSUE #1450: Can't read/discard body when ReadException is thrown as SttpClientException doesn't provide access to the request object
I’m using sttp3 (version 3.5.2) with the akka-http backend. Every now and then I see log messages like the following:

```
Response entity was not subscribed after 1 second. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`
```

I understand that these log message are emitted by akka-http. With the debugger I narrowed it down to the code throwing a `SttpClientException.ReadException`. While the `SttpClientException` class has the [request object](https://github.com/softwaremill/sttp/blob/46899565745ab1f0a9d19e6dcc656ba129558dc5/core/src/main/scala/sttp/client3/SttpClientException.scala#L24) it is not accessible, making it impossible to get to the response object.

geminicaprograms:
*Issue:*
When exception is thrown during the response entity processing it may or may not be consumed and as a result the following warning is being displayed to the console:

```
  Response entity was not subscribed after 1 second. Make sure to read the response `entity` body or call `entity.discardBytes()` on it -- in case you deal with `HttpResponse`, use the shortcut `response.discardEntityBytes()`
```

*Solution:*
Add handler that takes care of exhausting the response entity when exception is thrown during the response processing.

*Notes:*
* the original processing exception is propagated
* in case when exception is thrown while the response entity is consumed the original processing exception is propagated

Closes #1450

### DIFF
--- a/akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
+++ b/akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
@@ -11,7 +11,7 @@ import akka.http.scaladsl.model.{StatusCode => _, _}
 import akka.http.scaladsl.settings.ConnectionPoolSettings
 import akka.http.scaladsl.{ClientTransport, Http, HttpsConnectionContext}
 import akka.stream.Materializer
-import akka.stream.scaladsl.Flow
+import akka.stream.scaladsl.{Flow, Sink}
 import sttp.capabilities.akka.AkkaStreams
 import sttp.capabilities.{Effect, WebSockets}
 import sttp.client3
@@ -54,9 +54,13 @@ class AkkaHttpBackend private (
       .fromTry(ToAkka.request(r).flatMap(BodyToAkka(r, r.body, _)))
       .map(customizeRequest)
       .flatMap(request =>
-        http.singleRequest(request, connectionSettings(r)).map(response => customizeResponse(request, response))
+        http
+          .singleRequest(request, connectionSettings(r))
+          .flatMap(response =>
+            Future(customizeResponse(request, response))
+              .flatMap(response => responseFromAkka(r, response, None).recoverWith(consumeResponseOnFailure(response)))
+          )
       )
-      .flatMap(responseFromAkka(r, _, None))
   }
 
   private def sendWebSocket[T, R >: PE](r: Request[T, R]): Future[Response[T]] = {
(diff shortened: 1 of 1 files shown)


akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
```diff
@@ -54,9 +54,13 @@ class AkkaHttpBackend private (
       .fromTry(ToAkka.request(r).flatMap(BodyToAkka(r, r.body, _)))
       .map(customizeRequest)
       .flatMap(request =>
-        http.singleRequest(request, connectionSettings(r)).map(response => customizeResponse(request, response))
+        http
+          .singleRequest(request, connectionSettings(r))
+          .flatMap(response =>
+            Future.apply(customizeResponse(request, response)).recoverWith(consumeResponseOnFailure(response))
```
adamw, line 60: You can combine these two cases to recover only once, by flatMapping this future with the `responseFromAkka` funciton

akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
```diff
@@ -78,13 +82,22 @@ class AkkaHttpBackend private (
...
-          responseFromAkka(r, response, None)
+          responseFromAkka(r, response, None).recoverWith(consumeResponseOnFailure(response))
       }
   }
 
+  private def consumeResponseOnFailure[T](response: HttpResponse): PartialFunction[Throwable, Future[T]] = {
+    case t: Throwable =>
+      response.entity.dataBytes
+        .runWith(Sink.ignore)
+        .map(_ => ())
```
adamw, line 96: the `.map` is not necessary

akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
geminicaprograms: you mean that I can perform _recovery_ after the last `.flatMap` operation and as a result end up with:
```
      ...
      .flatMap(request =>
        http.singleRequest(request, connectionSettings(r)).map(response => customizeResponse(request, response))
      )
      .flatMap(response => responseFromAkka(r, response, None).recoverWith(consumeResponseOnFailure(response)))
```
right?

geminicaprograms: Done

akka-http-backend/src/main/scala/sttp/client3/akkahttp/AkkaHttpBackend.scala
geminicaprograms: Done

PR #1567, opened 2022-09-15
Add status text resolver

### LINKED ISSUE #1451: statusText for Response is always an empty string
Using the latest version of `async-http-client-backend-future % 3.6.2`

I always get an empty string for the `statusText` response.

```
case object Boot extends App {
  val backend: SttpBackend[Future, Any] = HttpClientFutureBackend(
    options = SttpBackendOptions.connectionTimeout(new FiniteDuration(2, SECONDS))
  )

  val response = basicRequest.get(uri"https://lgr9j.mocklab.io/json/1")
    .send(backend)

  val a = Await.result(response, Duration.Inf)
  assert(a.code.code == 200, s"Response code should be a 200, got ${a.statusText}")
  assert(a.statusText == "OK", s"Response statusText should be a OK, got ${a.statusText}")

}
```

Expected result: 
Result statusText should be based on the Result code
200 - OK
404 - Not Found
500 - Internal Server Error

Pask423:
closes #1451

### DIFF
--- a/core/src/main/scala/sttp/client3/Response.scala
+++ b/core/src/main/scala/sttp/client3/Response.scala
@@ -64,7 +64,7 @@ object Response {
     */
   def ok[T](body: T): Response[T] = apply(body, StatusCode.Ok)
 
-  def resolveStatusText(statusCode: StatusCode, provided: String = ""): String =
+  private def resolveStatusText(statusCode: StatusCode, provided: String = ""): String =
     if (provided.isEmpty) StatusText.default(statusCode).getOrElse(provided)
     else provided
 }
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sttp/client3/Response.scala
```diff
@@ -45,22 +45,26 @@ object Response {
...
-    Response(body, code, statusText, headers, Nil, ExampleGet)
+    Response(body, code, resolveStatusText(code, statusText), headers, Nil, ExampleGet)
 
   /** Convenience method to create a Response instance, mainly useful in tests using
     * [[sttp.client3.testing.SttpBackendStub]] and partial matchers.
     */
-  def ok[T](body: T): Response[T] = apply(body, StatusCode.Ok, "OK")
+  def ok[T](body: T): Response[T] = apply(body, StatusCode.Ok)
+
+  def resolveStatusText(statusCode: StatusCode, provided: String = ""): String =
```
adamw, line 67: should be private?

adamw:
Thanks :)

Issue #1571, opened 2022-09-16
async-http-client Netty v4.1.60.Final vulnerabilities


paoloboni:
Hi,
If I read correctly the dependency tree, `async-http-client-backend` currently depends on `async-http-client` v2.12.3, which pulls `netty-codec` v4.1.60.Final, which according to Sonatype Lift is subject to the following vulnerabilities:

- [[sonatype-2021-0789] CWE-119: Improper Restriction of Operations within the Bounds of a Memory Buffer](https://ossindex.sonatype.org/vulnerability/sonatype-2021-0789?component-type=maven&component-name=io.netty%2Fnetty-codec&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0)
- 	[[CVE-2021-37136] CWE-400: Uncontrolled Resource Consumption ('Resource Exhaustion')](https://ossindex.sonatype.org/vulnerability/CVE-2021-37136?component-type=maven&component-name=io.netty%2Fnetty-codec&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0)
- 	[[CVE-2021-37137] CWE-400: Uncontrolled Resource Consumption ('Resource Exhaustion')](https://ossindex.sonatype.org/vulnerability/CVE-2021-37137?component-type=maven&component-name=io.netty%2Fnetty-codec&utm_source=ossindex-client&utm_medium=integration&utm_content=1.7.0)

Those are high severity vulnerabilities that should be fixed in async-http-client, but it looks like it's [no longer getting updated](https://github.com/AsyncHttpClient/async-http-client/issues/1802#issuecomment-961023802), at least in the last year or so.

adamw:
I suppose since AHC is no longer maintained, we should depracte the client. And remove its usages from examples. What do you think, @Pask423 ?

Pask423:
Great idea, but I would leave an example anyway - just in case.

rafalambrozewicz:
I'm against leaving the examples because even if we would add a big, bold, shouting "IT'S DEPRECATED" note, a number of people would not notice it, starting using it thinking, that it is ok... but it is not ok. If magically AsyncHttpClient would be updated, we can always revert these changes. What do you think?

zendesk-thomas-hammond:
Apologies for posting on a closed issue, but is there a recommended upgrade path to move off the deprecated module?

adamw:
[...]

PR #1589, opened 2022-10-06
1571 async http client netty v4160final vulnerabilities

rafalambrozewicz:
Since `AsyncHttpClient` is no longer maintained and has major vulnerabilities, it has been dedicated to deprecate the code that is using it and to remove its usage examples and partly delete documentation related to it and to mark the rest of it so it is clear for the reader, that given solution is deprecated.

* Added `@deprecated` annotation to all entry-point companion objects for `AsyncHttpClient` based backends.
* Rewritten examples that were using `async-http-client`, so now they use`HttpClientBackend`.
* Partly removed  `async-http-client` documentation, partly marked given docs, so it is clear for a reader that the given solution is deprecated.
* In WebSocket examples used new echo service, since the old one is no longer available.


async-http-client-backend/cats-ce2/src/main/scala/sttp/client3/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
```diff
@@ -24,6 +24,7 @@ import sttp.client3.testing.SttpBackendStub
 import sttp.monad.MonadAsyncError
 import sttp.ws.WebSocket
 
+@deprecated(message = "AsyncHttpClient no longer maintained")
```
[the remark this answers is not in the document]
adamw: + I think we would really need to annotate the public `apply` methods from the companion objects, as that's how users create the backends

docs/websockets.md
```diff
@@ -91,16 +91,6 @@ configuring individual backends for more information.
 
 ## Implementation-specific configuration
 
-### async-http-client based backends
```
adamw, line 94: I think I would leave that - if somebody already uses the AHC backend, they might be looking for its config options

docs/resilience.md
```diff
@@ -32,7 +32,6 @@ Some backends have built-in retry mechanisms:
 
 * [akka-http](https://doc.akka.io/docs/akka-http/current/scala/http/client-side/host-level.html#retrying-a-request)
 * [OkHttp](http://square.github.io/okhttp) (see the builder's `retryOnConnectionFailure` method)
-* async-http-client: by default, the backend will attempt 5 retries in case an `IOException` is thrown during the connection. This can be changed by specifying the `org.asynchttpclient.maxRequestRetry` config option, or by providing custom configuration using when creating the backend (`setMaxRequestRetry`).
```
adamw, line 35: same here - I would leave this section

docs/conf/ssl.md
```diff
@@ -102,31 +102,6 @@ val backend = AkkaHttpBackend.usingActorSystem(actorSystem, customHttpsContext =
 
 For more information refer to [akka docs](https://doc.akka.io/docs/akka-http/current/client-side/client-https-support.html).
 
-## Using Async-http-client
```
adamw, line 105: and here :)

docs/backends/monix.md
```diff
@@ -34,48 +34,6 @@ Host header override is supported in environments running Java 12 onwards, but i
...
-
-To use, add the following dependency to your project:
-
-```scala
-"com.softwaremill.sttp.client3" %% "async-http-client-backend-monix" % "@VERSION@"
-```
-           
-This backend depends on [async-http-client](https://github.com/AsyncHttpClient/async-http-client), uses [Netty](http://netty.io) behind the scenes.
-
-Next you'll need to define a backend instance as an implicit value. This can be done in two basic ways:
```
adamw, line 47: with the exception that the implicit requirement is out of date, maybe we can move this section (describing the Task/Resource difference) up to the HttpClient one, which is much more succint; and this info might be useful to newcomers

docs/backends/fs2.md
```diff
@@ -55,86 +55,6 @@ Host header override is supported in environments running Java 12 onwards, but i
...
-"com.softwaremill.sttp.client3" %% "async-http-client-backend-fs2" % "@VERSION@" // for cats-effect 3.x & fs2 3.x
-// or
-"com.softwaremill.sttp.client3" %% "async-http-client-backend-fs2-ce2" % "@VERSION@" // for cats-effect 2.x & fs2 2.x
-```
- 
-This backend depends on [async-http-client](https://github.com/AsyncHttpClient/async-http-client) and uses [Netty](http://netty.io) behind the scenes.
-
-Next you'll need to define a backend instance as an implicit value. This can be done in two basic ways:
-
-* by creating a `Resource`, which will instantiate the backend (along with a `Dispatcher`) and close it after it has been used
```
adamw, line 72: same here - maybe move to the http client section

docs/backends/catseffect.md
```diff
@@ -1,82 +1,14 @@
...
-"com.softwaremill.sttp.client3" %% "async-http-client-backend-cats" % "@VERSION@" // for cats-effect 3.x
-// or
-"com.softwaremill.sttp.client3" %% "async-http-client-backend-cats-ce2" % "@VERSION@" // for cats-effect 2.x
-```
-           
-This backend depends on [async-http-client](https://github.com/AsyncHttpClient/async-http-client), uses [Netty](http://netty.io) behind the scenes. 
-
-Alternatively, the [http4s](http4s.md) backend can also be created for a type implementing the cats-effect's `Async` typeclass, and supports streaming as in [fs2](fs2.md).
-
-Next you'll need to define a backend instance. This can be done in two basic ways:
```
adamw, line 21: and here :)

async-http-client-backend/cats-ce2/src/main/scala/sttp/client3/asynchttpclient/cats/AsyncHttpClientCatsBackend.scala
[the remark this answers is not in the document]
rafalambrozewicz: Good catch, moved annotations to companion objects 👍

docs/websockets.md
rafalambrozewicz: Moved at the very bottom with a depreciated message.

docs/resilience.md
rafalambrozewicz: updated

docs/backends/summary.md
```diff
@@ -30,12 +30,6 @@ Class                                Effect type                      Supported
 ``HttpURLConnectionBackend``         None (``Identity``)              n/a                                               no                         no
 ``TryHttpURLConnectionBackend``      ``scala.util.Try``               n/a                                               no                         no
 ``AkkaHttpBackend``                  ``scala.concurrent.Future``      ``akka.stream.scaladsl.Source[ByteString, Any]``  yes (regular & streaming)  yes
-``AsyncHttpClientFutureBackend``     ``scala.concurrent.Future``      n/a                                               yes (regular)              no
```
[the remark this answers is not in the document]
rafalambrozewicz: Moved to a separate table with "depreciated" information.

docs/backends/monix.md
rafalambrozewicz: Moved to "HttpClient" section.

caoiltecs:
Could you be more specific about what the major vulnerabilities with AsyncHttpClient are?

adamw:
@caoiltecs See: https://github.com/softwaremill/sttp/issues/1571

PR #1602, opened 2022-10-20
Changed to customRequest

### LINKED ISSUE #1601: DELETE Method not working on Scala Native Curl backend
This line:
https://github.com/softwaremill/sttp/blob/b387e2f10571e49874a2a8c466537620ef639eaf/core/src/main/scalanative/sttp/client3/AbstractCurlBackend.scala#L144

Shouldn't be: 
case Method.DELETE     => handle.option(CustomRequest, "DELETE")

As suggested in:
https://github.com/softwaremill/sttp/blob/b387e2f10571e49874a2a8c466537620ef639eaf/core/src/main/scalanative/sttp/client3/curl/CurlOption.scala#L117-L120

Pask423:
closes #1601


adamw:
Thx :)

PR #1603, opened 2022-10-29
Fix curl backend on macos for scala native

k3rnL:
Fix segfault on macOS STTP native #1481

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,11 +45,11 @@ jobs:
           ~/.ivy2/cache
           ~/.coursier
         key: sbt-cache-${{ runner.os }}-${{ matrix.target-platform }}-${{ hashFiles('project/build.properties') }}
-    - name: Install libidn2-dev libcurl4-dev
+    - name: Install libidn2-dev libcurl4-openssl-dev
       if: matrix.target-platform == 'Native'
       run: |
         sudo apt-get update
-        sudo apt-get install libidn2-dev libcurl4-dev
+        sudo apt-get install libidn2-dev libcurl4-openssl-dev
         echo "STTP_NATIVE=1" >> $GITHUB_ENV
     - name: Compile
       run: |
@@ -129,10 +129,10 @@ jobs:
             ~/.ivy2/cache
             ~/.coursier
           key: sbt-cache-release-${{ runner.os }}-${{ hashFiles('project/build.properties') }}
-      - name: Install libidn2-dev libcurl4-dev
+      - name: Install libidn2-dev libcurl4-openssl-dev
         run: |
           sudo apt-get update
-          sudo apt-get install libidn2-dev libcurl4-dev
+          sudo apt-get install libidn2-dev libcurl4-openssl-dev
       - name: Compile
         run: sbt compile
       - name: Publish artifacts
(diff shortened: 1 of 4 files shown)


Pask423:
closes #1481

Pask423:
@k3rnL could you take a look at native tests, they are failing.

k3rnL:
@Pask423 I've tried to run them both on my macOS and on Linux, on the master branch it never go to the end. I have an infinite loop.

```
...
< 
* Connection #0 to host localhost left intact
*   Trying 127.0.0.1:51823...
* Connected to localhost (127.0.0.1) port 51823 (#0)
> POST /echo HTTP/1.1
Host: localhost:51823
Accept: */*
Transfer-Encoding: chunked
Accept-Encoding: gzip, deflate
Content-Type: application/x-www-form-urlencoded
Expect: 100-continue

* Mark bundle as not supporting multiuse
< HTTP/1.1 100 Continue
< Server: akka-http/10.2.10
< Date: Wed, 02 Nov 2022 20:39:32 GMT
^C
[warn] Canceling execution...
[info] CurlBackendHttpTest:
[info] follow redirects
[info] - keep a single history entry of redirect responses
[info] - keep whole history of redirect responses
[info] - break redirect loops
[info] - break redirect loops after user-specified count
[info] body
[info] - post a file
[info] download file
[info] - download a binary file using asFile
[info] - download a text file using asFile
[info] multipart
[info] - send a multipart message with a file
[info] parse response
[info] - as string
[info] - as string with utf-8 characters
[info] - as string with mapping using map
[info] - as string with mapping using mapResponse
[info] - as a byte array
[info] - as parameters
[info] - as string with response via metadata
[info] - as error string with response via metadata
[info] - as string, when the content type encoding is in quotes
[info] - as both string and mapped string
[info] parameters
[info] - make a get request with parameters
[info] body
[info] - post a string
[info] - post a byte array
[info] - post an input stream
[info] - post a byte buffer
[info] - post form data
[info] - post form data with special characters
[error] Test runner interrupted by fatal signal 2
[error] Test runner interrupted by fatal signal 2
[warn] Force close java.lang.RuntimeException: Process /home/parallels/IdeaProjects/sttp/core/target/native-2.13/core-test-out finished with non-zero value 130
[warn] Force close java.lang.RuntimeException: Process /home/parallels/IdeaProjects/sttp/core/target/native-2.13/core-test-out finished with non-zero value 130
Cancelled: coreNative/test
[error] Cancelled: coreNative/test
[error] Use 'last' for the full log.
testServer ... finished with exit code 130
```

Do you have the same issue ? What's the best environment to run the tests ?

adamw:
It's failing with `setopt.c:(.text+0x78): undefined reference to `curl_easy_setopt'`, so maybe we need to install a newer curl version in https://github.com/softwaremill/sttp/blob/master/.github/workflows/ci.yml#L52 ?

adamw:
Indeed locally when I run the tests they hang (on master) - something to investigate :)

Pask423:
As for running tests locally I usually run tests per subproject via sbt shell from InteliJ

kciesielski:
Closing in favor of https://github.com/softwaremill/sttp/pull/2082

PR #1611, opened 2022-11-03
Expose request and response in PrometheusBackend metric customizations

carstenlenz:

This PR refers to the conversation #1298.
It (re-)adds the possibility to use request tags in the metrics for responses because those are missing from `Response` and `response.request`.

I did not keep the compatibility as the change removing requests also did not - of course that can be added but sadly not via an overloading of `apply` because of the default parameters in there (will lead to a compile error). That's why I did not add that. It would need a dedicated constructor method but IMO the `apply` after this change should be the default as it's the most flexible one.


Pask423 (approved):

Pask423 (approved):

Pask423:
Thanks :)

carstenlenz:
Thank you - happy I could contribute something :)

PR #1615, opened 2022-11-07
Enable HTTP version changing for HttpClient based backend

Pask423:


core/src/main/scala/sttp/client3/HttpVersion.scala
```diff
@@ -0,0 +1,8 @@
+package sttp.client3
+
+// FIXME: Move to sttp-model ?
+sealed trait HttpVersion
+case object Empty extends HttpVersion
```
adamw, line 5: yeah moving to model once the backends work is done would be a good idea

also I would rename `Empty` to `Default` and move the enumeration values inside a `HttpVersion` object

unfortunately there are no tags on `Response` so I guess we won't be able to add a `.protocol` method there; something for the next major version :)

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -246,6 +246,18 @@ case class RequestT[U[_], T, -R](
...
+
+  // Used as a workaround to keep binary compatibility
+  // TODO: replace with additional parameter in RequestOptions when writing sttp4
+  def httpVersion_1: RequestT[U, T, R] = tag(httpVersionKey, HTTP_1)
+
+  def httpVersion_1_1: RequestT[U, T, R] = tag(httpVersionKey, HTTP_1_1)
+
+  def httpVersion_2: RequestT[U, T, R] = tag(httpVersionKey, HTTP_2)
+
+  def getHttpVersion: HttpVersion = tags.getOrElse(httpVersionKey, Empty).asInstanceOf[HttpVersion]
```
adamw, line 259: The convention is to have simply `.httpVersion` and `.httpVersion(HttpVersion)` so let's stick to that

core/src/main/scala/sttp/client3/RequestT.scala
```diff
@@ -246,6 +246,14 @@ case class RequestT[U[_], T, -R](
 
   def autoDecompressionDisabled: Boolean = tags.getOrElse(disableAutoDecompressionKey, false).asInstanceOf[Boolean]
 
+  private val httpVersionKey = "httpVersion"
+
+  // Used as a workaround to keep binary compatibility
+  // TODO: replace with additional parameter in RequestOptions when writing sttp4
+  def httpVersion(version: HttpVersion): RequestT[U, T, R] = tag(httpVersionKey, version)
```
adamw, line 253: it would be good to add some comment on what this does, and that this is not uniformly implemented in all backends (that is, some backends support some versinos, some - none)

http4s-backend/src/main/scala/sttp/client3/http4s/Http4sBackend.scala
```diff
@@ -36,12 +36,14 @@ class Http4sBackend[F[_]: Async](
   override def send[T, R >: PE](r: Request[T, R]): F[Response[T]] =
     adjustExceptions(r) {
       val (entity, extraHeaders) = bodyToHttp4s(r, r.body)
+      val version = r.httpVersion.map(versionToHttp4s).getOrElse(http4s.HttpVersion.`HTTP/1.1`)
```
adamw, line 39: can we not set the version if its not explicitly set?

http4s-ce2-backend/src/main/scala/sttp/client3/http4s/Http4sBackend.scala
```diff
@@ -37,12 +37,14 @@ class Http4sBackend[F[_]: ConcurrentEffect: ContextShift](
   override def send[T, R >: PE](r: Request[T, R]): F[Response[T]] =
     adjustExceptions(r) {
       val (entity, extraHeaders) = bodyToHttp4s(r, r.body)
+      val version = r.httpVersion.map(versionToHttp4s).getOrElse(http4s.HttpVersion.`HTTP/1.1`)
```
adamw, line 40: same here

http4s-backend/src/main/scala/sttp/client3/http4s/Http4sBackend.scala
Pask423: It defaults to http4s.HttpVersion.`HTTP/1.1` anyway as far as i have checked

adamw: well our contract is that we force the setting if present, otherwise we leave the default; if http4s at some point changes defaults, our code would become broken

PR #1619, opened 2022-11-08
Add .resourceUsingClient to HttpClient backends.

kamilkloch:
Rationale:
I am trying to migrate away from the deprecated `AsyncHttpClientFs2Backend` and am struggling to reproduce the functionality with https://github.com/softwaremill/sttp/blob/b387e2f10571e49874a2a8c466537620ef639eaf/effects/fs2/src/main/scalajvm/sttp/client3/httpclient/fs2/HttpClientFs2Backend.scala#L96. 
`HttpClientFs2Backend.resource` has a default http client baked in, `HttpClientFs2Backend.usingClient` has hard-coded flag `closeClient = false`. I am missing something akin to `HttpClientFs2Backend.resourceUsingClient`, which would allow me to customize the client and at the same time get back the resource.

### DIFF
--- a/effects/fs2-ce2/src/main/scalajvm/sttp/client3/httpclient/fs2/HttpClientFs2Backend.scala
+++ b/effects/fs2-ce2/src/main/scalajvm/sttp/client3/httpclient/fs2/HttpClientFs2Backend.scala
@@ -128,6 +128,16 @@ object HttpClientFs2Backend {
   ): Resource[F, SttpBackend[F, Fs2Streams[F] with WebSockets]] =
     Resource.make(apply(blocker, options, customizeRequest, customEncodingHandler))(_.close())
 
+  def resourceUsingClient[F[_]: ConcurrentEffect: ContextShift](
+      client: HttpClient,
+      blocker: Blocker,
+      customizeRequest: HttpRequest => HttpRequest = identity,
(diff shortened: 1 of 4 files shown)


adamw:
Sure, sounds good :) Maybe we could add similar methods to other HttpClient backend variants (where it makes sense)

kamilkloch:
I added methods to 2 other client variants. Surprisingly, ZIO backend already contains `HttpClientZioBackend.layerUsingClient`, but its implementation calls `usingClient`, which sets `closeClient` flag to `false`. Is that the desired behavior? (`HttpClientZioBackend.layer` sets the flag to `true`).

adamw:
@kamilkloch Yes, in ZIO the equivalent would be `scopedUsingClient`. A layer is just a description of how to create an object, lifted into ZIO-land. `Scope` is ZIO's `Resource`, where closing happens.

kamilkloch:
@adamw I understand. However, the flag `closeClient` is flipped (`Scope` vs `Resource`). Why is that?

adamw:
Thank you :)

kamilkloch:
Do we want to update the documentstion as well, or leave out this flavor?

adamw:
Ah the docs - yes of course :)

PR #1625, opened 2022-11-13
Attempt to implement CE backend

Pask423:

### DIFF
--- a/effects/cats/src/test/scalajvm/sttp/client3/impl/cats/HttpClientCatsHttpTest.scala
+++ b/effects/cats/src/test/scalajvm/sttp/client3/impl/cats/HttpClientCatsHttpTest.scala
@@ -1,0 +1,10 @@
+package sttp.client3.impl.cats
+
+import cats.effect.IO
+import sttp.client3.testing.HttpTest
+
+class HttpClientCatsHttpTest extends HttpTest[IO] with HttpClientCatsTestBase {
+  override def supportsHostHeaderOverride = false
+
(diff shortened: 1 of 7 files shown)


effects/cats/src/test/scalajvm/sttp/client3/impl/cats/HttpClientCatsHttpTest.scala
```diff
@@ -0,0 +1,10 @@
+package sttp.client3.impl.cats
+
+import cats.effect.IO
+import sttp.client3.testing.HttpTest
+
+class HttpClientCatsHttpTest extends HttpTest[IO] with HttpClientCatsTestBase {
+  override def supportsHostHeaderOverride = false
+
+  override def supportsDeflateWrapperChecking = false
```
adamw, line 9: it seems that's `true` for `HttpClientZioHttpTest`, maybe we can flip it here as well?

adamw (approved):

adamw:
Thanks - docs and should be done :)

effects/cats/src/test/scalajvm/sttp/client3/impl/cats/HttpClientCatsHttpTest.scala
Pask423: Zio have somewhat different impl of streaming this one is based on HttpClientFuture which also have this flags set to false

adamw:
Thanks!

PR #1628, opened 2022-11-14
Absorb exceptions while mapping responses with impure functions in backend stub

trobert:

In #1552, it was ensured that throwing functions used in mapped responses were correctly handled in effect systems with an error channel (namely, zio)

This pr replicates this behavior in `SttpBackendStub`, otherwise the stub does not behave as real backends.

### DIFF
--- a/async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
+++ b/async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
@@ -3,10 +3,9 @@ package sttp.client3.asynchttpclient.zio
 import org.scalatest.concurrent.ScalaFutures
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
-import sttp.capabilities.Streams
 import sttp.client3._
 import sttp.client3.impl.zio._
-import sttp.client3.testing.SttpBackendStub
+import sttp.client3.testing.{SttpBackendStub, TestStreams}
 import sttp.model.Method
 import zio._
 import zio.stream.ZStream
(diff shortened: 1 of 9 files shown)


async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -73,4 +74,45 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
...
+      .response(asStringAlways.map[Int](_ => throw error))
+      .send(backend)
+
+    runtime.unsafeRun(r.either) match {
+      case Left(_: IllegalStateException) => succeed
+      case _                              => fail(s"Should be a failure: $r")
+    }
+  }
+
+  trait TestStreams extends Streams[TestStreams] {
```
Pask423, line 95: there is already TestStreams in SttpBackendStubTests maybe import is enought ?

async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
```diff
@@ -73,4 +74,45 @@ class SttpBackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFuture
 
     runtime.unsafeRun(effect).map(_.body).toList shouldBe List(Right("a"), Right("b"), Right("c"), Right("a"))
   }
+
+  it should "lift errors due to mapping with impure functions into the response monad" in {
```
Pask423, line 78: I think that we have test that alredy cover this case in HttpTest. Are they not working ?
What is the difference between them and this new tests - besides using exact monad not generic one ?

Pask423 (changes requested):

async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
trobert: These new tests validate `SttpBackendStub`. HttpTests are not run against the stub backend which has its own set of tests.

async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
trobert: not possible because it's defined as an inner class

async-http-client-backend/zio1/src/test/scala/sttp/client3/asynchttpclient/zio/SttpBackendStubZioTests.scala
Pask423: Ok so just this comment up

Pask423 (approved):

PR #1629, opened 2022-11-14
Add .sbtopts.

kamilkloch:
Without it, `sbt compile` occasionally experiences OOM.


Pask423:
Sorry, unfortunately this has to be done on your machine as in this way it will interfere with our CI process - which may result in CI failing with 137 exit code.
image
You can set it via InteliJ setting 
image

kamilkloch:
I do have it in IntelliJ, but not in the command-line sbt...
https://github.com/softwaremill/tapir does have `.sbtopts` ;)

Pask423:
Tapir has somewhat different CI setup then sttp but let's try with options like this and see what how our CI behaves
```
-J-Xmx3000M
-J-Xss2M
-Dsbt.task.timings=true
```

kamilkloch:
> BTW I would recommend using sbt shell from InteliJ idea instead of normal command line sbt
And what about VSCode users? :)

Pask423:
Hmm I have never used vsc maybe there is something like sbt shell.

PR #1631, opened 2022-11-14
Changes in CI and build.sbt

Pask423:


adamw:
Great! I wanted to do that for a long time. This follows how things are done in tapir, right?

One thing missing is mergify configuration so that merging steward PR works

Pask423:
Yes, exactly. Sorry I forgot about .mergify config.

build.sbt
```diff
@@ -232,26 +240,56 @@ lazy val allAggregates = projectsWithOptionalNative ++
...
+
 lazy val rootProject = (project in file("."))
   .settings(commonSettings: _*)
   .settings(
     publish / skip := true,
     name := "sttp",
     testJVM := (Test / test).all(filterProject(p => !p.contains("JS") && !p.contains("Native"))).value,
-    testJS1 := (Test / test)
-      .all(filterProject(p => p.contains("JS") && (p.contains("2_11") || p.contains("2_12"))))
+    testJS2_11 := (Test / test)
```
adamw, line 272: how does this differ from `testScoped`?

Pask423: JS tests are run in different way then others there is a comment in ci file, it works the same way in tapir

adamw: Hm which comment?

PR #1649, opened 2022-11-29
Fix parsing of Duration.Inf timeouts

### LINKED ISSUE #1643: HttpClientBackend::convertRequest can raise an unexpected exception
Hi,

It seems that `HttpClientBackend::convertRequest` can raise an unexpected exception if the `readTimeout` is infinite.

Considering [this](https://github.com/softwaremill/sttp/blob/3a340f1f369ede2a61ac66697ddcc4f2b032f3a4/core/src/main/scalajvm/sttp/client3/HttpClientBackend.scala#L51) code:

```scala
builder.timeout(JDuration.ofMillis(request.options.readTimeout.toMillis)).build()
```

that converts `readTimeout` to milliseconds and the fact that `readTimeout` has the type `scala.concurrent.duration.Duration` that can be *infinite*, the operation `readTimeout.toMillis` can throw an exception if the duration is Infinite.

It might be worth either to prohibit infinite timeouts and have only `FiniteDuration` and consider that there might be an exception and handle it gracefully (using `monad.fromTry` ?).

Pask423:
Closes #1643

### DIFF
--- a/armeria-backend/src/main/scala/sttp/client3/armeria/AbstractArmeriaBackend.scala
+++ b/armeria-backend/src/main/scala/sttp/client3/armeria/AbstractArmeriaBackend.scala
@@ -99,7 +99,15 @@ abstract class AbstractArmeriaBackend[F[_], S <: Streams[S]](
       .disablePathParams()
       .method(methodToArmeria(request.method))
       .path(request.uri.toString())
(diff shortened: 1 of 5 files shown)


adamw:
Thanks!

PR #1651, opened 2022-12-04
draft/WIP: Prometheus addition

kevinmeredith:


observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
```diff
@@ -41,13 +42,17 @@ object PrometheusBackend {
...
         RequestListener.lift(
           new PrometheusListener(
-            requestToHistogramNameMapper,
-            requestToInProgressGaugeNameMapper,
-            (rr: (Request[_, _], Response[_])) => responseToSuccessCounterMapper(rr._1, rr._2),
-            (rr: (Request[_, _], Response[_])) => responseToErrorCounterMapper(rr._1, rr._2),
-            (r: (Request[_, _], Throwable)) => requestToFailureCounterMapper(r._1, r._2),
-            requestToSizeSummaryMapper,
-            (rr: (Request[_, _], Response[_])) => responseToSizeSummaryMapper(rr._1, rr._2),
+            (req: Request[_, _]) => requestToHistogramNameMapper(req).map(addLabelPairs(_, req, None)),
```
adamw, line 45: as I understand, this will *always* add the labels (if they are not present). So it's not possible not to add them - if somebody, for some reason would want that.

Maybe, instead we could modify the default values of parameters in `apply` so that the user has chance to fully override the behavior?

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
```diff
@@ -60,6 +65,54 @@ object PrometheusBackend {
...
+    *   user-supplied PrometheusBackend base collector
+    * @param req
+    *   Request
+    * @param maybeResp
+    *   Optional response
+    * @return
+    *   BaseCollectorConfig sub-type. Note that PrometheusBackend#apply and PrometheusListener's constructor reference
+    *   the sub-types of BaseCollectorConfig.
+    */
+  def addLabelPairs(config: BaseCollectorConfig, req: Request[_, _], maybeResp: Option[Response[_]]): config.T = {
```
adamw, line 80: Maybe a better name would be sth like `addMethodStatusLabels`? It would reflect more clearly what it does

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
```diff
@@ -90,7 +143,7 @@ object PrometheusBackend {
       cache.getOrElseUpdate(collectorRegistry, new ConcurrentHashMap[String, T]())
     }
 
-  type RequestCollectors = (Option[Histogram.Timer], Option[Gauge.Child])
+  final class RequestCollectors(val maybeTimer: Option[Histogram.Timer], val maybeGauge: Option[Gauge.Child])
```
adamw, line 146: maybe a `case class`? also, the optionality is already expressed in the types, so I'd just use `timer` / `gauge`

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
adamw, line 80: also, maybe a different signature, without the type member: `def addLabelPairs[T <: BaseCollectorConfig](cfg: T, ...)`?

adamw:
Thanks! I left some comments inline. We should probably make similar changes to the opentelemetry backend (either here or in a separate follow-up issue)

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Good call - will add.

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Side comment: I tend to avoid `case` if there's no need for equality, i.e. `class` w/ `val`'s meets the bare minimum. 

But, adding `case` is a trivial addition, so I'm happy to add.

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Yep - let me add that. `config.T` is a bit ugly, I agree.

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Done - https://github.com/softwaremill/sttp/pull/1651/commits/ec2ca4a498ea0dcc852c704541e45ed824eefaf7

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Done - https://github.com/softwaremill/sttp/pull/1651/commits/d802a5d4fd2d74fc4a70d222488c2317e95c3f80

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Done - https://github.com/softwaremill/sttp/pull/1651/commits/7870222104e55bde081e9d703608568a124ff04d

observability/prometheus-backend/src/main/scala/sttp/client3/prometheus/PrometheusBackend.scala
kevinmeredith: Done - https://github.com/softwaremill/sttp/pull/1651/commits/2e9728638835a51cb8fef88170bcc0f0b10d57ed. Not sure the `T#T` is any better. Let me know, please.

kevinmeredith:
A few tests are still failing locally for me:

```scala
[info] PrometheusBackendTest:
[info] prometheus
[info] - should use default histogram name (120 milliseconds)
[info] - should allow creating two prometheus backends (2 milliseconds)
[info] - should use mapped request to histogram name (4 milliseconds)
[info] - should use mapped request to histogram name with labels and buckets (5 milliseconds)
[info] - should use mapped request to gauge name with labels (6 milliseconds)
[info] - should disable histograms (1 millisecond)
[info] - should use default gauge name *** FAILED *** (15 seconds, 111 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 107 times over 15.070412125 seconds. Last failure message: The Option on which value was invoked was not defined.. (PrometheusBackendTest.scala:165)
[info] - should use mapped request to gauge name *** FAILED *** (15 seconds, 95 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 106 times over 15.081156 seconds. Last failure message: The Option on which value was invoked was not defined.. (PrometheusBackendTest.scala:196)
[info] - should disable gauge *** FAILED *** (15 seconds, 94 milliseconds)
[info]   The code passed to eventually never returned normally. Attempted 107 times over 15.075386708 seconds. Last failure message: The Option on which value was invoked was not defined.. (PrometheusBackendTest.scala:228)
[info] - should use default counter name (46 milliseconds)
[info] - should not override user-supplied 'method' and 'status' labels (9 milliseconds)
[info] - should use default summary name (14 milliseconds)
[info] - should use error counter when http error is thrown (23 milliseconds)
[info] - should use failure counter when other exception is thrown (4 milliseconds)
[info] - should use success counter on success response (1 millisecond)
```

I'm suspicious that they are flaky, but I'll review again another day.

kevinmeredith:
If anything stands out to you, @adamw, on the cause(s) of the failures on https://github.com/softwaremill/sttp/pull/1651#issuecomment-1374714854, let me know, please! Thanks

adamw:
@kevinmeredith thanks!

PR #1658, opened 2022-12-07
Armeria retry

### LINKED ISSUE #1616: Armeria backend returns previous failure result on retry
This behavior showed up in a commercial project, when we switched to Cats Effect 3. We replaced `async-http-client-backend-cats-ce2` with `armeria-backend-cats`, and some of our tests started to fail. It seems that recursive `task.recoverWith` calls keep getting the same HTTP error and eventually exhaust all retries, but the server receives only one request.

After switching from Armeria to FS2 IO backend our tests work again as expected. Maybe Armeria itself is somehow caching responses?

I created an example test showcasing this issue:  https://github.com/kciesielski/sttp-armeria-showcase

In `SttpRetrySpec` switch between `newBackend` and `newBackend2` and notice the difference.

Pask423:
Closes #1616 
In my opinion the problem was that a part of send in `AbstractArmeriaBackend.scala` was not wrapped by F context and thus was executing only once for all the retries.

### DIFF
--- a/armeria-backend/cats/src/test/scala/sttp/client3/armeria/cats/ArmeriaCatsHttpTest.scala
+++ b/armeria-backend/cats/src/test/scala/sttp/client3/armeria/cats/ArmeriaCatsHttpTest.scala
@@ -2,15 +2,10 @@ package sttp.client3.armeria.cats
 
 import cats.effect.IO
 import sttp.client3._
-import sttp.client3.impl.cats.CatsTestBase
+import sttp.client3.impl.cats.{CatsRetryTest, CatsTestBase}
 import sttp.client3.testing.HttpTest
-import sttp.client3.testing.HttpTest.endpoint
-import sttp.model.StatusCode
 
-import scala.concurrent.duration.{DurationInt, FiniteDuration}
-import scala.util.Random
-
-class ArmeriaCatsHttpTest extends HttpTest[IO] with CatsTestBase {
+class ArmeriaCatsHttpTest extends HttpTest[IO] with CatsRetryTest with CatsTestBase {
   override val backend: SttpBackend[IO, Any] = ArmeriaCatsBackend[IO]()
 
   "illegal url exceptions" - {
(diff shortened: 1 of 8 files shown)


armeria-backend/cats/src/test/scala/sttp/client3/armeria/cats/ArmeriaCatsHttpTest.scala
```diff
@@ -16,6 +26,66 @@ class ArmeriaCatsHttpTest extends HttpTest[IO] with CatsTestBase {
     }
   }
 
+  val port = 8080
```
Pask423, line 29: Maybe it will be beneficial to make this test usable in all backends

armeria-backend/src/main/scala/sttp/client3/armeria/AbstractArmeriaBackend.scala
```diff
@@ -59,7 +59,7 @@ abstract class AbstractArmeriaBackend[F[_], S <: Streams[S]](
   override def responseMonad: MonadError[F] = monad
 
   override def send[T, R >: SE](request: Request[T, R]): F[Response[T]] =
-    adjustExceptions(request)(execute(request))
+    monad.unit(request).flatMap(r => adjustExceptions(request)(execute(r)))
```
adamw, line 62: ah, good catch! though for this purpose there's `monad.suspend`

armeria-backend/cats/src/test/scala/sttp/client3/armeria/cats/ArmeriaCatsHttpTest.scala
```diff
@@ -16,6 +26,66 @@ class ArmeriaCatsHttpTest extends HttpTest[IO] with CatsTestBase {
...
+      )
+      .respond(
+        response()
+          .withStatusCode(code)
+      )
+  }
+
+  "retry test" - {
+    "should call the HTTP server twice" in {
+      var mockServer: ClientAndServer = null
```
adamw, line 63: hm maybe we can do the same with our stub backend?

armeria-backend/src/main/scala/sttp/client3/armeria/AbstractArmeriaBackend.scala
kciesielski: 🎖️

PR #1665, opened 2022-12-16
Fix jsoniter coordinates

tabdulradi:
As per https://repo1.maven.org/maven2/com/softwaremill/sttp/client3/jsoniter_3/3.8.5/


adamw:
Indeed ... thanks! :)

PR #1667, opened 2022-12-19
docs: fix simple client name in quickstart

francisdb:
Will `generated-docs/out/quickstart.md` be updated automatically?


adamw:
Thanks! Yes, the `out/...` is regenerated on each release from the sources.

PR #1671, opened 2022-12-24
Use effect runtimes' executors for JDK client

ybasket:
The JDK HttpClient will create its own cached thread pool if no executor is given explicitly, leading to more threads that contend for scheduling. As it performs no thread blocking, it should be more efficient to use the compute pool exposed by the effect runtime (CE, Monix, ZIO, Scala global).

Prior art in http4s-jdk-client: https://github.com/http4s/http4s-jdk-http-client/pull/641

In-depth explanation of why additional thread pools are problematic: https://github.com/typelevel/cats-effect/blob/26630c04e60afe7bba43bb1506ecfc4f773ad927/docs/core/starvation-and-tuning.md


adamw:
Thanks - definitely an important change :)

Issue #1707, opened 2023-01-19
Sttp starts a ForkJoinPool.commonPool-worker - can you explain why?


don41382:
Hi guys,

Thanks for the awesome library. I just have a quick question. I noticed when sending a request, sttp starts a `ForkJoinPool.commonPool-worker` and I don't understand why.

I thought by passing in an ExecutionContext it would only use that one.

Here is the sample program:
```
import sttp.client3.{HttpClientFutureBackend, UriContext, basicRequest}

import java.util.concurrent.Executors
import scala.concurrent.{Await, ExecutionContext}
import scala.concurrent.duration.DurationInt

object SttpTest extends App {

  val ec = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor())
  val backend = HttpClientFutureBackend()(ec)
  val result = basicRequest.get(uri"https://www.google.de").send(backend)

  Await.result(result, 10.seconds)
}
```

VisualVM View
<img width="620" alt="image" src="https://user-images.githubusercontent.com/1768764/213392806-9c64f52f-2147-4d54-9f8b-50b11008bb8c.png">

I am using the latest version `3.8.8` with Scala `2.13.10`. I would love to understand, why this is happening / needed. 

Thanks in advance!

adamw:
This thread seems to be always started when using HttpClient: https://github.com/openjdk/jdk/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L507-L531

As far as I understand, that's the thread that waits for the socket which has data to read, and then any processing is delegated to the provided executor.

don41382:
@adamw Understood. But what about the `ForkJoinPool.commonPool-woker`. Do you know where this pool is coming from?

adamw:
@don41382 that was more tricky, there are in fact two reasons:

1. https://github.com/openjdk/jdk/blob/master/src/java.net.http/share/classes/jdk/internal/net/http/HttpClientImpl.java#L990-L995

Not sure why this `if` exists, but it causes the default FJ pool to initialise (and run a dummy operation). Not sure if there's a way to escape it - the `exchangeExecutor` is not `null`, as we're setting it to our executor explicitly

2. in sttp itself: https://github.com/softwaremill/sttp/blob/2a1a478bac29e4a9bdb57369e0b62653d1c02b88/core/src/main/scalajvm/sttp/client3/HttpClientAsyncBackend.scala#L44

[...]

PR #1712, opened 2023-01-26
Optimize the way HttpClient-based backends read body as an input stream

adamw:
Use the given executor for callbacks, if provided.

Related to #1707


adamw:
The MiMA failures are due to changes in package-private code - which shouldn't be used by clients; hence ignoring.

PR #1718, opened 2023-01-31
Fix response.request accessible from PrometheusBackend

Krever:
We had a problem in our codebase that we reported `example.com` to prometheus because we took it from response and not request.

Im not sure if we want to merge it like that. The whole existence of ExampleGet exposed in the `core` looks quite dangerous to me, but I don't feel adventurous enough to try deeper refactor.


adamw:
True, this should be probably more clearly marked as testing utilities to avoid bugs as here. Can you create an issue to fix this in sttp4?

Meanwhile, let's see where else this `apply` is used :)

PR #1742, opened 2023-02-27
http4s: remove blaze dependency

Fristi:
I've set the `http4s-blaze-client` to `Test` as the docs now suggest that _Ember_ is the official client. It shouldn't be too hard to build a `Client[F]` your self. Ember uses keyed-pool which in turn will support otel4s which allows you track close connection pool metrics. As well Ember will provide a scala native version for which support can be added later.

I guess this will break MiMA?


adamw:
Thanks! MiMA only applies to `core`, as any libraries or integrations should not depend on concrete backend implementations.

Maybe we should include an optional dependency on ember, then, providing similar convenience methods to create an ember-backed-backend? Or at least providing a copy-pastable example of what to do to create a backend.

PR #1796, opened 2023-04-07
Add Scala Native build for jsoniter

kyri-petrou:
Resolves #1795 . This PR adds support Scala Native support for the `jsoniter` module. One caveat of this is that `jsoniter-scala` added support for SN in 2.17.4+, and these versions are not published for Scala 2.11. Are we happy to stop publishing Scala 2.11 versions of the jsoniter (JVM) module, or should we use the old version for JVM and the latest for JS / Native?

I'll create a PR to `master` to port these changes to master as well once this one has been approved

PR #1805, opened 2023-04-11
Add Scala Native build for circe

alfabetacain:
Similar to https://github.com/softwaremill/sttp/pull/1796, this adds scala native support for the circe submodule.
As is the case with jsoniter, there is no circe version that supports both scala native and 2.11, so this drops 2.11 support for the circe module.

  - All the `... should throw UnsupportedOperationException with reason` fails and the scala native tests hangs :sweat_smile: , but they also fail locally for the v3 branch, so hoping it's just my setup

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -119,9 +119,8 @@ val testServerSettings = Seq(
   })
 )
 
-val circeVersion: Option[(Long, Long)] => String = {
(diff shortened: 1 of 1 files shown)


adamw:
Thanks!

PR #1806, opened 2023-04-12
Use `-release:11` across for all JVM builds across Scala versions

kyri-petrou:
Followup of discussion on [this PR](https://github.com/softwaremill/sttp/pull/1796#issuecomment-1500779121).

With this PR, the `-release:11` flag is added to all JVM builds for Scala 2.12, 2.13 and 3. Note that I tried to use `-release:8` for Scala 2.12, but the code wouldn't compile because `core` uses Java 11 APIs. More info on making on making the compiler flags consistent across Scala versions [here](https://github.com/lampepfl/dotty/pull/14606)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,6 +45,7 @@ val commonSettings = commonSmlBuildSettings ++ ossPublishSettings ++ Seq(
 )
 
 val commonJvmSettings = commonSettings ++ Seq(
+  scalacOptions += "-release:11",
   Test / testOptions += Tests.Argument("-oD") // add test timings; js build specify other options which conflict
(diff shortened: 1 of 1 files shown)


adamw:
Thank you!

SethTisue:
Does this imply that the Scala Toolkit won't work on Java 8?

adamw:
@SethTisue It might, though it's not really a question to me :) As far as sttp is concerned, we require Java 11+ for quite some time now (probably a year or so)

SethTisue:
@szymon-rd something to document, I guess

PR #1808, opened 2023-04-13
add 'omitAcceptEncoding' (default is false) parameter to toCurl

### LINKED ISSUE #1807: Dropping AcceptEncoding in ToCurlConverter should be parameterised
Browsers do expose the compression either as `-H 'Accept-Encoding: gzip, deflate, br'` (Safari) or as `--compressed` (Chrome) via "Copy as cURL" action.

Hideous behaviour of _ToCurlConverter_ confused me today and I spent some time benchmarking if I do need compression for the service calls I perform. I did that in shell so there was difference and I ending up trying to explicitly set compression in sttp request and only then I figuring that all was done for nothing as compression *was* already enabled but not dumped by `toCurl`.

I suppose that behaviour should be parameterised and turned off by default (or replaced with `--compressed` if `Accept-Encoding` matches default for sttp, results will be still _human-readable_ and compression could be easily turned off in produced cURL snippet (if `--compressed` is the last arg as Chrome does).

aleksandr-vin:
Align behaviour with browsers' as they do not omit this header.

Closes #1807

### DIFF
--- a/core/src/main/scala/sttp/client4/internal/ToCurlConverter.scala
+++ b/core/src/main/scala/sttp/client4/internal/ToCurlConverter.scala
@@ -5,13 +5,20 @@ import sttp.model._
 
 object ToCurlConverter {
 
-  def apply(request: GenericRequest[_, _]): String = apply(request, HeaderNames.SensitiveHeaders)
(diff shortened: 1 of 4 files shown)


adamw:
Thanks :)

PR #1855, opened 2023-06-14
Issue 1853 empty content-encoding

flsh86:


testing/server/src/main/scala/sttp/client3/testing/server/HttpServer.scala
```diff
@@ -488,8 +488,18 @@ private class HttpServer(port: Int, info: String => Unit) extends AutoCloseable
...
         }
+    } ~ path("empty_content_encoding") {
+      get {
+        respondWithHeader(RawHeader("Content-Encoding", "")) {
+          complete("OK")
+        }
+      }
     }
 
+  def includeEmptyContentEncodingHeader: Directive0 = mapResponseHeaders { headers =>
```
adamw, line 499: unused?

okhttp-backend/monix/src/test/scala/sttp/client3/okhttp/monix/OkHttpMonixHttpTest.scala
```diff
@@ -17,6 +17,7 @@ class OkHttpMonixHttpTest extends HttpTest[Task] {
   override implicit val convertToFuture: ConvertToFuture[Task] = convertMonixTaskToFuture
 
   override def supportsDeflateWrapperChecking = false
+  override def supportsEmptyContentEncoding = false
```
adamw, line 20: `false`, but there are some code changes in the backend?

core/src/main/scalajvm/sttp/client3/HttpClientBackend.scala
```diff
@@ -75,9 +77,10 @@ abstract class HttpClientBackend[F[_], S, P, B](
       resBody: Either[B, WebSocket[F]],
       request: Request[T, R]
   ): F[Response[T]] = {
-    val headersMap = res.headers().map().asScala
+    val headersMap = filterEmptyContentEncoding(res.headers().map().asScala)
```
adamw, line 80: I think I'd leave the headers as they are received from the client, but instead move the filtering to our invocations of `customEncodingHandler`. That way the user gets the headers that were sent by the server, and we don't try to decode using an empty encoding

testing/server/src/main/scala/sttp/client3/testing/server/HttpServer.scala
flsh86: Yeah, forgot to remove it. Thanks!

kciesielski (commented):
**Question** Shouldn't this be a general filtering of all empty headers, with `content-encoding` being just an example? Looks like empty values are generally valid from the point of view of RFC, but pointless anyway. https://stackoverflow.com
/a/12131993

adamw:
@wydra98 ^ see the comment about empty header values

@kciesielski I think on the low-level, we should retain whatever is returned from the server; and the `response.headers` is a low-level representation (a multi map string -> string, basically). I agree that when we do any parsing (like when decoding the body), we should filter out empty values

okhttp-backend/monix/src/test/scala/sttp/client3/okhttp/monix/OkHttpMonixHttpTest.scala
flsh86: Yeah, it should not be changed to false. The test for that case was passing but I've got baited with an `OkHttp Dispatcher`  error and thought it might be something regarding the change. I've tested it locally and everything works fine.

core/src/main/scalajvm/sttp/client3/HttpClientBackend.scala
flsh86: Yeah, the required filtering got much simplified by that, thanks a lot for pointing that out

adamw (approved):

PR #1856, opened 2023-06-15
No way to handle invalid header value using http client zio backend

wydra98:
* I am not sure if I should add .catchSomeDefect somewhere else

### DIFF
--- a/effects/zio/src/test/scalajvm/sttp/client4/httpclient/zio/BackendStubZioTests.scala
+++ b/effects/zio/src/test/scalajvm/sttp/client4/httpclient/zio/BackendStubZioTests.scala
@@ -80,7 +80,7 @@ class BackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFutures wi
     }
   }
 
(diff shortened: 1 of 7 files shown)


adamw:
I think the main place this should be added are the proper backends, e.g. `HttpClientZioBackend`. There, you have to override the `send` method and add the catching of defects. Also, I think the `.resurrect` method does exactly this, but in a single line of code.

effects/zio/src/test/scalajvm/sttp/client4/httpclient/zio/BackendStubZioTests.scala
```diff
@@ -79,4 +79,21 @@ class BackendStubZioTests extends AnyFlatSpec with Matchers with ScalaFutures wi
       case _                              => fail(s"Should be a failure: $r")
     }
   }
+
+  it should "throw an exception instead of ZIO defect if the header value is inappropriate" in {
```
kciesielski, line 83:
```suggestion
  it should "throw an exception instead of ZIO defect if the header value is invalid" in {
```

kciesielski (approved):

PR #1879, opened 2023-07-04
1853 ignore content encoding header if it is empty

PanHNE:

### DIFF
--- a/core/src/test/scala/sttp/client4/testing/HttpTest.scala
+++ b/core/src/test/scala/sttp/client4/testing/HttpTest.scala
@@ -57,6 +57,7 @@ trait HttpTest[F[_]]
   protected def supportsCancellation = true
   protected def supportsAutoDecompressionDisabling = true
   protected def supportsDeflateWrapperChecking = true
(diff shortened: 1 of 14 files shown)


kciesielski (approved):

core/src/test/scala/sttp/client4/testing/HttpTest.scala
```diff
@@ -57,6 +57,7 @@ trait HttpTest[F[_]]
   protected def supportsCancellation = true
   protected def supportsAutoDecompressionDisabling = true
   protected def supportsDeflateWrapperChecking = true
+  protected def supportsEmptyContentEncoding = false
```
adamw, line 60: eee ... shouldn't this be `true` by default? As in here https://github.com/softwaremill/sttp/blob/v3/core/src/test/scala/sttp/client3/testing/HttpTest.scala#L60

kciesielski: Hmm seems so.

PR #1889, opened 2023-07-07
add generic type to FetchCatsBackend stub method

MirkoManojlovic:

This PR adds a generic type to stub method in FetchCatsBackend companion object. 
Reasoning for this change is, when you have `SttpBackend` defined as a dependency like this

```scala
 val catsBackend: SttpBackend[IO, capabilities.WebSockets] = FetchCatsBackend[IO]()
```

it is not possible to mock it using this code

```scala
val stubedBackend: SttpBackendStub[IO, Any] = FetchCatsBackend.stub[IO]
```

adding a type resolves this issue 😃 

```scala
val stubedBackend: SttpBackendStub[IO, WebSockets] = FetchCatsBackend.stub[IO, WebSockets]
```

### DIFF
--- a/effects/cats-ce2/src/main/scalajs/sttp/client3/impl/cats/FetchCatsBackend.scala
+++ b/effects/cats-ce2/src/main/scalajs/sttp/client3/impl/cats/FetchCatsBackend.scala
@@ -48,5 +48,5 @@ object FetchCatsBackend {
     *
     * See [[SttpBackendStub]] for details on how to configure stub responses.
     */
-  def stub[F[_]: Concurrent: ContextShift]: SttpBackendStub[F, Any] = SttpBackendStub(new CatsMonadAsyncError)
+  def stub[F[_]: Concurrent: ContextShift, P]: SttpBackendStub[F, P] = SttpBackendStub(new CatsMonadAsyncError)
 }
--- a/effects/cats-ce2/src/main/scalajs/sttp/client3/impl/cats/FetchCatsBackend.scala
+++ b/effects/cats-ce2/src/main/scalajs/sttp/client3/impl/cats/FetchCatsBackend.scala
@@ -48,5 +48,5 @@ object FetchCatsBackend {
     *
     * See [[SttpBackendStub]] for details on how to configure stub responses.
     */
-  def stub[F[_]: Concurrent: ContextShift, P]: SttpBackendStub[F, P] = SttpBackendStub(new CatsMonadAsyncError)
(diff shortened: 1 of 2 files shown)


effects/cats-ce2/src/main/scalajs/sttp/client3/impl/cats/FetchCatsBackend.scala
```diff
@@ -48,5 +48,5 @@ object FetchCatsBackend {
     *
     * See [[SttpBackendStub]] for details on how to configure stub responses.
     */
-  def stub[F[_]: Concurrent: ContextShift]: SttpBackendStub[F, Any] = SttpBackendStub(new CatsMonadAsyncError)
```
adamw, line 51: Thanks for the PR - following suit for other stub backends, I think it would be sufficient to simply type this using the `WebSockets` capability. Then you can upcast to `Any` because of variance. See e.g.: https://github.com/softwaremill/sttp/blob/d0fb7e962ae3efc0b9e808bb273bee199de8f66d/effects/fs2/src/main/scalajvm/sttp/client3/httpclient/fs2/HttpClientFs2Backend.scala#L173

MirkoManojlovic: No no, thank you for the awesome lib, I'm just trying to give back a little 😄 

For the reply, I'm not sure I get it. The docs for the second type parameter in `SttpBackend` say

> P – Capabilities supported by this backend, in addition to sttp.capabilities.Effect. This might be Any (no special capabilities), sttp.capabilities.Streams (the ability to send and receive streaming bodies) or sttp.capabilities.WebSockets (the ability to handle websocket requests).

So should the change be, this? (then we only support `Websockets`? Not sure if `FetchCatsBackend` can support anything else?)
```scala
  def stub[F[_]: Async, WebSockets]: SttpBackendStub[F, WebSockets] = SttpBackendStub(implicitly)
```
or there should ne now change at all 😄

adamw: Yes, it should be changed as you described. In general this should be the maximal set of capabilities - you can always upcast a backend to narrow the supported capabilities list. In case of cats, it's only WebSockets, no streaming is supported.

MirkoManojlovic: Oh okay understood, thank you.

Code changed 😄

adamw:
Thanks! :)

Issue #1766, opened 2023-03-14
Add a WebSocketSyncBackend type


adamw:
The OkHttp backend currently uses such combination

PR #1895, opened 2023-07-12
Http client websocket support

benzwreck:


core/src/main/scalajvm/sttp/client4/DefaultSyncBackend.scala
```diff
@@ -1,16 +1,16 @@
...
-import sttp.client4.testing.SyncBackendStub
+import sttp.client4.testing.WebSocketBackendStub
 
 object DefaultSyncBackend {
 
   /** Creates a default synchronous backend with the given `options`, which is currently based on
     * [[HttpClientSyncBackend]].
     */
-  def apply(options: BackendOptions = BackendOptions.Default): SyncBackend =
+  def apply(options: BackendOptions = BackendOptions.Default): WebSocketBackend[Identity] =
```
adamw, line 11: This should return a dedicated subtype, `WebSocketSyncBackend`; that's what https://github.com/softwaremill/sttp/issues/1766 is about

adamw: (the goal is to expose to the user "clean" signatures, without the `Identity` wrapper)

docs/testing.md
```diff
@@ -230,7 +230,7 @@ It is also possible to create a stub backend which delegates calls to another (p
 
 ```scala mdoc:compile-only
 val testingBackend =
-  SyncBackendStub.withFallback(DefaultSyncBackend())
+  WebSocketBackendStub.withFallback(DefaultSyncBackend())
```
adamw, line 233: I think we can keep `SyncBackendStub` here

core/src/main/scalajvm/sttp/client4/internal/httpclient/WebSocketImpl.scala
```diff
@@ -55,29 +52,61 @@ private[client4] class WebSocketImpl[F[_]](
...
-      Canceler { () =>
-        cf.cancel(true)
-        ()
-      }
+  protected[client4] def fromCompletableFuture(cf: CompletableFuture[JWebSocket]): F[Unit]
+
+}
+
+private[client4] object WebSocketImpl {
+  def sync[F[_]](
```
adamw, line 66: I think we still need a sequencer for the sync version - a sync websocket can still be used from multiple threads. I think using a semaphore as in `FutureSequencer` should work here (but in a sync version, of course)

core/src/main/scalajvm/sttp/client4/internal/httpclient/IdSequencer.scala
```diff
@@ -0,0 +1,19 @@
...
+
+private[client4] class IdSequencer extends Sequencer[Identity] {
+  private val semaphore = new Semaphore(1)
+
+  def apply[T](t: => Identity[T]): Identity[T] = {
+    blocking {
+      semaphore.acquire()
+    }
+    val result = t
+    semaphore.release()
```
adamw, line 16: the release should be in a finally block

core/src/main/scala/sttp/client4/backend.scala
```diff
@@ -52,6 +52,11 @@ trait SyncBackend extends Backend[Identity] {
   override def monad: MonadError[Identity] = IdMonad
 }
 
+/** A [[GenericBackend]] which is synchronous (side effects are run directly), and supports web sockets. */
+trait WebSocketSyncBackend extends SyncBackend with WebSocketBackend[Identity] {
```
adamw, line 56: I think we should have an override for `send` in `WebSocketRequest` with the simplified signature:

```
def send(backend: WebSocketSyncBackend): Response[T] = backend.send(this)
```

examples/src/main/scala/sttp/client4/examples/GetRawResponseBodySynchronous.scala
```diff
@@ -13,7 +13,7 @@ object GetRawResponseBodySynchronous extends App {
     .get(uri"https://httpbin.org/get")
     .response(asBoth(asJson[HttpBinResponse], asStringAlways))
 
-  val backend: SyncBackend = HttpClientSyncBackend()
+  val backend: WebSocketSyncBackend = HttpClientSyncBackend()
```
adamw, line 16: I think in the non-websocket examples we could leave `SyncBackend`, since it's a super-trait? We don't need the ws capability here, so maybe it's simpler when it's typed this way?

core/src/main/scalajvm/sttp/client4/internal/httpclient/WebSocketImpl.scala
```diff
@@ -55,29 +55,58 @@ private[client4] class WebSocketImpl[F[_]](
...
+      ws: JWebSocket,
+      queue: SimpleQueue[F, WebSocketEvent],
+      _isOpen: AtomicBoolean,
+      sequencer: Sequencer[F],
+      _monad: MonadError[F]
+  ): WebSocketImpl[F] =
+    new WebSocketImpl[F](ws, queue, _isOpen, sequencer) {
+      override implicit def monad: MonadError[F] = _monad
+
+      override protected[client4] def fromCompletableFuture(cf: CompletableFuture[JWebSocket]): F[Unit] =
```
adamw, line 79: [minor] maybe we should simply make `fromCompletableFuture` a parameter (of type `CF => F[Unit]`), instead of a method?

core/src/main/scalajvm/sttp/client4/httpclient/HttpClientSyncBackend.scala
```diff
@@ -1,48 +1,103 @@
...
+  }
 
-  override def monad: MonadError[Identity] = IdMonad
+  private def sendWebSocket[T](
+      request: GenericRequest[T, R],
+      queue: SimpleQueue[Identity, WebSocketEvent],
+      sequencer: Sequencer[Identity]
+  ): Identity[Response[T]] = {
+    val isOpen: AtomicBoolean = new AtomicBoolean(false)
+    val responseCell = new ArrayBlockingQueue[Either[Throwable, Future[Response[T]]]](5)
```
adamw, line 75: I wonder why this is 5 ... wouldn't 1 work as well?

docs/conf/redirects.md
```diff
@@ -36,8 +36,8 @@ You can disable the stripping of all sensitive headers using the following code:
 import sttp.client4._
 import sttp.client4.wrappers.{FollowRedirectsBackend, FollowRedirectsConfig}
 
-val myBackend: SyncBackend = DefaultSyncBackend()
-val backend: SyncBackend  = FollowRedirectsBackend(
+val myBackend: WebSocketSyncBackend = DefaultSyncBackend()
```
adamw, line 39: let's use `SyncBackend` here, since we don't need web socket functionality

docs/json.md
```diff
@@ -46,7 +46,7 @@ Response can be parsed into json using `asJson[T]`, provided there's an implicit
 import sttp.client4._
 import sttp.client4.circe._
 
-val backend: SyncBackend = DefaultSyncBackend()
+val backend: WebSocketSyncBackend = DefaultSyncBackend()
```
adamw, line 49: same here

docs/requests/authentication.md
```diff
@@ -32,7 +32,7 @@ In order to add digest authentication support just wrap other backend as follows
 ```scala mdoc:compile-only
 import sttp.client4.wrappers.DigestAuthenticationBackend
 
-val myBackend: SyncBackend = DefaultSyncBackend()
+val myBackend: WebSocketSyncBackend = DefaultSyncBackend()
```
adamw, line 35: and here :)

core/src/main/scalajvm/sttp/client4/internal/httpclient/IdSequencer.scala
```diff
@@ -0,0 +1,19 @@
...
+
+private[client4] class IdSequencer extends Sequencer[Identity] {
+  private val semaphore = new Semaphore(1)
+
+  def apply[T](t: => Identity[T]): Identity[T] =
+    try {
+      blocking {
+        semaphore.acquire()
+      }
+      val result = t
```
adamw, line 16: this can be simplified: the result is not needed

core/src/main/scalajvm/sttp/client4/internal/httpclient/IdSequencer.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import sttp.client4.Identity
+
+import java.util.concurrent.Semaphore
+import scala.concurrent.blocking
+
+private[client4] class IdSequencer extends Sequencer[Identity] {
+  private val semaphore = new Semaphore(1)
+
+  def apply[T](t: => Identity[T]): Identity[T] =
```
adamw, line 11: do we need the `Identity` in types, wouldn't `def apply[T](t: => T): T` work?

core/src/main/scalajvm/sttp/client4/httpclient/HttpClientBackend.scala
```diff
@@ -103,8 +127,46 @@ abstract class HttpClientBackend[F[_], S, P, B](
     monad.map(body)(Response(_, code, "", headers, Nil, request.onlyMetadata))
   }
 
+  protected def createSimpleQueue[T]: F[SimpleQueue[F, T]]
```
adamw, line 130: do these methods need to be in the super-class? are they used here at all?

core/src/main/scalajvm/sttp/client4/httpclient/HttpClientBackend.scala
```diff
@@ -103,8 +127,46 @@ abstract class HttpClientBackend[F[_], S, P, B](
...
+
+  protected def createSequencer: F[Sequencer[F]]
+
   protected def standardEncoding: (B, String) => B
 
+  private[client4] def prepareWebSocketBuilder[T](
```
adamw, line 136: maybe this can be `protected`, as it's only used from subclasses?

core/src/main/scalajvm/sttp/client4/internal/httpclient/WebSocketImpl.scala
```diff
@@ -42,42 +42,27 @@ private[client4] class WebSocketImpl[F[_]](
   override def send(f: WebSocketFrame, isContinuation: Boolean = false): F[Unit] =
     // ws.send* is not thread-safe - at least one can run at a time. Hence, adding a sequencer to ensure that
     // even if called concurrently, these will be run in sequence.
-    sequencer(monad.flatten(monad.eval {
+    sequencer(monad.suspend {
       f match {
         case WebSocketFrame.Text(payload, finalFragment, _) =>
-          fromCompletableFuture(ws.sendText(payload, finalFragment))
+          handleWS(ws.sendText(payload, finalFragment))
```
adamw, line 48: I don't think the name is right here, this method doesn't "handle the websocket" - it just converts the future to the effect, ignoring the result?

core/src/main/scalajvm/sttp/client4/httpclient/HttpClientSyncBackend.scala
```diff
@@ -1,48 +1,114 @@
...
+          isOpen,
+          sequencer,
+          monad,
+          cf =>
+            monad.suspend {
+              Try(cf.get()).fold(monad.error, _ => monad.unit(()))
+            }
+        )
+        val baseResponse = Response((), StatusCode.SwitchingProtocols, "", Nil, Nil, request.onlyMetadata)
+        val body = Future(blocking(bodyFromHttpClient(Right(webSocket), request.response, baseResponse)))
```
adamw, line 97: since we're in a blocking world I think we can simply do `Future.successfull(bodyFrom...)`

core/src/main/scalajvm/sttp/client4/internal/httpclient/IdSequencer.scala
adamw: `t` will be evaluted each time it's referenced in code

so:
```
blocking {
    semaphore.acquire()
}

try t
finally semaphore.release()
```

core/src/main/scalajvm/sttp/client4/httpclient/HttpClientBackend.scala
benzwreck: this is needed in WebSocket implementation and used by all classes in `httpclient` package

adamw: yes, but does it need to be declared in `HttpClientBackend`? Is there any common code that uses this method?

benzwreck: It was declared in `HttpClientAsyncBackend` before because it was only needed in async implementation (`HttpClientFutureBackend`). I thought it would be a good idea to move it to `HttpClientBackend` since it is used in `HttpClientSyncBackend` which implements `HttpClientBackend` and `HttpClientFutureBackend` which implements `HttpClientAsyncBackend` which implements `HttpClientBackend`.
So from my point of view it is not a bad idea to have it here. WDYT?

adamw: from the point of view of the `HttpCLientBackend` class, this method is not used, and hence an implement detail of child classes - so it shouldn't be declared there

benzwreck: I can see your point and it makes sense to me right now. Thanks for the explanation :)

core/src/main/scalajvm/sttp/client4/internal/httpclient/WebSocketImpl.scala
benzwreck: I changed the name to `fromCompletableFutureToEffect`. Couldn't figured out a better one :/

PR #1938, opened 2023-08-31
add structured logging feature

### LINKED ISSUE #1932: Add default structured logging ?
It would be interesting to add a structured logging backend by default (using slf4j's MDC for example). 

Among the things that would be tracked are the uri, the method, the status code... (with the possibility of enabling/disabling entries if required). 

Any comments?

fmeriaux:
Close #1932 

### DIFF
--- a/core/src/main/scala/sttp/client4/logging/LogContext.scala
+++ b/core/src/main/scala/sttp/client4/logging/LogContext.scala
@@ -1,0 +1,53 @@
+package sttp.client4.logging
+
+import sttp.client4.Response
+import sttp.model.{HeaderNames, RequestMetadata}
+
+import scala.collection.mutable
+import scala.concurrent.duration.Duration
+
+trait LogContext {
+  def ofRequest(request: RequestMetadata): Map[String, Any]
+
+  def ofResponse(response: Response[_], duration: Option[Duration]): Map[String, Any]
+}
+
+object LogContext {
+
+  def noop: LogContext = new LogContext {
+    def ofRequest(request: RequestMetadata): Map[String, Any] = Map.empty
+    def ofResponse(response: Response[_], duration: Option[Duration]): Map[String, Any] = Map.empty
+  }
+
+  def default(
+      logRequestHeaders: Boolean = true,
+      logResponseHeaders: Boolean = true,
+      sensitiveHeaders: Set[String] = HeaderNames.SensitiveHeaders
+  ): LogContext = new LogContext {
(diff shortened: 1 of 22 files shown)


adamw:
Sorry not sure how I missed this. Testing logging is difficult, but if you have an idea on how to approach this, go ahead :)

fmeriaux:
We should be able to test `DefaultLog` by making a `SpyLogger` that buffers what we send it. 

I'm also going to separate the context construction into a separate class to test it in isolation and allow users to provide their own implementation of the `context`.

adamw:
It seems that once again the JS builds are broken because of problems with Chrome version / Selenium / some plugin. I'll take a look once I get some time :)

core/src/main/scala/sttp/client4/logging/LogContext.scala
```diff
@@ -0,0 +1,53 @@
+package sttp.client4.logging
+
+import sttp.client4.Response
+import sttp.model.{HeaderNames, RequestMetadata}
+
+import scala.collection.mutable
+import scala.concurrent.duration.Duration
+
+trait LogContext {
+  def ofRequest(request: RequestMetadata): Map[String, Any]
```
adamw, line 10: I'm not a native speaker, but shouldn't this be `forRequest` and `forResponse`?

logging/slf4j/src/main/scala/sttp/client4/logging/slf4j/Slf4jLogger.scala
```diff
@@ -1,29 +1,64 @@
...
+
+        case LogLevel.Warn if underlying.isWarnEnabled =>
+          underlying.warn(message, throwable)
+
+        case LogLevel.Error if underlying.isErrorEnabled =>
+          underlying.error(message, throwable)
+
+        case _ => ()
+      }
+      MDC.clear()
```
adamw, line 62: shouldn't we only clear the keys we've set?

adamw:
This looks great, thanks for your work! I fixed the JS build & left two minor comments on the code

core/src/main/scala/sttp/client4/logging/LogContext.scala
fmeriaux: I'm not a native either, I think the 2 are correct depending on the meaning you want to give.

I'm changing it to `for` because that suits me too.

adamw:
Thanks for the fixes :)

PR #1969, opened 2023-10-11
More flexible upickle integration

### LINKED ISSUE #1821: Investigate why a separate upickle integration is needed when using non-standard serialization
See: https://softwaremill.com/more-than-just-helloworld-json-in-sttp-and-upickle/

adamw:
Closes #1821

### DIFF
--- a/json/upickle/src/main/scala/sttp/client4/upicklejson/SttpUpickleApi.scala
+++ b/json/upickle/src/main/scala/sttp/client4/upicklejson/SttpUpickleApi.scala
@@ -1,29 +1,29 @@
 package sttp.client4.upicklejson
 
-import upickle.default.{read, write, Reader, Writer}
 import sttp.client4._
 import sttp.client4.internal.Utf8
 import sttp.model.MediaType
 import sttp.client4.json._
 
 trait SttpUpickleApi {
+  val upickleApi: upickle.Api
 
-  implicit def upickleBodySerializer[B](implicit encoder: Writer[B]): BodySerializer[B] =
-    b => StringBody(write(b), Utf8, MediaType.ApplicationJson)
+  implicit def upickleBodySerializer[B](implicit encoder: upickleApi.Writer[B]): BodySerializer[B] =
+    b => StringBody(upickleApi.write(b), Utf8, MediaType.ApplicationJson)
 
   /** If the response is successful (2xx), tries to deserialize the body from a string into JSON. Returns:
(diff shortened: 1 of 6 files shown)


adamw:
@szymon-rd what do you think? :)

json/upickle/src/main/scala/sttp/client4/upicklejson/SttpUpickleApi.scala
```diff
@@ -1,29 +1,29 @@
 package sttp.client4.upicklejson
 
-import upickle.default.{read, write, Reader, Writer}
 import sttp.client4._
 import sttp.client4.internal.Utf8
 import sttp.model.MediaType
 import sttp.client4.json._
 
 trait SttpUpickleApi {
+  val upickleApi: upickle.Api
```
kciesielski, line 9: **question:** Can't we just have a default implementation `= upickle.default` here?

adamw: we could but that wouldn't be very helpful. The type of the `upickleApi` needs to be overriden to the singleton type of the value (I'll extend the docs). Otherwise, things don't work as expected, as the compiler can't lookup the correct `Reader`/`Writer` values.

kciesielski: Oh right, the value type has to be overridden by a concrete type. This dependent type management on the library user side is quite confusing.

szymon-rd:
It looks good and works very similarly to the upickle itself. I wonder if it will always type check, given using the path-dependent types in this trait.

adamw:
@szymon-rd that's a good question, but I'm not sure how to verify if that's the case :) In a way, you'ar always using path-dependent types with uPickle (by importing `upickle.default`), so I would say it should work fine.

adamw:
@szymon-rd I'd give this a go, however note that it will require changing imports (in the tutorials etc.) the import from `sttp.client4.json.upickle._` to `sttp.client4.json.upickle.default._`

But that's probably version in the toolkit? (that is, a specific version of the toolkit has its own docs & fixed sttp/upickle versions)

szymon-rd:
@adamw We have one version of tutorials right now, we will add different versions when adoption grows. However, now, it is sufficient to just keep the most recent one.

adamw:
ok, releasing as 4.0.0-M6 :) we can always revert in case of problems

PR #1992, opened 2023-11-07
[v3] Override blocking inside CatsMonadError

### LINKED ISSUE #1991: Missing sttp.monad.MonadError#blocking overriding for cats-effect
Follow-up to a thread at [Discourse](https://softwaremill.community/t/sttp-monad-monaderror-blocking-for-cats-effect/314).

For now, `CatsMonadAsyncError` uses the default implementation of `blocking` (see `sttp.monad.MonadError#blocking`):

```scala
def blocking[T](t: => T): F[T] = eval(t)
```

`eval` is implemented as follows (see `sttp.client3.impl.cats.CatsMonadError#eval`):

```scala
override def eval[T](t: => T): F[T] = F.delay(t)
```

As a result, `Sync[F].delay` is used for blocking actions. That seems to be wrong.

flipp5b:
Fix #1991


flipp5b:
@adamw, I added a test reproducing the issue to the JVM test root. It looks like everything works except JS tests. But I can't figure out why.

adamw:
@flipp5b argh it's something with chrome driver again :/

```
[error] Caused by: org.openqa.selenium.SessionNotCreatedException: session not created: This version of ChromeDriver only supports Chrome version 114
[error] Current browser version is 118.0.5993.117 with binary path /usr/bin/google-chrome
```

I've been fixing this lately, I'll take a look later

adamw:
@mergifyio rebase

adamw:
ah it's for the v3 branch ... I updated its build plugins, let's see
could you submit a PR for `master` later as well?

flipp5b:
Done: https://github.com/softwaremill/sttp/pull/1993.

PR #2012, opened 2023-11-29
Make AbstractCurlBackend async friendly

natsukagami:
Just some structural changes to make implementing an async Curl backend (e.g. with `curl_multi_*`) possible as
a subclass of `AbstractCurlBackend`.
Mainly:
- Make header lists per-request.
- Allow a customizable `CurlHandle -> F[CurlCode]` method, `performCurl`. This can be done with `curl_multi_*`
  instead of just `curl_easy_perform`. Also seems to be the only thing needed to be changed.

We might have to change `curl.option` calls as well, if we want to properly handle them as monadic calls.

### DIFF
--- a/core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
+++ b/core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
@@ -25,35 +25,59 @@ import scala.scalanative.unsigned._
 abstract class AbstractCurlBackend[F[_]](_monad: MonadError[F], verbose: Boolean) extends GenericBackend[F, Any] {
   override implicit def monad: MonadError[F] = _monad
 
+  /** Given a [[CurlHandle]], perform the request and return a [[CurlCode]]. */
+  protected def performCurl(c: CurlHandle): F[CurlCode]
+
+  /** Same as [[performCurl]], but also checks and throws runtime exceptions on bad [[CurlCode]]s. */
+  final def perform(c: CurlHandle) = performCurl(c).flatMap(lift)
+
   type R = Any with Effect[F]
 
   override def close(): F[Unit] = monad.unit(())
 
-  private var headers: CurlList = _
-  private var multiPartHeaders: Seq[CurlList] = Seq()
+  /** A request-specific context, with allocated zones and headers. */
+  private class Context() {
+    val zone: Zone = Zone.open()
+    var headers: CurlList = _
+    var multiPartHeaders: Seq[CurlList] = Seq()
+
+    def close() = {
+      zone.close()
+      if (headers.ptr != null) headers.ptr.free()
+      multiPartHeaders.foreach(_.ptr.free())
+    }
+  }
+
+  private object Context {
+    def apply[T](body: Context => F[T]): F[T] = {
+      implicit val ctx = new Context()
+      body(ctx).ensure(monad.unit(ctx.close()))
+    }
+  }
 
   override def send[T](request: GenericRequest[T, R]): F[Response[T]] =
     adjustExceptions(request) {
-      unsafe.Zone { implicit z =>
+      def perform(implicit ctx: Context): F[Response[T]] = {
+        implicit val z = ctx.zone
         val curl = CurlApi.init
         if (verbose) {
           curl.option(Verbose, parameter = true)
         }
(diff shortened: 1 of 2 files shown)


core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
```diff
@@ -25,35 +25,59 @@ import scala.scalanative.unsigned._
...
-  private var multiPartHeaders: Seq[CurlList] = Seq()
+  /** A request-specific context, with allocated zones and headers. */
+  private class Context() {
+    val zone: Zone = Zone.open()
+    var headers: CurlList = _
+    var multiPartHeaders: Seq[CurlList] = Seq()
+
+    def close() = {
+      zone.close()
+      if (headers.ptr != null) headers.ptr.free()
```
adamw, line 46: maybe `headers` can be an `Option`? That we would avoid using `null` :)

core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
```diff
@@ -25,35 +25,59 @@ import scala.scalanative.unsigned._
...
   type R = Any with Effect[F]
 
   override def close(): F[Unit] = monad.unit(())
 
-  private var headers: CurlList = _
-  private var multiPartHeaders: Seq[CurlList] = Seq()
+  /** A request-specific context, with allocated zones and headers. */
+  private class Context() {
+    val zone: Zone = Zone.open()
+    var headers: CurlList = _
```
adamw, line 41: instead of `var`s, mabye we can prepare the headers/multiPartHeaders parameters in `send`, and then create an (immutable) instance of `Context`? E.g. `case class Context(zone: Zone, headers: Option[CurlList], multiPartHeaders: Seq[CurlList])`

core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
```diff
@@ -62,6 +86,8 @@ abstract class AbstractCurlBackend[F[_]](_monad: MonadError[F], verbose: Boolean
           case None       => handleBase(request, curl, spaces)
         }
       }
+
+      Context(ctx => perform(ctx))
```
adamw, line 90: at first sight I thought that this creates a new context, but I think it runs the given operation using a context in safe way? Maybe instead of `apply`, we could have a `evaluateUsing` method or sth like that

adamw:
(sorry wrong button ;) )

core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
natsukagami: A bit tricky to do, since `setRequestBody` is setting up the multi part headers as it goes through the multipart body. I can on the other hand just exposes a `addHeader` function to create a header that gets cleared when the context is closed.

core/src/main/scalanative/sttp/client4/curl/AbstractCurlBackend.scala
```diff
@@ -25,33 +26,68 @@ import scala.scalanative.unsigned._
...
-      unsafe.Zone { implicit z =>
+      def perform(implicit ctx: Context): F[Response[T]] = {
+        implicit val z = ctx.zone
         val curl = CurlApi.init
         if (verbose) {
           curl.option(Verbose, parameter = true)
         }
         if (request.tags.nonEmpty) {
-          monad.error(new UnsupportedOperationException("Tags are not supported"))
+          return monad.error(new UnsupportedOperationException("Tags are not supported"))
```
adamw, line 79: nice catch! :)

adamw:
Thank you! Do you need a release with these changes, or will you continue working on an async curl backend?

natsukagami:
Thanks! 

> Do you need a release with these changes

Not yet, as I am waiting for scala-native 0.5 and everything else to fall into place ;) 

> will you continue working on an async curl backend

I have a prototype over at https://github.com/natsukagami/sttp/tree/curl-async-gears , but as the branch name suggests, it currently depends on gears stuff and serves mostly as a demonstration that it's possible. I might come back and write a proper backend later :D

Issue #2050, opened 2024-01-10
async-http-client-backend-zioЖ3ю


mev42:
I'm trying to get a response using an http proxy with authorization. The application crashes with the following error:
```
timestamp=2024-01-10T07:00:06.145899Z level=ERROR thread=#zio-fiber-1 message="" cause="Exception in thread "zio-fiber-4" java.lang.NullPointerException: flowPublisher
	at java.base/java.util.Objects.requireNonNull(Objects.java:246)
	at org.reactivestreams.FlowAdapters.toPublisher(FlowAdapters.java:28)
	at sttp.client3.httpclient.zio.HttpClientZioBackend.bodyHandlerBodyToBody(HttpClientZioBackend.scala:55)
	at sttp.client3.httpclient.zio.HttpClientZioBackend.bodyHandlerBodyToBody(HttpClientZioBackend.scala:28)
	at sttp.client3.HttpClientAsyncBackend.$anonfun$sendRegular$3(HttpClientAsyncBackend.scala:57)
	at sttp.client3.HttpClientAsyncBackend.$anonfun$sendRegular$3$adapted(HttpClientAsyncBackend.scala:55)
	at sttp.client3.internal.SttpToJavaConverters$$anon$2.accept(SttpToJavaConverters.scala:12)
	at java.base/java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:859)
	at java.base/java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:837)
	at java.base/java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:478)
	at zio.internal.ZScheduler$$anon$4.run(ZScheduler.scala:478)
	at sttp.client3.impl.zio.RIOMonadAsyncError.error(RIOMonadAsyncError.scala:24)
	at sttp.client3.impl.zio.RIOMonadAsyncError.flatMap(RIOMonadAsyncError.scala:12)
	at sttp.client3.impl.zio.RIOMonadAsyncError.map(RIOMonadAsyncError.scala:9)"
```
Zio code:

```scala
import sttp.client3._
import sttp.client3.httpclient.zio.HttpClientZioBackend
import zio._

object ProxyTest extends ZIOAppDefault {
  override def run: ZIO[Any with ZIOAppArgs with Scope, Any, Any] = {
    HttpClientZioBackend(
      options = SttpBackendOptions.httpProxy(
        "1.2.3.4",
[...]

PR #2088, opened 2024-02-27
Improve proxy authentication preventing NPE and allow Basic Authentication

DybekK:
This PR addresses the problem described in https://github.com/softwaremill/sttp/issues/2050, where requests sent by `HttpClientBackend` to a proxy requiring Basic authentication were failing due to [JDK's default settings](https://www.oracle.com/java/technologies/javase/8u111-relnotes.html), resulting in a 407 response and a null body, leading to errors. By removing `Basic` from `jdk.http.auth.tunneling.disabledSchemes`, we now allow `HttpClient` to use proxy authentication successfully. This behaviour has been documented to guide users on enabling Basic authentication for HTTP tunneling and inform them about security flaws.

Additionally, to avoid NPE from null response bodies, we return empty stream.

### DIFF
--- a/docs/conf/proxy.md
+++ b/docs/conf/proxy.md
@@ -31,6 +31,17 @@ import sttp.client3._
 SttpBackendOptions.httpProxy("some.host", 8080, "username", "password")
 ```
 
+If your backend implements `HttpClientBackend` and your proxy server requires Basic authentication, you will need to enable it by setting
+the `jdk.http.auth.tunneling.disabledSchemes` system property to an empty string (`""`). This can be done by
+calling `System.setProperty("jdk.http.auth.tunneling.disabledSchemes", "")` before creating the backend.
+
+```scala mdoc:compile-only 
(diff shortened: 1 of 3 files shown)


kciesielski:
@DybekK could you add tests for these cases?

docs/conf/proxy.md
```diff
@@ -31,6 +31,16 @@ import sttp.client3._
 SttpBackendOptions.httpProxy("some.host", 8080, "username", "password")
 ```
 
+If your backend implements `HttpClientBackend` and your proxy server requires Basic authentication, you will need to enable it by
```
adamw, line 34: minor correction: if you use a `HttpClient`-based backend (e.g. `HttpClientBackend`)

that is, backends are used, not implemented :)

adamw (approved):

adamw:
@kciesielski I'm not sure if a test using proxies is doable in a reasonable amount of time, but maybe :)

@DybekK we'll also need a PR to master w/ the same changes please :)

PR #2108, opened 2024-03-15
All backends write headers as UTF-8

ghik:
Requires https://github.com/softwaremill/sttp-model/pull/336 and closes https://github.com/softwaremill/sttp-model/issues/317

### DIFF
--- a/async-http-client-backend/src/test/scala/sttp/client3/asynchttpclient/AsyncHttpClientHttpTest.scala
+++ b/async-http-client-backend/src/test/scala/sttp/client3/asynchttpclient/AsyncHttpClientHttpTest.scala
@@ -1,0 +1,7 @@
+package sttp.client3.asynchttpclient
+
+import sttp.client3.testing.HttpTest
+
+abstract class AsyncHttpClientHttpTest[F[_]] extends HttpTest[F] {
+  override protected def supportsNonAsciiHeaderValues = false
+}
--- a/async-http-client-backend/src/test/scala/sttp/client3/asynchttpclient/AsyncHttpClientHttpTest.scala
+++ b/async-http-client-backend/src/test/scala/sttp/client3/asynchttpclient/AsyncHttpClientHttpTest.scala
@@ -3,5 +3,7 @@ package sttp.client3.asynchttpclient
 import sttp.client3.testing.HttpTest
 
 abstract class AsyncHttpClientHttpTest[F[_]] extends HttpTest[F] {
+  override protected def throwsExceptionOnUnsupportedEncoding = false
+  override protected def supportsAutoDecompressionDisabling = false
   override protected def supportsNonAsciiHeaderValues = false
 }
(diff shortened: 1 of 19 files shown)


async-http-client-backend/src/test/scala/sttp/client3/asynchttpclient/AsyncHttpClientHttpTest.scala
```diff
@@ -0,0 +1,7 @@
+package sttp.client3.asynchttpclient
+
+import sttp.client3.testing.HttpTest
+
+abstract class AsyncHttpClientHttpTest[F[_]] extends HttpTest[F] {
+  override protected def supportsNonAsciiHeaderValues = false
+}
```
ghik, line 7: I decided to extract this class because it makes it clear that it's the `AsyncHttpClient` that does not support non-ascii header fields, which would not be apparent if this method was overridden in every `AsyncHttpClient*HttpTest`.

adamw: That's a good idea, but probably there are more such common flags that are set in each test class individually which could be de-duplicated here

ghik: I'll go through them and see what else I can extract.

okhttp-backend/monix/src/test/scala/sttp/client3/okhttp/monix/OkHttpMonixHttpTest.scala
```diff
@@ -18,6 +18,8 @@ class OkHttpMonixHttpTest extends HttpTest[Task] {
 
   override def supportsDeflateWrapperChecking = false
 
+  override def supportsNonAsciiHeaderValues = false
```
ghik, line 21: This is where I'd like to use `OkHttpHttpTest` base class but I can't because there's no `test->test` dependency between modules.

adamw: ah ... sure, here's that fine - and we already have this kind of dep in the async http client tests for example, no?

ghik:
@adamw seems like `sttp-model` dropped support for Scala 2.11 while `sttp` still retains it. What should we do?

adamw:
Ah it's the v3 branch ... let's drop 2.11 here as well, since we got so far ;)

ghik:
Or maybe a more sensible thing to do would be to backport the `sttp-model` changes to the version used in `sttp3`, i.e. 1.5.x, while using 1.7.9 in `master` of `sttp`?

adamw:
I think we'll be maintaining the v3 branch for a while, and maintaining too many forks across projects is just too much work. So if we need to drop 2.11, let's do it :)

ghik:
There would be no forks involved, only a bugfix targeting a slightly earlier version (1.5.x instead of 1.7.x, which is probably what I should have done from the beginning). But if you think it's fine to drop 2.11 for v3 and use newest `sttp-model` then sure, I won't hesitate :)

adamw:
My worry is that this won't be the last fix in sttp-model that we'll want to use in sttp3, so we'll have to maintain two sttp branches (v3 & main), plus two sttp-model branches. I don't want to go that route ;)

adamw (approved):
Looks good :)

PR #2114, opened 2024-03-20
feat: adding full logging options to log settings

Andrapyre:
## Description

There are use cases (e.g. when building a library on top of sttp) where it is useful to pass all logging options as a single parameter to the end user to specify. This PR adds a `.logSettings` method that takes an `Option[LoggingOptions]` parameter, instead of the individual logging options parameters (e.g. `logRequestBody`, etc.).


adamw:
Thank you!

PR #2129, opened 2024-04-12
added support tethys json

chernykhSG:

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -861,6 +861,7 @@ lazy val tethysJson = (projectMatrix in file("json/tethys-json"))
     settings = commonJvmSettings
   )
   .jsPlatform(scalaVersions = scala2 ++ scala3, settings = commonJsSettings)
+  .nativePlatform(scalaVersions = scala2 ++ scala3, settings = commonNativeSettings)
   .dependsOn(core, jsonCommon)
 
 lazy val upickle = (projectMatrix in file("json/upickle"))
--- a/build.sbt
+++ b/build.sbt
@@ -163,7 +163,7 @@ val resilience4jVersion = "2.2.0"
 val http4s_ce2_version = "0.22.15"
 val http4s_ce3_version = "0.23.26"
 
-val openTelemetryVersion = "1.36.0"
+val openTelemetryVersion = "1.37.0"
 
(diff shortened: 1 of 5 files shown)


build.sbt
```diff
@@ -843,6 +843,23 @@ lazy val zio1Json = (projectMatrix in file("json/zio1-json"))
   .jsPlatform(scalaVersions = scala2 ++ scala3, settings = commonJsSettings)
   .dependsOn(core, jsonCommon)
 
+lazy val tethysJson = (projectMatrix in file("json/tethys-json"))
```
adamw, line 846: this should also be added to `allAggregates`

build.sbt
```diff
@@ -843,6 +843,23 @@ lazy val zio1Json = (projectMatrix in file("json/zio1-json"))
...
+  .settings(
+    name := "tethys-json",
+    libraryDependencies ++= Seq(
+      "com.tethys-json" %% "tethys-core" % "0.26.0",
```
adamw, line 850: maybe define a constant in the top of the build file with the version?

build.sbt
```diff
@@ -843,6 +843,23 @@ lazy val zio1Json = (projectMatrix in file("json/zio1-json"))
...
+      "com.tethys-json" %% "tethys-jackson" % "0.26.0",
+      "com.tethys-json" %% "tethys-derivation" % "0.26.0"
+    ),
+    scalaTest
+  )
+  .jvmPlatform(
+    scalaVersions = Seq(scala2_12, scala2_13),
```
adamw, line 857: no scala 3 support?

build.sbt
chernykhSG: There is support, but the new version only supports 3 scala; for 3 scala you need to make a separate implementation.

build.sbt
chernykhSG: There is support, but the new version only supports 3 scala; for 3 scala you need to make a separate implementation.

build.sbt
chernykhSG: corrected

build.sbt
chernykhSG: added support scala3

adamw:
Thanks :)

PR #2135, opened 2024-04-13
fix filtering content-type and content-length headers in HttpClientBa…

ashashev:

## Problems
The java HttpClient prohibits to add `Content-Type` header to request, it's reason HttpClientBacked filters this header. But the check is case-sensitive. So passing `content-type` (in lower case) header occurs IllegalArgumentException.

Another problem is duplicating `content-type` header inside request if the header name passed in lowercase.

These problems might happen if an application gets these headers from an incoming HTTP request and then pass them to a new HTTP request to another service.


adamw:
Thank you!

PR #2145, opened 2024-04-23
Updated Prometheus to 1.2.1

### LINKED ISSUE #2011: Update to Prometheus 1.x.x
Prometheus made [incompatible changes in v1.x.x](https://github.com/prometheus/client_java/releases) and some libraries such as Tapir already migrated to the latest version of Prometheus so it would be helpful to update this project also into latest Prometheus to allow use of Tapir with sttp backend.

Right now, this project still depends on Prometheus v0.x.x [code](https://github.com/softwaremill/sttp/blob/ee57f39d5222aa7abcbae27b7e1ccfbc5e516d17/build.sbt#L918).

Pask423:
closes #2011

### DIFF
--- a/observability/prometheus-backend/src/main/scala/sttp/client4/prometheus/PrometheusBackend.scala
+++ b/observability/prometheus-backend/src/main/scala/sttp/client4/prometheus/PrometheusBackend.scala
@@ -14,6 +14,7 @@ import sttp.model.StatusCode
 import scala.collection.mutable
 
 object PrometheusBackend {
+  // TODO: Refactor metrics names according to Prometheus/OpenTelemetry standards
   val DefaultHistogramName = "sttp_request_latency"
   val DefaultRequestsInProgressGaugeName = "sttp_requests_in_progress"
   val DefaultSuccessCounterName = "sttp_requests_success_count"
@@ -121,15 +122,10 @@ object PrometheusBackend {
   /** Clear cached collectors (gauges and histograms) both from the given collector registry, and from the backend.
     */
   def clear(prometheusRegistry: PrometheusRegistry): Unit = {
-    unregister(prometheusRegistry, histograms)
-    histograms.remove(prometheusRegistry)
-    unregister(prometheusRegistry, gauges)
-    gauges.remove(prometheusRegistry)
-    unregister(prometheusRegistry, counters)
-    counters.remove(prometheusRegistry)
-    unregister(prometheusRegistry, summaries)
-    summaries.remove(prometheusRegistry)
-    ()
+    clear(prometheusRegistry, histograms)
+    clear(prometheusRegistry, gauges)
+    clear(prometheusRegistry, counters)
+    clear(prometheusRegistry, summaries)
   }
 
   /*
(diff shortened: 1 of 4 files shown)


observability/prometheus-backend/src/main/scala/sttp/client4/prometheus/PrometheusBackend.scala
```diff
@@ -132,19 +138,25 @@ object PrometheusBackend {
...
+  private val counters = new mutable.WeakHashMap[PrometheusRegistry, ConcurrentHashMap[String, Counter]]
+  private val summaries = new mutable.WeakHashMap[PrometheusRegistry, ConcurrentHashMap[String, Summary]]
 
   private def cacheFor[T](
-      cache: mutable.WeakHashMap[CollectorRegistry, ConcurrentHashMap[String, T]],
-      collectorRegistry: CollectorRegistry
+      cache: mutable.WeakHashMap[PrometheusRegistry, ConcurrentHashMap[String, T]],
+      prometheusRegistry: PrometheusRegistry
   ): ConcurrentHashMap[String, T] =
     cache.synchronized {
```
adamw, line 156: as a follow-up task, we should remove the `.synchronized`, as it shouldn't be used in Java21+, and repalce it with some kind of a concurrent data structure (or improve the cache in general)

adamw: or maybe we can do it right away, this is invoked only during the backend's construction, so a simple lock should suffice in fact; but we should also protect the cleanup method, then

Pask423: how about separate PR focused on cache I would prefer not to mix it with this update

adamw: ok :)

observability/prometheus-backend/src/main/scala/sttp/client4/prometheus/PrometheusBackend.scala
```diff
@@ -132,19 +138,25 @@ object PrometheusBackend {
   Hence, we need to store a global cache o created histograms/gauges, so that we can properly re-use them.
    */
 
-  private val histograms = new mutable.WeakHashMap[CollectorRegistry, ConcurrentHashMap[String, Histogram]]
-  private val gauges = new mutable.WeakHashMap[CollectorRegistry, ConcurrentHashMap[String, Gauge]]
-  private val counters = new mutable.WeakHashMap[CollectorRegistry, ConcurrentHashMap[String, Counter]]
-  private val summaries = new mutable.WeakHashMap[CollectorRegistry, ConcurrentHashMap[String, Summary]]
+  private def unregister[T <: Collector](prometheusRegistry: PrometheusRegistry, collectors: mutable.WeakHashMap[PrometheusRegistry, ConcurrentHashMap[String, T]]): Unit =
+    collectors
```
kciesielski, line 142: **nitpick:** How about 

```scala
collectors.get(prometheusRegistry).foreach(_.values.forEach(prometheusRegistry.unregister))
```

observability/prometheus-backend/src/test/scala/sttp/client4/prometheus/PrometheusBackendTest.scala
```diff
@@ -419,15 +423,24 @@ class PrometheusBackendTest
...
   }
 
-  private[this] def getMetricValue(name: String): Option[lang.Double] =
-    Option(CollectorRegistry.defaultRegistry.getSampleValue(name))
-
-  private[this] def getMetricValue(name: String, labels: List[(String, String)]): Option[lang.Double] =
-    Option(CollectorRegistry.defaultRegistry.getSampleValue(name, labels.map(_._1).toArray, labels.map(_._2).toArray))
+  private[this] def getMetricSnapshot[T](name: String): Option[T] =
+    Option(PrometheusRegistry.defaultRegistry.scrape((s: String) => s.equals(name)))
+      .filter(_.size() > 0)
```
kciesielski, line 434: **nitpick:** I'm not sure what's the underlying type, but you could possibly replace `_.size() > 0` with `_.nonEmpty`, and `_.get(0)` with `.head`?

kciesielski (commented):
This probably should get backported to the v3 branch after it's merged, right @adamw?

adamw (approved):

adamw:
@kciesielski I don't know, it's a breaking change, although in an integration module ... let's see if somebody will need it ;)

Issue #2106, opened 2024-03-14
Preserve request-metadata in SttpBackendStub


jumale:
## Current behaviour
SttpBackendStub does not include actual metadata of the dispatched request. 
Instead it always includes `sttp.client3.Response.ExampleGet`

```scala
val testingBackend = SttpBackendStub.asynchronousFuture
  .whenRequestMatches(_.uri.path.contains("foo"))
  .thenRespond("success")

val response = basicRequest.post(uri"https://myapp.com/foo").send(testingBackend).futureValue
println(s"${response.request.method} ${response.request.uri}")
```
prints `GET http://example.com`

## Expected behaviour
It prints `POST https://myapp.com/foo`

## Impact
It makes it harder to write functional tests for integrated modules which suppose to handle the RequestMetadata in some way (e.g. error handlers which should capture and obfuscate URI).
Besides that, I think it would be nice have it just for the sake of consistency.

As far as I can see from the code, in order to support it, we only need to change one line:
```scala
class WhenRequest(p: GenericRequest[_, _] => Boolean) {
  ...
  def thenRespond[T](resp: => Response[T]): Self = {
    ...
    case r if p(r) => monad.eval(resp.copy(request = RequestMetadata(r.method, r.uri, r.headers)))
```
Unless there are other reasons I'm not aware of why it shouldn't be done like that.

Pask423:
closed by #2146

PR #2146, opened 2024-04-23
Preserved request metadata in response

### LINKED ISSUE #1719: Make it harder to use ExampleGet and connected Reponse factory methods
Currently `Response` can be easily created with `ExampleGet`Request metadata, as manifested in #1718.

It would be better if the api itself could prevent such bugs (e.g. by moving this into some testing module).

Pask423:
closes #1719 #2106

### DIFF
--- a/core/src/main/scala/sttp/client4/Response.scala
+++ b/core/src/main/scala/sttp/client4/Response.scala
@@ -33,23 +33,8 @@ case class Response[+T](
 
 private[sttp] object Response {
 
-  val emptyGet: RequestMetadata = new RequestMetadata {
-    override def method: Method = Method.GET
-    override def uri: Uri = uri"http://example.com"
-    override def headers: Seq[Header] = Nil
-  }
-
-  def apply[T](body: T, code: StatusCode): Response[T] =
-    Response(body, code, resolveStatusText(code), Nil, Nil, emptyGet)
-
-  def apply[T](body: T, code: StatusCode, statusText: String): Response[T] =
-    Response(body, code, resolveStatusText(code, statusText), Nil, Nil, emptyGet)
-
   def apply[T](body: T, code: StatusCode, requestMetadata: RequestMetadata): Response[T] =
     Response(body, code, resolveStatusText(code), Nil, Nil, requestMetadata)
-
-  def apply[T](body: T, code: StatusCode, statusText: String, requestMetadata: RequestMetadata): Response[T] =
-    Response(body, code, resolveStatusText(code, statusText), Nil, Nil, requestMetadata)
 
   def ok[T](body: T, requestMetadata: RequestMetadata): Response[T] =
     Response(body, StatusCode.Ok, resolveStatusText(StatusCode.Ok), Nil, Nil, requestMetadata)
--- a/core/src/main/scala/sttp/client4/Response.scala
+++ b/core/src/main/scala/sttp/client4/Response.scala
@@ -31,38 +31,28 @@ case class Response[+T](
     s"Response($body,$code,$statusText,${Headers.toStringSafe(headers)},$history,$request)"
 }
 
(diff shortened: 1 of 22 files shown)


core/src/test/scalajvm/sttp/client4/testing/TestResponse.scala
```diff
@@ -0,0 +1,34 @@
+package sttp.client4.testing
+
+import sttp.client4.Response
+import sttp.model._
+
+private[sttp] object TestResponse {
```
adamw, line 6: maybe let's call it `ResponseStub`, keeping the terminology consistent with `BackendStub`? Also, this should be public?

core/src/main/scala/sttp/client4/Response.scala
```diff
@@ -31,38 +31,28 @@ case class Response[+T](
     s"Response($body,$code,$statusText,${Headers.toStringSafe(headers)},$history,$request)"
 }
 
-object Response {
+private[sttp] object Response {
 
-  /** Mainly useful in tests using [[sttp.client4.testing.SttpBackendStub]], when creating stub responses.
-    */
-  val ExampleGet: RequestMetadata = new RequestMetadata {
+  val emptyGet: RequestMetadata = new RequestMetadata {
```
adamw, line 36: I guess we should remove all of these methods from `Response` once `TestResposne` is there?

adamw: ah sorry, we should only keep the helper methods, but with `requestMetadata: RequestMetadata`. We should move `emptyGet`, though, which kind of is the point of the issue that is being fixed

core/src/main/scala/sttp/client4/testing/AbstractBackendStub.scala
```diff
@@ -77,7 +77,7 @@ abstract class AbstractBackendStub[F[_], P](
     def thenRespond[T](body: T, statusCode: StatusCode): Self = thenRespond(Response[T](body, statusCode))
     def thenRespond[T](resp: => Response[T]): Self = {
       val m: PartialFunction[GenericRequest[_, _], F[Response[_]]] = {
-        case r if p(r) => monad.eval(resp)
+        case r if p(r) => monad.eval(resp.copy(request = r.onlyMetadata))
```
adamw, line 80: would be good to have a test for this

core/src/test/scalajvm/sttp/client4/testing/TestResponse.scala
Pask423: yeah good idea with naming, as for public hmmm not sure if we want it to be expose to users ?

core/src/main/scala/sttp/client4/Response.scala
Pask423: there is a problem here with the methods here as we use them in AbstractStubBackend when request.metadata is not computed yet, I think that we will have to leave it 
image

emptyGet should not be available outside sttp package as Respone is package scope, I thought it should be ok from user perspective

core/src/test/scalajvm/sttp/client4/testing/TestResponse.scala
adamw: yeah we do, it was exposed so far, and the idea is to use it when defining the behavior of a backend stub

core/src/main/scala/sttp/client4/Response.scala
adamw: `emptyGet` or whatever the name should still live in `ResponseStub`/`TestResponse`. The point of the issue is that this shouldn't be exposed in `Response`, but explicitly marked as a test-only thing. In a backend stub, we can of course use a `ResponseStub`

adamw:
Great, thanks!

Issue #1918, opened 2023-08-07
Request body gzip compression


nikola-mladenovic:
Hi,

Do you have a built-in mechanism for gzip compressing request body? So far I've tried the following:
```
def gzipCompressBytes(dataBytes: Array[Byte]): Array[Byte] = {
  val baos = new ByteArrayOutputStream
  val gzos = new GZIPOutputStream(baos)
  gzos.write(dataBytes)
  gzos.finish()
  gzos.close()
  baos.close()
  baos.toByteArray
}

val body: JValue = JArray(objs.toList)
val compressedBody = gzipCompressBytes(body.show.getBytes())

val baseRequest = basicRequest
 .header(ContentEncoding, "gzip")
 .header(ContentType, "application/json")
 .body(compressedBody)
 .header("accept", "application/json")
```

However this approach didn't seem to work - is there any recommendation from your end how to achieve request body compression? Thanks! 🙏

adamw:
So far there's no built-in mechanism, but you are right that this is a good candidate for a feature (just as we have auto-decompression). 

Until this is done, your "manual" approach looks fine. I tested it with https://postman-echo.com/post and got back correct results. Can you describe what problems do you encounter, and which backend you are using?

iyfedorov:
#2148

PR #2148, opened 2024-04-23
#1918 gzip and deflate content codecs

iyfedorov:
Hello! Can you look at my suggestion of concept for request content encoding support?


iyfedorov:
> Thanks for looking into this!
> 
> I think I would have two general suggestions:
> 
> 1. the compression should probably be handled in the backends themselves. Right now, if you have a chain of `Backend`s with delegates, the first one would perform the compression of the body and passed the compressed body to its delegate. I think this should rather happen at the end of the chain, right before sending. So any metrics, logging, redirect etc. backends would get the uncompressed body
> 2. then, the backends could use our default methods of compression, or use their built-in ones. For sure the default backend doesn't handle compression (based on Java's `HttpClient`), but maybe pekko-http, or async-http-client do? Something to check :)
> 
> In a somewhat similar fashion we handle decompressino (see e.g. `customEncodingHandler` and `standardEncoding` in `HttpClientBackend`)

Hello! Thank u for the answer. I will try to implement compression on one of the backends in the near future.

adamw:
Great :) I suspect the default backend - `HttpClientBackend` - would be the best one to tackle first

adamw:
Superseded by https://github.com/softwaremill/sttp/pull/2381

Issue #1681, opened 2023-01-03
Metrics names are not following OpenTelemetry conventions


chameleon82:
I know it is currently short list of agreed metrics names
https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/http-metrics/#http-client

So open this for discussion of possible span name standardization
```
sttp_request_latency          -> http.client.duration
sttp_requests_in_progress     -> http.client.active_requests
sttp_requests_success_count   -> http.client.success_count ?
sttp_requests_error_count     -> http.client.error_count ?
sttp_requests_failure_count   -> http.client.failure_count ?
sttp_request_size_bytes       -> http.client.request.size
sttp_response_size_bytes      -> http.client.response.size
```
ref:
https://github.com/softwaremill/sttp/blob/a87da572ba66ee3e9016322cf021e21ffd1ea06f/observability/opentelemetry-metrics-backend/src/main/scala/sttp/client3/opentelemetry/OpenTelemetryMetricsBackend.scala#L12

Also one of the ways those metrics can be collected and visualized with grafana: https://grafana.com/grafana/plugins/novatec-sdg-panel/
where particullary important to have next information in the metrics attributes:
- [service.name](https://opentelemetry.io/docs/reference/specification/resource/semantic_conventions/#semantic-attributes-with-dedicated-environment-variable)
- [peer.name](https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/span-general/#general-remote-service-attributes)

chameleon82:
Yes, `service.name` and `peer.name` are constant per client. But additional constant attributes may also be provided, for example for `rpc` services `rpc.name` is expected.

There are also cases where attributes can be additionally defined by request - for example, service mesh can add attribute `http.resend_count` on retries (https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/http/#http-client), but those tracing specific. For metrics, probably, that doesn't make sense to have attributes per request as attributes will be applied to aggregators.

chameleon82:
I would postpone it due semantic convention updates https://docs.google.com/presentation/d/1QUeV4v9tlCRB9X3G2wrbjAKmCLaFN5qDKAa9wS9jx0s/edit?usp=drivesdk

adamw:
@chameleon82 thanks for the heads-up :) I think it would make even more sense then to introduce a number of different metric name sets, maybe not even designating a default one, as to leave the space open for future changes (as the one above)

PR #2173, opened 2024-05-09
Refactor metrics: naming and model

### LINKED ISSUE #2151: Refactor metrics names according to standards
Research standards for metrics from Prometheus/OpenTelemetry and implement them in respective sttp Backends
Some inspiration may come from tapir

Pask423:
Conventions from:

- https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client
- https://github.com/open-telemetry/opentelemetry-specification/blob/v1.31.0/specification/metrics/api.md#instrument
- https://prometheus.io/docs/practices/naming/
- https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

closes #2151 #1681

### DIFF
--- a/observability/opentelemetry-metrics-backend/src/main/scala/sttp/client4/opentelemetry/OpenTelemetryMetricsBackend.scala
+++ b/observability/opentelemetry-metrics-backend/src/main/scala/sttp/client4/opentelemetry/OpenTelemetryMetricsBackend.scala
@@ -11,13 +11,13 @@ import java.time.Clock
 import java.util.concurrent.ConcurrentHashMap
 
 object OpenTelemetryMetricsBackend {
-  val DefaultLatencyHistogramName = "sttp.request.latency"
+  val DefaultLatencyHistogramName = "sttp.request.duration"
   val DefaultRequestSizeHistogramName = "sttp.request.size.bytes"
   val DefaultResponseSizeHistogramName = "sttp.response.size.bytes"
-  val DefaultRequestsInProgressCounterName = "sttp.requests.in_progress"
-  val DefaultSuccessCounterName = "sttp.requests.success.count"
-  val DefaultErrorCounterName = "sttp.requests.error.count"
-  val DefaultFailureCounterName = "sttp.requests.failure.count"
+  val DefaultRequestsInProgressCounterName = "sttp.requests.active"
+  val DefaultSuccessCounterName = "sttp.requests.success"
+  val DefaultErrorCounterName = "sttp.requests.error"
+  val DefaultFailureCounterName = "sttp.requests.failure"
 
 
   def apply(delegate: SyncBackend, openTelemetry: OpenTelemetry): SyncBackend =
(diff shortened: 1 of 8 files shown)


observability/opentelemetry-metrics-backend/src/main/scala/sttp/client4/opentelemetry/OpenTelemetryMetricsBackend.scala
```diff
@@ -1,23 +1,23 @@
...
 
 object OpenTelemetryMetricsBackend {
-  val DefaultLatencyHistogramName = "sttp_request_latency"
-  val DefaultRequestsInProgressCounterName = "sttp_requests_in_progress"
-  val DefaultSuccessCounterName = "sttp_requests_success_count"
-  val DefaultErrorCounterName = "sttp_requests_error_count"
-  val DefaultFailureCounterName = "sttp_requests_failure_count"
-  val DefaultRequestSizeHistogramName = "sttp_request_size_bytes"
-  val DefaultResponseSizeHistogramName = "sttp_response_size_bytes"
+  val DefaultLatencyHistogramName = "sttp.request.duration"
```
adamw, line 14: can we add some comments here which point to the conventions used? both for OT and Prometheus. I suppose using the `sttp` prefix is fine, there are no standard labels to use?

Pask423: I will add a comment 
these one https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#http-client describes the standard labels and metrics for HTTP client, I just decide to change `http.client` prefix to `sttp` beside that labels are the same

Pask423: BTW should we use Long histogram for request/response sizes instead of current Double histogram?

adamw: > BTW should we use Long histogram for request/response sizes instead of current Double histogram?

I have no idea :) How to others do it?

Pask423: Ok so I will change to `http.client prefix`
Not sure as to the histogram in convention it is describe as experimental metric.
I just realized that we do not expect floating point bytes in these histogram so Long one should be sufficient

kciesielski (approved):

observability/opentelemetry-metrics-backend/src/main/scala/sttp/client4/opentelemetry/OpenTelemetryMetricsBackend.scala
```diff
@@ -88,13 +93,13 @@ private object OpenTelemetryMetricsListener {
 private class OpenTelemetryMetricsListener(
     meter: Meter,
     clock: Clock,
-    requestToLatencyHistogramMapper: GenericRequest[_, _] => Option[CollectorConfig],
+    requestToLatencyHistogramMapper: GenericRequest[_, _] => Option[HistogramCollectorConfig],
```
chameleon82, line 96: Sorry, I'm late for the party.
To generate opentelemetry http metrics is better to use opentelemetry instrumentation api and supported metrics will be generated out of the box

PR #2187, opened 2024-05-20
Ox WebSockets

kciesielski:


effects/ox/src/main/scala/sttp/client4/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,53 @@
...
+import ox.channels.Channel
+import ox.channels.ChannelClosed
+import ox.channels.Sink
+import ox.channels.Source
+import sttp.client4.ws.SyncWebSocket
+import sttp.ws.WebSocketFrame
+
+import scala.util.control.NonFatal
+
+  extension (ws: SyncWebSocket)(using Ox) def asSourceAndSink: (Source[WebSocketFrame], Sink[WebSocketFrame]) =
```
adamw, line 13: maybe simply a function, I'm not sure if the extension method here gives us anything? you have to deconstruct it anyway.

effects/ox/src/main/scala/sttp/client4/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,53 @@
...
+import ox.channels.ChannelClosed
+import ox.channels.Sink
+import ox.channels.Source
+import sttp.client4.ws.SyncWebSocket
+import sttp.ws.WebSocketFrame
+
+import scala.util.control.NonFatal
+
+  extension (ws: SyncWebSocket)(using Ox) def asSourceAndSink: (Source[WebSocketFrame], Sink[WebSocketFrame]) = 
+    val srcChannel = Channel.bufferedDefault[WebSocketFrame]
```
adamw, line 14: maybe add a `StageCapacity` implicit param, as with stream processing stages, which controls the buffer size?

effects/ox/src/main/scala/sttp/client4/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,53 @@
...
+                false
+            case frame: WebSocketFrame =>
+                ws.send(frame)
+                true
+            case ChannelClosed.Done =>               
+                ws.send(WebSocketFrame.close)
+                false
+            case ChannelClosed.Error(err) =>
+                // There's no proper "client error" status. Statuses 4000+ are available for custom cases
+                ws.send(WebSocketFrame.Close(4000, "Client error")) // TODO should we bother the server with client error?
```
adamw, line 48: shouldn't we also then complete the `srcChannel`?

examples3/src/main/scala/sttp/client4/examples/WebSocketOx.scala
```diff
@@ -0,0 +1,32 @@
+package sttp.client4.examples
+
+import ox.*
+import ox.channels.Source
+import sttp.client4.*
```
kciesielski, line 5: @adamw There is a slight inconvenience that I keep running into in such cases. Typically I import `sttp.client4.*` first. Then I add some more code, for example a method with `(using Ox)`. My IDE suggests importing `ox.Ox`, which goes to the end of import list. This becomes an error, because `value Ox is not a member of sttp.client4.ox`. Since I am aware of this, I quickly add `import ox.*` as the first import to fix the issues, but I can imagine that such errors may confuse people pretty frequently. 
This is not only a matter of ox, some other integrations use the same pattern (like `fs2`). Or maybe it's just a problem of Metals, which should be smarter and put the import before `sttp.client4.*` automatically? How does this work for you in IJ? :)

adamw: yeah that's a problem; it's only solved for tapir+ZIO by having a `ztapir` package. So ideally we would need to have some other package name for sttp-ox and tapir-ox integrations than `xyz.ox` ;)

kciesielski: Maybe we could add one level, like `sttp.client4.interop.ox`?

adamw: I don't think this would solve the issue, since the problem is that the last component of the package is called `ox`, same as the top-level `ox` package?

adamw: Ah no, maybe not. Let's try then `sttp.client4.integ.ox`, as in tapir-cats for example?

adamw: Or maybe as in other sttp4 modules, we've got e.g. `sttp.client4.impl.cats` and `sttp.client4.httpclient.cats`. We could do the same here?

kciesielski: So `sttp.client4.impl.ox`? Ok, this would be consisted. I was pretty sure we had `sttp.client4.cats` and `sttp.client4.fs2`, etc. but there's indeed `impl`, so let's follow this convention.

adamw: Yeah, not the most beautiful, but might just work

kciesielski: I would prefer `interop` to `impl`, but let's be consistent.

effects/ox/src/main/scala/sttp/client4/ox/ws/OxWebSockets.scala
kciesielski: This question made me rework how source and sink interact. I implemented following logic:
1. When processing responses fails, close the request channel
2. When processing requests fails, close the response channel
3. When processing responses completes (a `Close` frame is received), close the request channel

There's one more thing that I'm not sure about: how to handle a completed request channel - when `channel.receiveOrClosed()` returns a `ChannelClosed.Done`. Current logic handles this by sending a `Close` frame, but I'm not sure we want it. For example, sending requests like:
```scala
    someSource.map(i => WebSocketFrame.text(s"test$i")).pipeTo(ws)
```
Sends a bunch of frames *immediately* followed by a `Close` frame when the source ends. What should we expect from the backend in such a case? I compared tapir-http4s with tapir-netty-sync:
* `tapir-http4s` processes the `Close` frame sequentially after processing all the previous frames
* `tapir-netty-sync` handles `Close` when it arrives, and sends a `Close` back immediately, regardless of all the frames that are "in-flight".
There is no specification for how a WS endpoint should behave, and I'm not sure myself which mode is better as a default.

kciesielski: I added an issue for tapir-netty servers https://github.com/softwaremill/tapir/issues/3776

docs/websockets.md
```diff
@@ -95,6 +95,55 @@ effect type      class name
...
+```
+
+See the [full example here](https://github.com/softwaremill/sttp/blob/master/examples/src/main/scala/sttp/client4/examples3/WebSocketOx.scala).
+
+Make sure that the `Source` is contiunually read. This will guarantee that server-side Close signal is received and handled.
+If you don't want to process frames from the server, you can at least handle it with a `fork { source.drain() }`.
+
+You don't need to manually call `ws.close()` when using this approach, this will be
+handled automatically underneath, according to following rules:
+ - If the request `Sink` fails with an error, the `Source` is automatically completed, sending a `Close` frame to
```
adamw, line 138: isn't this the other way round? what's the "request Sink" here?

effects/ox/src/main/scala/sttp/client4/impl/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,117 @@
+package sttp.client4.impl.ox.ws
+
+import ox.*
+import ox.channels.*
+import sttp.client4.ws.SyncWebSocket
+import sttp.ws.WebSocketFrame
+
+import scala.util.control.NonFatal
+
+/** Converts a [[SyncWebSocket]] into a pair of `Source` of server responses and a `Sink` for client requests. The
```
adamw, line 10: maybe let's use "messages" instead of request-response? With websockets, it's hard to tell which are requests, which are responses ;)

effects/ox/src/main/scala/sttp/client4/impl/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,117 @@
...
+  * @param ws
+  *   a `SyncWebSocket` where the underlying `Sink` will send requests, and where the `Source` will pull responses from.
+  * @param concatenateFragmented
+  *   whether fragmented frames from the server should be concatenated into a single frame (true by default).
+  */
+def asSourceAndSink(ws: SyncWebSocket, concatenateFragmented: Boolean = true)(using
+    Ox,
+    StageCapacity
+): (Source[WebSocketFrame], Sink[WebSocketFrame]) =
+  val requestsChannel = StageCapacity.newChannel[WebSocketFrame]
```
adamw, line 34: so that's for the messages that are sent *to* the server?

kciesielski: Yes. And if we want to speak about "messages" instead of requests/responses, how would we differentiate them? incomingMessages / outgoingMessages?

effects/ox/src/main/scala/sttp/client4/impl/ox/ws/OxWebSockets.scala
```diff
@@ -0,0 +1,117 @@
...
+              case _: ChannelClosed => false
+              case _                => true
+          case WebSocketFrame.Close(status, msg) if status > 1001 =>
+            responsesChannel.errorOrClosed(new WebSocketClosedWithError(status, msg)).discard
+            false
+          case _: WebSocketFrame.Close =>
+            responsesChannel.doneOrClosed().discard
+            false
+          case ping: WebSocketFrame.Ping =>
+            requestsChannel.sendOrClosed(WebSocketFrame.Pong(ping.payload)).discard
```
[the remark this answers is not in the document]
kciesielski: Good point, yes.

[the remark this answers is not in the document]
kciesielski: Actually, maybe not? If requests channel is closed, there may be still responses waiting to be received. That's why we don't close "responses" right after the request channel completes.

effects/ox/src/main/scala/sttp/client4/impl/ox/ws/OxWebSockets.scala
adamw: `toServer` / `fromServer`? But maybe I just got confused at the beginning and request/response is fine

docs/websockets.md
adamw: I think what I don't get here is why, when the sink fails, source is completed (that's ok - meaning it's `done()`?), and this somehow sends the `Close` frame? Logically, you send frames through the requests sink, no? :)

kciesielski: Yeah, this doesn't sound accurate. The logic is that, if the sink channel is closed with an error (for example, something unhandled failed in the user's pipeline), a Close frame is sent, and the request processing loop ends. I shall rephrase this.
> source is completed (that's ok - meaning it's done()?)
Yes, it will be `done()`, so that the responses loop stops receiving.

adamw: ok, thanks :)

PR #2193, opened 2024-05-28
Change string part encoder

### LINKED ISSUE #1966: Default multipart form encoding uses noncompliant URLEncoder for tilde (~)
**Problem:**
Following the examples in the docs for submitting form data using `multipart` method results in encoding that is not compliant with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3) specifically related to how the tilde (~) character is handled.
`multipart("form", Map("some_key" -> "~"))` results in a `StringBody(some_key=%7E,utf-8,text/plain)` form component generated where the tilde has been encoded into %7E. As per the latest spec on URL encoding, ~ is not a reserved character and should not be encoded.

It looks like this problem was partially tackled already within STTP as the internal UriCompatibility object has an additional encodeDNSHost method which uses a spec compliant Rfc3986.encode method. Unfortunately the multipart method uses the noncompliant encodeQuery method on UriCompatibility.

Is it possible to switch to using the Rfc3986 encoder instead of the URLEncoder as part of the multipart method's form handling? (edit: originally mentioned incorrect allowedCharacters set that would work here) Looks like a new character set would need to be defined within Rfc3986 object to get a fully correct allowedCharacters set. Want a PR for this?

Pask423:
closes #1966

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -151,7 +151,7 @@ val zio1InteropRsVersion = "1.3.12"
 val zio2InteropRsVersion = "2.0.2"
 
 val oxVersion = "0.2.0"
-val sttpModelVersion = "1.7.10"
+val sttpModelVersion = "1.7.11"
 val sttpSharedVersion = "1.3.19"
 
 val logback = "ch.qos.logback" % "logback-classic" % "1.5.6"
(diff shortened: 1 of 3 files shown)


adamw:
Is it possible to add a test?

PR #2250, opened 2024-08-06
Bump zio-opentelemetry version and update OpenTelemetryTracingZioBackend

Grryum:
zio-opentelemetry seems to become enough stable (RC-24 ) and it would be nice to upgrade to it latest version.


adamw:
Thanks!

PR #2252, opened 2024-08-08
Fix opentelemetry `url.full` attribute value.

Grryum:
Previously in #2250 attribute corresponding to full uri value was incorrectly filled.


kciesielski:
Thanks!

PR #2362, opened 2024-12-09
Remove BodySerializer

adamw:
Instead, bodies need to be serialized explicitly. E.g. instead of

```scala
import sttp.client4.jsoniter._

basicRequest.body(someInstance)
```

you now need to call the serialization method explicitly:

```scala
import sttp.client4.jsoniter._

basicRequest.body(asJson(someInstance))
```


Ravenow:
With this change all code with body serialization should be touched. It would be interesting to know, whats the reason for removing BodySerializers? I mean - one import on top vs explicit functions calls everywhere

adamw:
Yes, that's true. The reason is that implicit `BodySerializers` are ... well, implicit ;). They're non-obvious when you look at the code - to which form the code is serialized (if it's json or xml etc.). It's also hard to "discover", using auto-import for example.

So yes, the new approach is more explicit, requires more code, but is also predictable (when working with JSON, you always need `asJson`, both for request & response) and discoverable

Ravenow:
Is it possible to keep old way with implecits? In our code base we supply 'case classes' with circe encoder, and thats enough. Without importing of required serializer code will be 'red'. So its not obvious why sttp force users to use single way as a general purpose library. 

just my 2 cents :)

adamw:
Yes, some code will be read upon migration, I'm aware of that - but that's why this change happens across major versions (v3 -> v4). 

I'd like to keep a single way to do things, if possible. You can always create your own sttp stack (by extending the `SttpApi`, `SttpCirceApi` traits), and adding an implicit conversion from `T: Encoder` to `StringBody` using `asJson`

Then you could have a single-sttp import with all the features that you use: `import my.package.mysttp.*`

PR #2365, opened 2024-12-10
Improve usability of response specifications

### LINKED ISSUE #1771: Add a Variant of asJson that always throws
## version affected

4.0.0-M1

## Describe the issue

To me it seems strange that `asJsonAlways` merges success and error bodies for http errors, but still wraps the body in `Either` in case of deserialisation problems, which still requires the user to explicitly handle deserialisation errors.

In the spirit of `import sttp.client4.quick.*` it seems it would make sense to have a variant of `asJson` that extends `asJsonAlways` but also throws in the case of deserialisation errors, so that for scripting, assuming a happy path has no friction from handling errors.

e.g. imagine a script like

```scala
import sttp.client4.quick.*
import sttp.client4.upicklejson.*

def allNotes: ujson.Value =
  val resp = quickRequest
    .get(uri"http://localhost:8080/api/notes/all")
    .response(asJsonOrThrow[ujson.Value])
    .send()
  resp.body

allNotes.arr.foreach(println)
```

adamw:
Closes #1771

Changes:
* introduce `asXyzOrFail` methods which throw exceptions for non-2xx responses
* add `asJsonOrFail`, `asJsonOrFailEither` variants to all integrations
* `quickRequest` now uses `asStringOrFail` as the default response type
* scaladocs & docs updates
* move `.mapRight`, `.orFail` to `ResponseAs` for better discoverability (instead of extension methods), however this causes that type parameters have to be provided explicitly sometimes

cc @bishabosha @lbialy

### DIFF
--- a/async-http-client-backend/zio/src/test/scala/sttp/client4/asynchttpclient/zio/AsyncHttpClientZioHttpTest.scala
+++ b/async-http-client-backend/zio/src/test/scala/sttp/client4/asynchttpclient/zio/AsyncHttpClientZioHttpTest.scala
@@ -3,8 +3,9 @@ package sttp.client4.asynchttpclient.zio
 import sttp.client4._
 import sttp.client4.asynchttpclient.AsyncHttpClientHttpTest
 import sttp.client4.impl.zio.ZioTestBase
-import sttp.client4.testing.{ConvertToFuture, HttpTest}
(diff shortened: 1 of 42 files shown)


bishabosha:
I like the new `*orFail` stuff (and good for json)

Issue #2372, opened 2024-12-17
Upgrade org.asynchttpclient:async-http-client <3.0.1 due to CVE


gergelyfabian:
async-http-client in versions below 3.0.1 has a new CVE (2 weeks old):

https://github.com/advisories/GHSA-mfj5-cf8g-g2fv

sttp depends on this library (even in its newest version).
Unfortunately async-http-client 2.* is not binary compatible with 3.0.1, so it's not possible to upgrade the transitive dependency when using sttp.

Could this be fixed for both client3 and client4?

adamw:
I tried working on a PR, however AHC [removed](https://github.com/AsyncHttpClient/async-http-client/discussions/1925) reactive-streams support, which means that our streaming integrations top working. This means that modules such as `async-http-client-fs2` need to be removed, as they don't make sense.

At this point, I'm wondering if it's simply not better to switch to another backend and remove AHC altogether?

gergelyfabian:
> At this point, I'm wondering if it's simply not better to switch to
> another backend and remove AHC altogether?
>
Thank you. This is exactly what I did. Moved to HttpClient. It's in the
cats dependency just for the record (I had a hard time finding it).

>

gergelyfabian:
FYI, the CVE was updated: https://github.com/advisories/GHSA-mfj5-cf8g-g2fv
async-http-client 0.12.4 is not any more affected

gergelyfabian:
> At this point, I'm wondering if it's simply not better to switch to another backend and remove AHC altogether?
> Thank you. This is exactly what I did. Moved to HttpClient (both for cats-effect and Future). The cats-effect backend is in the cats dependency just for the record (I had a hard time finding it).

Just for the record I moved to OkHttp backend, as with HttpClient there were thread pool issues.

PR #2373, opened 2024-12-18
Remove async-http-client backends

adamw:
async-http-client v2 is not maintained and has security vulnerabilities 

async-http-client v3 lacks reactive-streams integration (see https://github.com/AsyncHttpClient/async-http-client/discussions/1925), hence the backends would either be significantly less useful, or much harder to implement; good alternatives (http-client, ok-http, armeria) exist

Closes https://github.com/softwaremill/sttp/issues/2372


gergelyfabian:
Thank you!

PR #2399, opened 2025-01-14
Examples of uploading and downloading files with os-lib support.

adamwkaczmarek:


examples/src/main/scala/sttp/client4/examples/DowloadFileWitOsLib.scala
```diff
@@ -0,0 +1,18 @@
+package sttp.client4.examples
```
adamw, line 1: the test needs to be scala-cli runnable and needs the metadata header for generating docs - see other examples

examples/src/main/resources/some-text-file.txt
```diff
@@ -0,0 +1 @@
+JUST SIMPLE TEXT FILE WITH SOME TEXT CONTENT
```
adamw, line 1: we can't rely on people having the sources checked out in examples, so we need to make them self-contained - all files need to be created on startup & cleaned up afterwards (in a tmp directory)

examples/src/main/scala/sttp/client4/examples/DowloadFileWitOsLib.scala
adamwkaczmarek: Yeap. I already noticed I missed that.

examples/src/main/scala/sttp/client4/examples/UploadFileWithOsLib.scala
```diff
@@ -0,0 +1,17 @@
...
+import os.*
+
+private val path: os.Path =
+  os.pwd / "examples" / "src" / "main" / "resources" / "some-text-file.txt"
+private val backend: SyncBackend = DefaultSyncBackend()
+
+@main def uploadFileWithOsLib(): Unit = {
+  val request = basicRequest
+    .post(uri"http://httpbin.org/post")
+    .body(os.read.inputStream(path))
```
adamw, line 16: ok, looks easy enough :)

can we add an example of uploading the result of a execute command-line command? (even if it's a simple `cat ...`)

examples/src/main/scala/sttp/client4/examples/UploadFileWithOsLib.scala
```diff
@@ -0,0 +1,17 @@
+package sttp.client4.examples
```
adamw, line 4: the file name should reflect the main method/class - here it should start with a small letter

examples/src/main/resources/some-text-file.txt
adamwkaczmarek: Good point. I will fix that.

examples/src/main/scala/sttp/client4/examples/UploadFileWithOsLib.scala
adamwkaczmarek: I think it should not be problem.

examples/src/main/scala/sttp/client4/examples/UploadFileWithOsLib.scala
adamwkaczmarek: Done.

examples/src/main/scala/sttp/client4/examples/downloadFileWitOsLib.scala
```diff
@@ -0,0 +1,20 @@
+// {cat=Other; effects=Direct; backend=HttpClient}: Download file
+
+//> using dep com.lihaoyi::os-lib:0.11.3
```
adamw, line 3: it also needs sttp ;) did you try running those locally?

adamwkaczmarek: Sorry I  didn't push all changes to Github , now should be good.
BTW all examples I tested locally using sbt and scala-cli.

PR #2406, opened 2025-01-19
fix warning in build.sbt

xuwei-k:
avoid deprecated `java.net.URL` constructor

- https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String)
- https://bugs.openjdk.org/browse/JDK-8295949
- https://github.com/openjdk/jdk/commit/4338f527aa81350e3636dcfbcd2eb17ddaad3914


adamw:
Thank you! :)

PR #2445, opened 2025-02-11
Add otel4s OpenTelemetry backend

### LINKED ISSUE #2432: Add otel4s integration
https://github.com/typelevel/otel4s is the OpenTelemetry library for the Cats Effect ecosystem. 

Besides [opentelemetry-java](https://github.com/open-telemetry/opentelemetry-java) integration, it also offers a Scala implementation of the OpenTelemetry specification, which is available for all platforms: JVM, Scala.js, and Scala Native.

Here is a prototype implementation: https://github.com/iRevive/otel4s-sttp:
- [MetricBackend](https://github.com/iRevive/otel4s-sttp/blob/main/modules/metrics/src/main/scala/sttp/client3/otel4s/metrics/MetricsBackend.scala)
- [TracingBackend](https://github.com/iRevive/otel4s-sttp/blob/main/modules/trace/src/main/scala/sttp/client3/otel4s/trace/TracingBackend.scala)

The integration follows OpenTelemetry Semantic Conventions for spans and metrics:
- https://opentelemetry.io/docs/specs/semconv/http/http-metrics/
- https://opentelemetry.io/docs/specs/semconv/http/http-spans/

______

It would be nice to have these integrations in the sttp repository. But there are a few questions:
- otel4s hasn't released the `1.0` yet. Even though the `otel4s-core` module is stable, binary breaking changes occur occasionally. Does sttp require only stable versions (e.g. `1.0` onwards)?
- should I provide integrations for `client3` and `client4`?

iRevive:

____

Closes #2432.

I don't think we should merge it till the otel4s `0.12.0` is released.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -14,6 +14,7 @@ val scala3 = "3.3.5"
 
 val scala2 = List(scala2_12, scala2_13)
 val scala2And3 = scala2 ++ List(scala3)
+val scala2_13And3 = List(scala2_13, scala3)
 
 val examplesScalaVersion = scala3
 val documentationScalaVersion = scala3
@@ -166,6 +167,7 @@ val osLibVersion = "0.11.4"
 val tethysVersion = "0.29.4"
 val openTelemetryVersion = "1.48.0"
 val openTelemetrySemconvVersion = "1.30.0"
+val otel4s = "0.12.0"
 val slf4jVersion = "1.7.36"
 
 val compileAndTest = "compile->compile;test->test"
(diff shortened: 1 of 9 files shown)


adamw:
Great PR, thanks! Do you know when 0.12.0 will be released?

iRevive:
We are waiting for the cats-effect 3.6.0 because some features rely on new CE functionality. So, a bit of unknown. 

I will keep the PR tidy meanwhile: resolve conflicts, fix CI issues, etc.

adamw:
@iRevive ah this one ... yes it started failing a couple of days ago. The test is to connect to an https site, which is difficult to setup in a test, so it just connect to `httpbin` AFAIK.

adamw:
I think all dependencies are now ready :) Merging, thanks for the work! :)

PR #2466, opened 2025-02-26
Support multipart streaming bodies in HttpClient based backends

### LINKED ISSUE #2407: Add support for streaming multipart bodies to the default HttpClient-based backend
Hello,

As discussed here: https://softwaremill.community/t/async-http-client-backends-removal/468/4 the default [http backend based on fs2](https://sttp.softwaremill.com/en/latest/backends/fs2.html#using-httpclient) doesn't support multipart streaming.

Thank you!

Kamil-Lontkowski:
Closes #2407


core/src/main/scalajvm/sttp/client4/internal/httpclient/BodyToHttpClient.scala
```diff
@@ -46,10 +45,17 @@ private[client4] trait BodyToHttpClient[F[_], S, R] {
       case FileBody(f, _)        => BodyPublishers.ofFile(f.toFile.toPath).unit
       case StreamBody(s)         => streamToPublisher(s.asInstanceOf[streams.BinaryStream])
       case m: MultipartBody[_] =>
-        val multipartBodyPublisher = multipartBody(m.parts)
         val baseContentType = contentType.getOrElse("multipart/form-data")
-        builder.header(HeaderNames.ContentType, s"$baseContentType; boundary=${multipartBodyPublisher.getBoundary}")
-        multipartBodyPublisher.build().unit
+        if (streams == NoStreams) {
```
adamw, line 49: alternative implementation idea: we could have a `MultipartBodyBuilder` trait, with the types of implementations: one which supports only non-streaming parts (the default), and another one,  per-streaming-type.

The trait could have methods such as `addByteArrayPart`, `addInputStreamPart` etc., and in the end it would have to produce ("somehow") a `BodyPublisher`. Maybe this would be cleaner, as then the `concatStreams` methods etc could be part of the implementations of this trait?

Kamil-Lontkowski: I added `MultipartBodyBuilder` that accepts `Seq[Part[GenericRequestBody[_]]]` and produces `F[BodyPublisher]`. It reuses logic for cases without stream support.

Kamil-Lontkowski:
For now monix and zio1 backends have disabled streaming. With zio1 problem was with reading file because it produced `ZStream[blocking ...]` and not any. Monix tests keep failing in CI but I can't reproduce it locally.

core/src/main/scalajvm/sttp/client4/internal/httpclient/MultipartBodyBuilder.scala
```diff
@@ -0,0 +1,152 @@
...
+trait StreamMultipartBodyBuilder[BinaryStream, F[_]] extends MultipartBodyBuilder[BinaryStream, F] {
+  def fileToStream(file: File): BinaryStream
+
+  def byteArrayToStream(array: Array[Byte]): BinaryStream
+
+  def concatStreams(stream1: BinaryStream, stream2: BinaryStream): BinaryStream
+
+  def toPublisher(stream: BinaryStream): F[HttpRequest.BodyPublisher]
+
+  override def multipartBodyPublisher(
```
adamw, line 75: maybe let's just call the method `apply`? the repetition doesn't give much more additional info

Kamil-Lontkowski: Done

core/src/main/scalajvm/sttp/client4/internal/httpclient/MultipartBodyBuilder.scala
```diff
@@ -0,0 +1,152 @@
...
+        case ByteArrayBody(b, _) =>
+          concatBytesToStream(accumulatedStream, encodeBytes(b, partHeaders, boundary))
+        case ByteBufferBody(b, _) =>
+          if ((b: Buffer).isReadOnly) {
+            val buffer = new ByteBufferBackedInputStream(b)
+            concatBytesToStream(accumulatedStream, encodeBytes(buffer.readAllBytes(), partHeaders, boundary))
+          } else
+            concatBytesToStream(accumulatedStream, encodeBytes(b.array(), partHeaders, boundary))
+        case InputStreamBody(b, _) =>
+          concatBytesToStream(accumulatedStream, encodeBytes(b.readAllBytes(), partHeaders, boundary))
```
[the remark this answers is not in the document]
Kamil-Lontkowski: Done

PR #2500, opened 2025-03-24
Hide sensitive information in URL query params

### LINKED ISSUE #2497: Support hiding sensitive information in URL query params
Hi! 

I'm using STTP 4 and the `Slf4jLoggingBackend` allows to set a configuration to hide sensitive headers which is great if you want to hide auth tokens but I'm calling a service provider that encodes the API Key in URL query params (something like http://blabla/one/two?since=12345&key=XXXXX).

It would be great to add a mechanism to hide this information too. WDYT?

aartigao:
Closes #2497 

When I ran the SBT tests I had a bunch of failures but doesn't seem related to the changes here (probably because of my Linux setup?), so I'd wait for the CI to have a proper report 📜

### DIFF
--- a/core/src/main/scala/sttp/client4/internal/ToCurlConverter.scala
+++ b/core/src/main/scala/sttp/client4/internal/ToCurlConverter.scala
@@ -14,10 +14,21 @@ private[client4] object ToCurlConverter {
   def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String]): String =
     apply(request, sensitiveHeaders, omitAcceptEncoding = false)
 
-  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], omitAcceptEncoding: Boolean): String = {
+  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], sensitiveQueryParams: Set[String]): String =
+    apply(request, sensitiveHeaders, omitAcceptEncoding = false, sensitiveQueryParams)
+
+  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], omitAcceptEncoding: Boolean): String =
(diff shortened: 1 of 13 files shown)


core/src/main/scala/sttp/client4/internal/ToCurlConverter.scala
```diff
@@ -14,10 +14,21 @@ private[client4] object ToCurlConverter {
...
     apply(request, sensitiveHeaders, omitAcceptEncoding = false)
 
-  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], omitAcceptEncoding: Boolean): String = {
+  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], sensitiveQueryParams: Set[String]): String =
+    apply(request, sensitiveHeaders, omitAcceptEncoding = false, sensitiveQueryParams)
+
+  def apply(request: GenericRequest[_, _], sensitiveHeaders: Set[String], omitAcceptEncoding: Boolean): String =
+    apply(request, sensitiveHeaders, omitAcceptEncoding, Set.empty)
+
+  def apply(
```
adamw, line 23: maybe we should simplify here and just keep a single constructor with all the parameters?

core/src/main/scala/sttp/client4/internal/ToRfc2616Converter.scala
```diff
@@ -12,15 +12,25 @@ private[client4] object ToRfc2616Converter {
   def requestToRfc2616(request: GenericRequest[_, _], sensitiveHeaders: Set[String]): String =
     apply(request, sensitiveHeaders)
 
+  def requestToRfc2616(
```
adamw, line 15: same here, it's an `internal` class, we can just have a single method with all params

adamw:
Thanks!

PR #2568, opened 2025-05-15
Added support for authenticated proxies to Armeria backend

threeseed:


adamw:
Ah, `asJavaProxySelector` indeed doesn't take into account authentication. Do you think it's possible to support `onlyProxyHosts` /  `nonProxyHosts` in your code? Also, is it correct to always add proxy config, even if none is specified in the backend options?

threeseed:
a) It works with your existing asJavaProxySelector code which already handles onlyProxyHosts/nonProxyHosts.

b) ProxyConfig.DIRECT means no proxy as described [here](https://javadoc.io/doc/com.linecorp.armeria/armeria-javadoc/1.11.0/com/linecorp/armeria/client/proxy/ProxyConfig.html) so should be fine.

PR #2934, opened 2026-07-03
Fix digest auth failing with MD5-sess algorithm (#2933)

### LINKED ISSUE #2933: Digest auth broken with MD5-sess
If the algorithm in digest auth is MD5-sess, the code fails because a message digest algorithm for MD5-sess does not exist. The code should instead just use MD5

Katrix:
Fixes #2933

Scalafmt modified files I have not touched. Everything commited is formatted well.


adamw:
Looks good, thanks! :)

PR #2935, opened 2026-07-03
Fix #1465: No way to send continuous WebSocketFrame with Fs2 stream pipes

### LINKED ISSUE #1465: No way to send continuous WebSocketFrame with Fs2 stream pipes
Hi,

I'm using sttp with tapir/http4s/fs2 to build a websocket server.
My websocket endpoint is described with tapir as:

```
  val instance = endpoint.get
    .in(PREFIX)
    .out(webSocketBodyRaw(Fs2Streams[IO]))
```
and my business logic returns a `fs2.Pipe[IO, WebSocketFrame, WebSocketFrame]`

I have an use case when I would like to send a continuous WebSocketFrame (ie make a stream emit a continuous WebSocketFrame)
which I can't since WebSocketFrame exposes the `finalFragment` parameter only, not the `isContinuation` one

`isContinuation` is exposed in `sttp.ws.WebSocket#send` which is called by `sttp.client3.impl.fs2.Fs2WebSockets#handleThroughPipe`
with `isContinuation` always set to false.

Would that be possible to somehow expose the `isContinuation` flag when using the web socket with stream pipes ?

Thanks !

Regards,
Sebastien

dabrowski-adam:
Updates `Fs2WebSocket.handleThroughPipe` to mark continuation frames. I think this was a regression introduced in [afd307744](https://github.com/softwaremill/sttp/commit/afd307744#diff-f6c63a5db0d17a07ce7182efc57b8ca6dec36126427c12ef06ac8da8ffc07e8bL43).

When writing tests, I encountered an issue with duplicate Close frames, so I fixed that as well.

---
---
Closes #1465

### DIFF
--- a/effects/fs2-ce2/src/main/scala/sttp/client4/impl/fs2/Fs2WebSockets.scala
+++ b/effects/fs2-ce2/src/main/scala/sttp/client4/impl/fs2/Fs2WebSockets.scala
@@ -3,6 +3,7 @@ package sttp.client4.impl.fs2
 import cats.effect.ConcurrentEffect
 import cats.effect.concurrent.Ref
 import cats.effect.implicits._
+import cats.syntax.all._
 import fs2.{Pipe, Stream}
 import sttp.ws.{WebSocket, WebSocketClosed, WebSocketFrame}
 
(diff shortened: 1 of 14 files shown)


dabrowski-adam:
I think that both issues also affect `ZioWebSockets` and `MonixWebSockets`.

dabrowski-adam:
I think there could be some further improvements made to handling of Close frames, namely checking if the user hasn't already sent a Close frame before echoing the server's Close frame, but I don't won't to bloat this PR.

effects/fs2-ce2/src/main/scala/sttp/client4/impl/fs2/Fs2WebSockets.scala
```diff
@@ -24,33 +25,47 @@ object Fs2WebSockets {
...
+                case _: WebSocketFrame.Close => closeSent.set(true)
+                case _                       => ().pure[F]
+              })
+            } // send messages
+        }
+        .compile
+        .drain
+        // the stream already emits a matching/user Close when appropriate; only fall back to a
+        // default Close if none was sent (e.g. the pipe finished without one)
+        .guarantee(closeSent.get.flatMap(sent => if (sent) ().pure[F] else ws.close()))
```
dabrowski-adam, line 67: I'm tempted to use `Applicative.unlessA`, but it would make it slightly harder to cross-check the FS2, ZIO and Monix implementations.

adamw:
Nice, thanks! :)

PR #2958, opened 2026-08-02
docs: add readTimeout param description to RequestOptions scaladoc

mmustafasenoglu:
## Summary
Adds the missing  description to  scaladoc.

## Motivation
The  field was the only parameter in  without a  description. This was noted in [#2835](https://github.com/softwaremill/sttp/issues/2835).

## Changes
- : Added  description explaining the timeout behavior and which backends support it.

## Testing
- Compiled with  — no issues (scaladoc-only change)

Closes #2835


adamw:
Thanks!

PR #2978, opened 2026-08-24
feat: add closeFactory/closeClient parameter to usingClient/usingDefaultClient

### LINKED ISSUE #2628: Confusing ArmeriaCatsBackend initialization
ArmeriaCatsBackend can be created with `apply` and then `closeFactory` will be true, but if it is created with `usingClient` or `usingDefaultClient` then `closeFactory` is`false`.

I don't know how to interpret it and what was the motivation behind it.

What I wanted to do is to create a method that would return SttpBackend with all the needed decorations of WebClient, but it's only possible to do it with `usingClient` which has `closeFactory` set to `false`.

mmustafasenoglu:
The usingClient and usingDefaultClient factory methods on Armeria and OkHttp backends hardcoded closeFactory/closeClient to false, giving users no way to have the backend close a user-provided client.

This adds an optional closeFactory/closeClient: Boolean = false parameter to all public factory methods, allowing users to control whether the backend should close the underlying client on backend.close().

Armeria backends (9 files):
- ArmeriaCatsBackend (CE3 + CE2)
- ArmeriaFutureBackend
- ArmeriaFs2Backend (CE3 + CE2)
- ArmeriaZioBackend (ZIO 2 + ZIO 1)
- ArmeriaMonixBackend
- ArmeriaScalazBackend

OkHttp backends (2 files):
- OkHttpSyncBackend
- OkHttpFutureBackend

Closes #2628

### DIFF
--- a/armeria-backend/cats-ce2/src/main/scala/sttp/client4/armeria/cats/ArmeriaCatsBackend.scala
+++ b/armeria-backend/cats-ce2/src/main/scala/sttp/client4/armeria/cats/ArmeriaCatsBackend.scala
@@ -57,11 +57,11 @@ object ArmeriaCatsBackend {
   def resourceUsingClient[F[_]: Concurrent](client: WebClient): Resource[F, Backend[F]] =
     Resource.make(Sync[F].delay(apply(client, closeFactory = true)))(_.close())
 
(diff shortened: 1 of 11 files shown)


adamw:
Thanks - although compilation fails in a number of cases

Issue #2992, opened 2026-09-11
Support Java 25 without warnings


atennapel:
We have an application that uses various Scala dependencies, including sttp. When we run with Java 25 we get the following warning:
```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$
```
I believe this is causes by the dependencies being compiled with a version of Scala before 3.8 and not using the flag `-Yfuture-lazy-vals`.

Sttp seems to use Scala 3.3.8 (so not 3.8+) and does not use `-Yfuture-lazy-vals`, and also uses `lazy val`, so using this dependency will also cause the above warnings.

Can there be a version of Sttp released using Scala 3.8+ and/or using `-Yfuture-lazy-vals`, so that this warning can be fixed?

adamw:
We'll update to Scala 3.9 LTS soon probably, so this should fix it for Scala 3. We could also probably just add the flag ... did you check if it works if you add it to the build?

atennapel:
I tried it and adding the flag works (together with `-java-output-version 11`), except for the scala-native's `nscplugin`. It gives the following error:
```
[error] ## Exception when compiling 92 sources to C:\Users\Albert.ten.Napel\git\sttp\core\target\native-3\classes
[error] java.lang.IllegalArgumentException: requirement failed: symbol class AbstractCurlBackend val Context$lzy1$lzyHandle is not a method
[error] scala.Predef$.require(Predef.scala:337)
[error] scala.scalanative.nscplugin.NirGenType.resolve$1(NirGenType.scala:318)
[error] scala.scalanative.nscplugin.NirGenType.genMethodSigImpl$$anonfun$1(NirGenType.scala:334)
[error] scala.collection.mutable.HashMap.getOrElseUpdate(HashMap.scala:469)
[error] scala.scalanative.nscplugin.NirGenType.genMethodSigImpl(NirGenType.scala:334)
[error] scala.scalanative.nscplugin.NirGenType.genMethodSig(NirGenType.scala:305)
[error] scala.scalanative.nscplugin.NirGenType.genMethodSig$(NirGenType.scala:22)
[error] scala.scalanative.nscplugin.NirCodeGen.genMethodSig(NirCodeGen.scala:19)
[error] scala.scalanative.nscplugin.NirGenExpr$ExprBuffer.genApplyStaticMethod(NirGenExpr.scala:1323)
[...]

PR #2996, opened 2026-09-15
Use -Yfuture-lazy-vals for Scala 3 on the JVM

atennapel:
This PR adds the `-Yfuture-lazy-vals` flag to Scala 3 on the JVM. See https://github.com/softwaremill/sttp/issues/2992 . With this change JVM 25 will no longer give warnings about using a terminally deprecated API. Note that this will cause Java 8 to be unsupported. Java 9 is the minimum due to the fact that `-Yfuture-lazy-vals` uses the `VarHandle` API.

---


atennapel:
Looks like the Scala 3 Native CI jobs are failing because the flag is also added there. I thought `commonJvmSettings` was not used for native, but it looks like it is.

adamw:
Pushed a fix to the branch.

The cause: `http4sBackend` and `scribeBackend` called `.settings(commonJvmSettings)` on the matrix itself, which applies to every platform row, not just JVM. So Native and JS picked up the flag too. Both matrices already pass the right settings to each `.jvmPlatform` / `.jsPlatform` / `.nativePlatform`, so the matrix-level call was redundant and I removed it.

Checked locally: `scribeBackendNative3`, `http4sBackendNative3` and `scribeBackendJS3` no longer have the flag, `scribeBackend3` and `http4sBackend3` still do, and `scribeBackendNative3/compile` passes.

adamw:
One more thing pushed: the `ox` backend now emits Java 21 bytecode instead of 11.

`-java-output-version` also decides which JDK APIs the compiler sees, not just the class file version, so the number needs to match what a module actually requires. Ox needs JDK 21, so that module now says 21. Made the version a setting key (`javaOutputVersion`, default 11) so a module can override it.

Checked: `compileScoped 3 JVM` passes, with and without `ONLY_LOOM`.

adamw:
Reverted the Java 21 commit — it broke the build, sorry.

CI builds on JDK 11 as well as 21, and a JDK 11 compiler rejects `-java-output-version 21` outright: "21 is not a valid choice". The `ox` module gets compiled in the JDK 11 jobs (directly on the Scala 3 one, and through `compileDocs` on the 2.12 and 2.13 ones), so all of them failed, plus MiMa.

11 everywhere is fine anyway. Nothing in the ox module calls a JDK 21 API, and users get the JDK 21 requirement from ox's own artifact.

